PR #3, opened 2017-09-19
Add `opened_ports()` function

galgalesh:
Returns opened ports as a list of strings: `['8080/tcp', '8081-8083/tcp']`

This PR includes working tests but I haven't tested this on an actual charm because I don't know how. How do I build a charm using my local copy of charm-helpers?


johnsca:
You can either use a modified copy of [layer:basic](https://github.com/juju-solutions/layer-basic) or, with the `--candidate` (possibly `--edge`) version of the `charm` snap, use `--wheelhouse` to override the charmhelpers dep.  You can then point it to this branch using:

```
-e git+https://github.com/galgalesh/charm-helpers#egg=charmhelpers
```

I would appreciate the testing of the `--wheelhouse` feature if you wouldn't mind using that.

galgalesh:
Thanks, @johnsca, this works with the candidate channel.

What is the format of wheelhouse.txt? Can I point this to a local directory instead of the github repo?

galgalesh:
I tested it in a charm and it works as it should. I also updated `HACKING.md` to explain how to build a charm with a custom charmhelpers branch.

marcoceppi:
It's whatever works for typical python requirements.txt -
https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format

johnsca:
The file format is the standard `requirements.txt` format for pip.  I know you can use `-e git+file://` URLs but you do have to ensure that the changes you are testing are committed, as uncommitted changes will be ignored with that.  I think pip might also support `-e /foo` local paths directly, but I'm not certain.

I was aware that both copies end up in the wheelhouse; this is actually an existing issue with the wheelhouse creation, and it happens any time the same package is pulled in multiple times when processing the dependencies of packages in the wheelhouse.  Unfortunately, I'm not really sure how to address it without reimplementing the entire dependency resolution algorithm from pip.  I would very much appreciate help with that.  The actual installation will use the highest version in the wheelhouse, including dev versions, so it might be worth mentioning in the `HACKING.md` to append `dev0` to the VERSION to ensure it gets used.

PR #8, opened 2017-09-22
Modulo Distribution

thedac:
Charms have needed a way to distribute load during expensive operations
in large scale deployments. This change adds two helpers to mitigate
load.

modulo_distribution uses unit number, a modulo value and a constant wait time
to produce a calculated wait time distribution.

If you have 1000 nodes that need to restart 100 at a time 1 minute at a time:

      time.wait(modulo_distribution(modulo=100, wait=60))
      restart()

If you need restarts to happen serially set modulo to the exact number of
nodes and set a high constant wait time:

      time.wait(modulo_distribution(modulo=10, wait=120))
      restart()

distributed_wait is a helper for OpenStack charms to use which checks
for config values to pass to modulo_distribution and uses the returned
value to use with time.wait().

### DIFF
--- a/charmhelpers/contrib/hahelpers/cluster.py
+++ b/charmhelpers/contrib/hahelpers/cluster.py
@@ -27,6 +27,7 @@ clustering-related helpers.
 
 import subprocess
 import os
+import time
 
 from socket import gethostname as get_unit_hostname
 
@@ -44,6 +45,9 @@ from charmhelpers.core.hookenv import (
     unit_get,
     is_leader as juju_is_leader,
     status_set,
+)
+from charmhelpers.core.host import (
+    modulo_distribution,
 )
 from charmhelpers.core.decorators import (
     retry_on_exception,
@@ -361,3 +365,24 @@ def canonical_url(configs, vip_setting='vip'):
     else:
         addr = unit_get('private-address')
     return '%s://%s' % (scheme, addr)
+
+
+def distributed_wait(modulo=None, wait=None):
+    ''' Distribute operations by waiting based on modulo_distribution
+
+    If modulo and or wait are not set, check config_get for those values.
+
+    :side effect: Calls config_get()
+    :side effect: Calls log()
+    :side effect: Calls status_set()
+    :side effect: Calls time.sleep()
+    '''
+    if modulo is None:
+        modulo = config_get('modulo-nodes')
+    if wait is None:
+        wait = config_get('known-wait')
+    calculated_wait = modulo_distribution(modulo=modulo, wait=wait)
+    msg = "Waiting {} seconds for operation ...".format(calculated_wait)
+    log(msg, DEBUG)
+    status_set('maintenance', msg)
+    time.sleep(calculated_wait)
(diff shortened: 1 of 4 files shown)


charmhelpers/contrib/hahelpers/cluster.py
```diff
@@ -361,3 +365,24 @@ def canonical_url(configs, vip_setting='vip'):
...
+def distributed_wait(modulo=None, wait=None):
+    ''' Distribute operations by waiting based on modulo_distribution
+
+    If modulo and or wait are not set, check config_get for those values.
+
+    :side effect: Calls config_get()
+    :side effect: Calls log()
+    :side effect: Calls status_set()
+    :side effect: Calls time.sleep()
+    '''
```
wolsen: I think we should provide some :param: docstrings which explain the values a bit (even though its clear to me). Also, I think we need to be explicit with what the unit wait interval is specified as (seconds).

charmhelpers/contrib/hahelpers/cluster.py
```diff
@@ -361,3 +365,24 @@ def canonical_url(configs, vip_setting='vip'):
...
+    if modulo is None:
+        modulo = config_get('modulo-nodes')
+    if wait is None:
+        wait = config_get('known-wait')
+    calculated_wait = modulo_distribution(modulo=modulo, wait=wait)
+    msg = "Waiting {} seconds for operation ...".format(calculated_wait)
```
wolsen: I find operation to be a vague word from a user perspective here. The end user will get more value out of it if this says what the operation is rather than just operation (e.g. "Waiting n seconds to restart" or something). What about making this customizable?

charmhelpers/core/host.py
```diff
@@ -946,3 +946,31 @@ def updatedb(updatedb_text, new_path):
...
     output = "\n".join(lines)
     return output
+
+
+def modulo_distribution(modulo=3, wait=30):
+    """ Modulo distribution
+
+    This helper uses the unit number, a modulo value and a constant wait time
+    to produce a calculated wait time distribution.
+    This is useful in large scale deployments to distribute load during an
```
wolsen: nit: blank line between paragraphs or make one paragraph.

charmhelpers/contrib/hahelpers/cluster.py
thedac: Yep, that was an oversight.

charmhelpers/contrib/hahelpers/cluster.py
thedac: Sounds good.

charmhelpers/core/host.py
thedac: Fixed.

thedac:
Ready for a re-review

johnsca:
I think this is a fantastic light-weight solution, but for reference and posterity, if you **really** need to ensure no overlap, the [coordinator layer](https://git.launchpad.net/layer-coordinator/tree/README.md) uses leadership (via [`charmhelpers.coordinator`](https://github.com/juju/charm-helpers/blob/master/charmhelpers/coordinator.py)) to coordinate cross-unit distributed operations.  The guaranteed non-overlap does come at the cost of latency and network chatter, due to using leadership data and hooks, though, so there's a significant trade-off to consider.

PR #11, opened 2017-09-25
update readme to include python benefits

dshcherb:
Some useful traits that I found about using python for charms as opposed
to using compiled languages.

Might be worthwhile to have that directly on the first page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,25 @@
 
 ## Overview
 
-CharmHelpers provides an opinionated set of tools for building Juju
-charms that work together. In addition to basic tasks like interact-
-ing with the charm environment and the machine it runs on, it also
-helps keep you build hooks and establish relations effortlessly.
+CharmHelpers provides an opinionated set of tools for building Juju charms.
+
+## Common Usage Examples
+
+* interaction with charm-specific Juju unit agents via hook tools;
+* processing of events and execution of decorated functions based on event names;
+* handling of persistent storage between independent charm invocations;
+* rendering of configuration file templates;
+* modification of system configuration files;
+* installation of packages;
+* retrieval of machine-specific details;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,9 +3,16 @@
 ## Overview
 
 CharmHelpers provides an opinionated set of tools for building Juju
-charms that work together. In addition to basic tasks like interact-
-ing with the charm environment and the machine it runs on, it also
-helps keep you build hooks and establish relations effortlessly.
+charms that work together. In addition to basic tasks like interacting
+with the charm environment and the machine it runs on, it also helps
+keep you build hooks and establish relations effortlessly.
```
ajkavanagh: This line doesn't read very well?  I'm not sure what it is saying.  Is it "... also helps you to build hooks and establish relations effortlessly"?

README.md
```diff
@@ -3,9 +3,16 @@
...
+
+## Why Python?
+
+* Low entry barrier, fairly common in automation tools;
+* An interpreted language helps with charm portability across different CPU architectures;
+* Doesn't require debugging symbols (just use pdb in-place);
+* An author or a user is able to make debugging changes without recompiling a charm.
```
ajkavanagh: I would drop the first bullet point, and probably replace it with:

* Python is an extremely popular, easy to learn, and powerful language which is also common in automation tools.

README.md
tvansteenburgh: s/keep//

dshcherb: Well, that was the original one - I just fixed formatting. I'll fix the comment itself as well.

dshcherb:
Pull-request updated, HEAD is now 67687eeeb5a5ac14643a96cc82580e5372983170

PR #15, opened 2017-09-27
Support python dict in sysctl_create

javacruft:
Add support for directly passing a python dict rather than a
YAML formatted string to sysctl_create to make it easier for
charms to set sane defaults without hopping via YAML first.


charmhelpers/core/sysctl.py
```diff
@@ -31,18 +31,22 @@
...
     :type sysctl_file: str or unicode
     :returns: None
     """
-    try:
-        sysctl_dict_parsed = yaml.safe_load(sysctl_dict)
-    except yaml.YAMLError:
-        log("Error parsing YAML sysctl_dict: {}".format(sysctl_dict),
-            level=ERROR)
-        return
+    if type(sysctl_dict) is not dict:
```
ajkavanagh: This would probably be more flexible if it used:

```python
if isinstance(syctl_dict, dict):
```

as it would allow dict like objects to also be used.  If you wanted to be *really* flexible, then you could use:

```python
if isinstance(sysctl_dict, collections.abc.Mapping):
```

Then it would allow any object that supports the dictionary specific methods (e.g. __getattr__, __iter__, etc.) to be used.

johnsca: You might consider using `if not isinstance(sysctl_dict, collections.Mapping):` instead, on the off chance that the charm wants to pass in, e.g., an `OrderedDict` instead.  Not really a blocker, though, since the order wouldn't be preserved with the existing code, either.

ajkavanagh: Indeed, `collections.Mapping` is Py2 compatible.

stub42: importing collections.Mapping raises a warning under Py3.7, and should be collections.abc.Mapping. We might need a conditional import.

PR #16, opened 2017-09-27
Drop use of SharedDBContext in Neutron plugin config

javacruft:
Database information has been configured at the neutron.conf
level since the migration of neutron services out of the
nova-cloud-controller charm.

Drop all use of the SharedDBContext in plugin setup.


javacruft:
I've done some testing in the neutron-api charm - this change considerably reduces the hook execution time as the __call__ method was being executed multiple times (one for every instance of the context) on each hook execution.

PR #17, opened 2017-09-29
Fixing incorrect definition

joedborg:
'int' should be 'internal' to avoid KeyError when deploying with DNS-HA
set to True.

### DIFF
--- a/charmhelpers/contrib/openstack/ip.py
+++ b/charmhelpers/contrib/openstack/ip.py
@@ -28,7 +28,7 @@ from charmhelpers.contrib.network.ip import (
 from charmhelpers.contrib.hahelpers.cluster import is_clustered
 
 PUBLIC = 'public'
-INTERNAL = 'int'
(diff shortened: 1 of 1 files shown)


joedborg:
Fixes
```python
  File "/var/lib/juju/agents/unit-ceph-radosgw-0/charm/hooks/charmhelpers/contrib/openstack/ip.py", line 138, in resolve_address
    net_type = ADDRESS_MAP[endpoint_type]['config']
KeyError: 'internal'
```

joedborg:
It could be a bug the other way around, i.e. that ip.py should be looking for 'int', rather than 'internal'.  If someone could confirm, that'd be great.

thedac:
It does need to be 'int' for resolve_address. I will submit a separate change.

ajkavanagh:
This PR was superseded by https://github.com/juju/charm-helpers/pull/18 which addresses this bug.  Thanks for working with us on that one.

PR #19, opened 2017-10-02
Enable IPv6 check in haproxy template file

tytus-kurek:
This patchset adds IPv6 check to the haproxy template file.

Partial-Bug: #1719280


javacruft:
Hi Tytus

Looks like a good start; with regards to unit tests, I'd like to see some coverage to a) ensure that is_ipv6_disabled is being called and that b) its return value is being placed into the context.

You'll need to update the unit tests to setup the return values for this function

tytus-kurek:
Hi James,

I've updated unit tests.

thedac:
This is looking good. But as James requested we still need a bit more coverage in the unit tests. 

"a) ensure that is_ipv6_disabled is being called and that b) its return value is being placed into the context."

tytus-kurek:
@javacruft / @thedac : unit tests have been updated.

PR #23, opened 2017-10-17
Add function to control password expiry

gnuoy:
Add a wrapper around chage to all password expiry settings to be
altered. Also add a function to remove password expiry entirely.

Partial-Bug: #1686085

### DIFF
--- a/charmhelpers/core/host.py
+++ b/charmhelpers/core/host.py
@@ -441,6 +441,48 @@ def add_user_to_group(username, group):
     subprocess.check_call(cmd)
 
 
+def chage(username, lastday=None, expiredate=None, inactive=None,
+           mindays=None, maxdays=None, root=None, warndays=None):
+    """Change user password expiry information
+
+    :param str username: User to update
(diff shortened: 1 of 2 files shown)


charmhelpers/core/host.py
```diff
@@ -441,6 +441,48 @@ def add_user_to_group(username, group):
...
+                         -1 will remove an account's inactivity.
+    :param str mindays: Set the minimum number of days between password
+                        changes to MIN_DAYS.
+                        0 indicates the password can be changed anytime.
+    :param str maxdays: Set the maximum number of days during which a
+                        password is valid.
+                        -1 as MAX_DAYS will remove checking maxdays
+    :param str root: Apply changes in the CHROOT_DIR directory
+    :param str warndays: Set the number of days of warning before a password
+                         change is required
```
ajkavanagh: Be good to know which exceptions might be raised.
e.g. I'm expecting subprocess.CalledProcessError will be thrown if any of those params are incorrect.

PR #25, opened 2017-10-19
[wip] Fixed to handle unicode in charm metadata

hloeung:
Per LP#1721200 [1], charm metadata.yaml may contain unicode characters. We can't open with 'encoding=' because that's Python 3.x only so let's use decode('utf-8') instead.

[1]https://bugs.launchpad.net/bugs/1721200

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -657,12 +657,19 @@ def _port_op(op_name, port, protocol="TCP"):
     else:
         _args.append('{}/{}'.format(port, protocol))
     try:
-        subprocess.check_call(_args)
-    except subprocess.CalledProcessError:
+        subprocess.check_output(_args, stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError as e:
         # Older Juju pre 2.3 doesn't support ICMP
         # so treat it as a no-op if it fails.
-        if not icmp:
-            raise
+        if icmp:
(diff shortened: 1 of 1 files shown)


ajkavanagh:
Please can we have a test with some UTF-8 (non ASCII) characters in a metadata.yaml that exercises this code?  This is to prevent regressions in the future.  Thanks.

hloeung:
Okay, added a test. I think I have it right, but it's not often I write tests.

johnsca:
The PR description mentions decoding as utf8, but the code decodes as ascii.  Seems like utf8 would be the right choice.

ajkavanagh:
@hloeung @johnsca The `.decode('ascii', 'ignore')` is to just strip any unicode chars that can't be decoded to ascii.  Is that the intention?  If so, I'd rather we had a Py2/Py3 break, in that the PY2 version of the code decodes to 'ascii', and the PY3 version decodes to utf-8 as it uses unicode strings by default, anyway.

Thoughts?

hloeung:
I'll take a look and redo/fix this soon. I'm not sure of how to mark a pull request as "Work in Progress" in GitHub.

javacruft:
Pull request out of date and not actually subject of PR.

hloeung:
Still a WIP, revisiting this after https://github.com/juju/charm-helpers/issues/232

PR #31, opened 2017-10-26
Add support for snap branch usage

javacruft:
Allow use of snap branches in openstack-origin, for example:

   snap:pike/edge/jamespage
   snap:ocata/stable/thedac


coreycb:
LGTM as well

PR #35, opened 2017-11-02
Ingress address function

thedac:
The new features in Juju set ingress-address and egress-networks on all
relations based on network spaces. Ingress-address should be used
over private-address when it is available. As this is a new feature we
fall back to private-address when ingress-address is not available.

This new function consumes the relation data and returns the
ingress-address when available or the private-address when not.

The code is relatively simple. As a review, I request input on the nomenclature as well. Is ingress_address too vague? Some other options:
get_ingress_address
get_relation_ingress_address
get_remote_ingress_address
etc.

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -22,6 +22,7 @@ from __future__ import print_function
 import copy
 from distutils.version import LooseVersion
 from functools import wraps
+from collections import namedtuple
 import glob
 import os
 import json
@@ -1135,28 +1136,6 @@ def network_get(endpoint, relation_id=None):
     return yaml.safe_load(response)
 
 
-def ingress_address(rid=None, unit=None):
-    """
-    Retrieve the ingress-address from a relation when available. Otherwise,
-    return the private-address. This function is to be used on the consuming
-    side of the relation.
-
-    Usage:
-    addresses = []
-    for rid in relation_ids(relation_name):
-        for unit in related_units(rid):
-             addresses.append(ingress_address(rid=rid, unit=unit))
-
-    :param rid: string relation id
-    :param unit: string unit name
-    :side effect: calls relation_get
-    :return: string IP address
-    """
-    settings = relation_get(rid=rid, unit=unit)
-    return (settings.get('ingress-address') or
-            settings.get('private-address'))
-
-
 def add_metric(*args, **kwargs):
     """Add metric values. Values may be expressed with keyword arguments. For
     metric names containing dashes, these may be expressed as one or more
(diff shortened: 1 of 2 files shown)


coreycb:
LGTM. I think the name makes sense if we will eventually drop the pivate-address bit.

stub42:
Should this fall back to returning unit_private_ip() if network-get fails with old versions of Juju?

Given that all 'provider' charms need to be updated to support cross model relations, if there is no fallback then these charms will likely drop Juju 1.25 support at the same time. Which is not necessarily a terrible thing.

charmhelpers/core/hookenv.py
```diff
@@ -1135,6 +1135,28 @@ def network_get(endpoint, relation_id=None):
...
+    """
+    Retrieve the ingress-address from a relation when available. Otherwise,
+    return the private-address. This function is to be used on the consuming
+    side of the relation.
+
+    Usage:
+    addresses = []
+    for rid in relation_ids(relation_name):
+        for unit in related_units(rid):
+             addresses.append(ingress_address(rid=rid, unit=unit))
```
ajkavanagh: Not a problem here, but something that really does annoy me, is that number of times I see this pattern in code; it's crying out for a helper.

e.g.

```python

RelationAddr = NamedTuple('RelationAddr', 'rid unit')

def iter_units_for_relation_name(relation_name):
    for rid in relation_ids(relation_name):
        for unit in related_units(rid):
            yield RelationAddr(rid, unit)

# usage
addresses = [ingress_address(rid=u.rid, unit=u.unit)
             for u in iter_units_for_relation_name(name)]
```

I feel like I need to add this to charmhelpers ...

ajkavanagh:
I think the function name maps nicely to what it's asking juju to do.  @stub42 makes an interesting point.  I realised I don't know how old you need to go with Juju before `relation-get ...` doesn't have a private-address and you have to use `unit-get ...`.

stub42:
Oh, private-address is always there (unless the charm removed it, which is it technically allowed to do). I'm just confusing myself.

charmhelpers/core/hookenv.py
thedac: I totally agree. I actually think juju should provide a simple:
relation-get --all -r RID
But that is wishful thinking.
I am +1 on your generator.

thedac:
@Stub42,

Yes just to be clear, this is the other end consuming the relation data that was set with network-get on the other side of the relation. unit-get private-address is now replaced with network-get.

@All,

It sounds like people are OK with the name. Unless Alex wants to hold off for the generator I think this is ready for an -1/+1 vote.

PR #36, opened 2017-11-04
Update NRPE to not restart the nrpe-server on update-status hooks

hloeung:
update-status hooks are configured to fire every 5 minutes by default. Unfortunately, charms written to use this code doesn't always handle this properly (especially with reactive charms) and as a result, update-status triggers a restart of nrpe-server causing various alerts.

Let's work around it here to not restart.

### DIFF
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -30,6 +30,7 @@ import yaml
 
 from charmhelpers.core.hookenv import (
     config,
+    hook_name,
     local_unit,
     log,
     relation_ids,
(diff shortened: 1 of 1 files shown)


charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -302,7 +303,12 @@ def write(self):
                 "command": nrpecheck.command,
             }
 
-        service('restart', 'nagios-nrpe-server')
+        # update-status hooks are configured to firing every 5 minutes by
+        # default. When nagios-nrpe-server is restarted, the nagios server
+        # reports checks failing causing unneccessary alerts. Let's not restart
+        # on update-status hooks.
+        if not hook_name == 'update-status':
```
stub42: You want hook_name(), not hook_name. Its a function.

hloeung:
Fixed, thanks.

PR #37, opened 2017-11-05
py3: Encode template before writing to file

javacruft:
Ensure that under Python 3, the rended template content is encoded
before writing to the target file.


gnuoy:
LGTM, thanks

PR #47, opened 2017-11-15
Changes for py3 support in keystone py3 migration

ajkavanagh:


ajkavanagh:
Actually, this change is overtaken by events, so I'm dropping this PR.

PR #63, opened 2017-12-03
Increase timeout in _deploy() to 1800 for snap packaged charm payloads

ajkavanagh:
The timeout increase allows the gate-basic-snap-xenial-ocata to pass with Python 3 which needs extended timescales as there is a manager file that gets called multiple times.  Due to how snaps don't allow cached python files (readonly fs), this means that the tests run much more slowly.

### DIFF
--- a/charmhelpers/contrib/amulet/deployment.py
+++ b/charmhelpers/contrib/amulet/deployment.py
@@ -78,7 +78,7 @@ class AmuletDeployment(object):
 
     def _deploy(self):
         """Deploy environment and wait for all hooks to finish executing."""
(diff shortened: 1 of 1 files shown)


ajkavanagh:
This timeout may not be long enough.  I have a keystone failure, but I need to check whether it is this timeout or one in the basic_deployment.py tests.

thedac:
Alex,

The point of that code is to allow override of the timeout via AMULET_SETUP_TIMEOUT. Which for the record, gets set in tox.ini in our OS charms as 2700. 

I would be a -1 on this. As upping the default value will not help you.

ajkavanagh:
David, fair enough.  I wasn't sure about this one.  I'm closing this one.  Thanks!

Issue #65, opened 2017-12-04
need to provide volume api version for internal endpoint context


dosaboy:
... in contrib.openstack.context.InternalEndpointContext

dosaboy:
See https://bugs.launchpad.net/charm-nova-compute/+bug/1733566 for more info

PR #66, opened 2017-12-04
Use volumev3 for Openstack >= Pike (#65)

dosaboy:
The InternalEndpointContext() should select volume
api version based on which version of Openstack is
used.

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -101,6 +101,8 @@ from charmhelpers.contrib.openstack.utils import (
     git_determine_python_path,
     enable_memcache,
     snap_install_requested,
+    CompareOpenStackReleases,
+    os_release,
 )
 from charmhelpers.core.unitdata import kv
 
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/context.py
```diff
@@ -1567,7 +1569,14 @@ class InternalEndpointContext(OSContextGenerator):
     endpoints.
     """
     def __call__(self):
-        return {'use_internal_endpoints': config('use-internal-endpoints')}
+        ctxt = {'use_internal_endpoints': config('use-internal-endpoints')}
+        rel = os_release('cinder-common', base='icehouse')
```
javacruft: Is the intent that this context is generated on units hosting services which don't have cinder installed, i.e. they make use of cinder endpoints, but won't have cinder-common installed?

If so checking the cinder-common package does not make sense to me.

dosaboy: fair enough, i think i'll just have the package passed in via the instantiator or somesuch since im not aware of a generic way to get ost release version.

ajkavanagh:
I can't help feeling that this code went into the wrong place.

PR #71, opened 2017-12-05
Handle profile creation in luminous

chris-sanders:
Luminous changes the failure domain flag in profile creation, this simply checks the version and makes the change. This time squashed into a single commit.


chris-sanders:
Sure thing. 
While adding the second verification I've noticed a typo in this function. I don't think failure domains ever worked. At least a quick check [on the Jewel docs](http://docs.ceph.com/docs/jewel/rados/operations/erasure-code/) seems to indicate it's always used '-' not '_' so I've updated that as well.

chris-sanders:
@wolsen (or anyone able to review) is there anything else you need me to address to review this?

wolsen:
@chris-sanders lgtm, thanks for the updates!

PR #72, opened 2017-12-06
Add funcs for listing & extending logical volumes

gnuoy:
This patch adds two new functions list_logical_volumes and
extend_logical_volume_by_device. It also adds 3 functools.partial
based functions for common use cases for list_logical_volumes.

These new functions are needed by the cinder charm to resolve
Bug #1735931.

Partial-Bug: #1735931


andrewdmcleod:
lgtm +1 (manually validated lvs commands / switches)

PR #73, opened 2017-12-11
Update HAProxy default timeout values

thedac:
The default HAProxy timeout values are fairly strict. On a busy cloud
it is common to exceed one or more of these timeouts. The only
indication that HAProxy has exceeded a timeout and dropped the
connection is errors such as "BadStatusLine" or "EOF." These can be
very difficult to diagnose when intermittent.

This change updates the default timeout values to more real world
settings. These values have been extensively tested in ServerStack.
Configured values will not be overridden.


coreycb:
LGTM

PR #74, opened 2017-12-12
Ensure json file only changed if necessary

dosaboy:
contrib.openstack.utils.update_json_file must only write
the file of the dictionary has actually been modified to
avoid unnecessary service restarts from code checking
for file changes.

### DIFF
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -2045,14 +2045,25 @@ def token_cache_pkgs(source=None, release=None):
 
 def update_json_file(filename, items):
     """Updates the json `filename` with a given dict.
-    :param filename: json filename (i.e.: /etc/glance/policy.json)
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/utils.py
```diff
@@ -2045,14 +2045,28 @@ def token_cache_pkgs(source=None, release=None):
...
     :param items: dict of items to update
     """
+    if not items:
+        return
+
     with open(filename) as fd:
         policy = json.load(fd)
+
+    if not policy:
+        return
```
ajkavanagh: This is a change of behaviour.  If it's an 'empty' json file it would either break or be updated (not sure which).  This will now just silently return.

i.e. previous code might have relied on it writing something or breaking.

dosaboy: That's fair, if it were an empty file the json.load would error with ValueError. I'll rework this.

PR #80, opened 2017-12-22
Fix for pool_set function and validator handling of strings

chris-sanders:
Fix non-string inputs for pool_set
 - New test for pool_set with bool
 - New test for pool_set with str
 - Fixed test for pool_set with int

Fix handling of strings in validator.
 - New test that six.string_type still works with this change
 - New test that type 'str' works, test fails w/o the above fix

### DIFF
--- a/charmhelpers/contrib/storage/linux/ceph.py
+++ b/charmhelpers/contrib/storage/linux/ceph.py
@@ -113,7 +113,7 @@ def validator(value, valid_type, valid_range=None):
         assert isinstance(valid_range, list), \
             "valid_range must be a list, was given {}".format(valid_range)
         # If we're dealing with strings
-        if valid_type is six.string_types:
+        if isinstance(value, six.string_types):
             assert value in valid_range, \
                 "{} is not in the list {}".format(value, valid_range)
         # Integer, float should have a min and max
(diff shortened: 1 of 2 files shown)


ajkavanagh:
Is this in relation to a bug?  If so, please could we have the Issue raised so that this is a fix for it?  (i.e. why is this fix being proposed).  Thanks!

chris-sanders:
Sure, I'm not aware of a bug. I raised this because I was trying to use the relation with Ceph-fs and Ceph-mon. Trying to pass a non-string value to the Ceph-mon via the broker dies and this was the cause. Hence the reason I found both the error in casing (or lack there of) and the issue with types during validation. I didn't raise a bug b/c I had a fix by the time I knew what was wrong, so I cherry-picked these fixes out of my charm for the MR.

ajkavanagh:
@chris-sanders:

> There isn't a requirement for creating a bug for an MR is there?

No there isn't, and I was being a bit 'over process' in asking for it.  Sorry about that.  I guess what I was asking for was the explanation that you've just given, which was excellent.  I admit, that I usually *do* raise a bug when I hit a bug in something that's been shipped (e.g. charm-helpers) to allow a conversation to happen if necessary, but also occasionally forgot to and just submit a fix.

Thanks for the pull-request!

chris-sanders:
Thanks, that is a good point I'll be sure to open bugs for anything I don't fix as a side effect of other changes 😀

PR #82, opened 2018-01-05
Add helpers for hacluster interface type

javacruft:
Most OpenStack API charms that make use of the hacluster charm
follow the same model with regards to ha resource configuration.

Provide a helper function to support generic generation of the
data set used when charms are deployed in HA configurations with
the hacluster charm.

### DIFF
--- a/charmhelpers/contrib/openstack/ha/utils.py
+++ b/charmhelpers/contrib/openstack/ha/utils.py
@@ -23,6 +23,8 @@
 Helpers for high availability.
 """
 
+import json
+
 import re
 
 from charmhelpers.core.hookenv import (
@@ -32,6 +34,7 @@ from charmhelpers.core.hookenv import (
     config,
     status_set,
     DEBUG,
+    WARNING,
 )
 
 from charmhelpers.core.host import (
@@ -40,6 +43,23 @@ from charmhelpers.core.host import (
 
 from charmhelpers.contrib.openstack.ip import (
     resolve_address,
+    is_ipv6,
+)
+
+from charmhelpers.contrib.network.ip import (
+    get_iface_for_address,
+    get_netmask_for_address,
+)
+
+from charmhelpers.contrib.hahelpers.cluster import (
+    get_hacluster_config
+)
+
+JSON_ENCODE_OPTIONS = dict(
+    sort_keys=True,
+    allow_nan=False,
+    indent=None,
+    separators=(',', ':'),
 )
 
 
@@ -53,8 +73,8 @@ class DNSHAException(Exception):
 def update_dns_ha_resource_params(resources, resource_params,
                                   relation_id=None,
                                   crm_ocf='ocf:maas:dns'):
-    """ Check for os-*-hostname settings and update resource dictionaries for
-    the HA relation.
+    """ Configure DNS-HA resources based on provided configuration and
+    update resource dictionaries for the HA relation.
 
     @param resources: Pointer to dictionary of resources.
                       Usually instantiated in ha_joined().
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/ha/utils.py
```diff
@@ -93,50 +185,75 @@ def update_dns_ha_resource_params(resources, resource_params,
...
             continue
 
         hostname_group.append(hostname_key)
-        resources[hostname_key] = crm_ocf
-        resource_params[hostname_key] = (
-            'params fqdn="{}" ip_address="{}" '
+        relation_data['resources'][hostname_key] = crm_ocf
+        relation_data['resource_params'][hostname_key] = (
+            'params fqdn="{}" ip_address="{}"'
             ''.format(hostname, resolve_address(endpoint_type=endpoint_type,
```
ajkavanagh: Shouldn't need the `''` on this line.  i.e. just the `.format(hostname, ...)

(I know it's not part of the change, but an opportunity to clean up?)

charmhelpers/contrib/openstack/ha/utils.py
```diff
@@ -64,7 +84,79 @@ def update_dns_ha_resource_params(resources, resource_params,
...
+
+def assert_charm_supports_dns_ha():
+    """Validate prerequisites for DNS HA
+    The MAAS client is only available on Xenial or greater
+    """
+    if lsb_release().get('DISTRIB_RELEASE') < '16.04':
+        msg = ('DNS HA is only supported on 16.04 and greater '
+               'versions of Ubuntu.')
+        status_set('blocked', msg)
+        raise DNSHAException(msg)
```
ajkavanagh: More of a question, than an issue:  Is the `DNSHAException` supposed to be caught by the caller for it to do something with?  (If so, probably should add a `:raises DNSHAException: if release is < 16.04` to the docstring.

If not, doesn't this mean on a default Juju deploy that juju will just retry the failing hook?

charmhelpers/contrib/openstack/ha/utils.py
javacruft: Thanks for the review; update incoming.

charmhelpers/contrib/openstack/ha/utils.py
javacruft: I think documenting the fact that this raises an exception is good anyway; I think that currently juju will just end up trying the failed hook execution which is not awesome - we don't have a nice way to handle this in non-reactive charms right now.

charmhelpers/contrib/openstack/ha/utils.py
```diff
@@ -93,50 +189,75 @@ def update_dns_ha_resource_params(resources, resource_params,
...
 
         hostname_group.append(hostname_key)
-        resources[hostname_key] = crm_ocf
-        resource_params[hostname_key] = (
-            'params fqdn="{}" ip_address="{}" '
-            ''.format(hostname, resolve_address(endpoint_type=endpoint_type,
-                                                override=False)))
+        relation_data['resources'][hostname_key] = crm_ocf
+        relation_data['resource_params'][hostname_key] = (
+            'params fqdn="{}" ip_address="{}"'
```
thedac: Are we certain the space after 'ip_address="{}" ' is not necessary?

charmhelpers/contrib/openstack/ha/utils.py
```diff
@@ -64,7 +84,83 @@ def update_dns_ha_resource_params(resources, resource_params,
     @param crm_ocf: Corosync Open Cluster Framework resource agent to use for
                     DNS HA
     """
+    _relation_data = {'resources': {}, 'resource_params': {}}
```
thedac: The doc strings above are now lying to us. Need to sort the doc strings for update_dns_ha_resource_params() and update_hacluster_dns_ha.

charmhelpers/contrib/openstack/ha/utils.py
```diff
@@ -64,7 +84,83 @@ def update_dns_ha_resource_params(resources, resource_params,
...
 
+    return {
+        'json_{}'.format(k): json.dumps(v, **JSON_ENCODE_OPTIONS)
+        for k, v in _relation_data.items() if v
+    }
+
+
+def update_hacluster_dns_ha(service, relation_data,
+                            crm_ocf='ocf:maas:dns'):
+    """ Configure DNS-HA resources based on provided configuration
```
thedac: See above regarding doc string confusion.

charmhelpers/contrib/openstack/ha/utils.py
javacruft: To be honest I'm not sure - it just looked a bit wonky so I tidied it up.

charmhelpers/contrib/openstack/ha/utils.py
javacruft: update_dns_ha_resource_params does what it did before - so I don't think this needs changing?

charmhelpers/contrib/openstack/ha/utils.py
javacruft: See comment above - however I did spot a missing doc string so re-added it.

charmhelpers/contrib/openstack/ha/utils.py
thedac: If you expand above, the update_dns_ha_resource_params, starts with "Check for os-*-hostname" but the block of code that corresponds to that is now in update_hacluster_dns_ha.

charmhelpers/contrib/openstack/ha/utils.py
thedac: Have we tested this? There are several places where there are " '. I think this is because these strings get concatenated.

I'll run a quick test.

PR #84, opened 2018-01-07
add extra-policy context and section

dshcherb:
* Extra policy context is added to render service-specific policy
drop-in files (newer yaml or json format);
* section-extra-policy is trivial but is included in service-specific
policy-drop-in templates - this is mainly for naming policy files
differently to allow primary and subordinate charm policy files to
co-exist in a single directory without collisions;
* every classic charm to use this context will need to get extra-policy
config option;
* a policy directory is created if not present (policy.d is used by
default but it is intentionally passed as an __init__ argument to allow
potential overrides.

See pad.lv/1741723.

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -1779,3 +1779,25 @@ class MemcacheContext(OSContextGenerator):
                     ctxt['memcache_server_formatted'],
                     ctxt['memcache_port'])
         return ctxt
+
+
+class ExtraPolicyContext(OSContextGenerator):
+    """A context to render extra policy for a service"""
+
+    def __init__(self, policyd):
+        """
+        @param policyd: an absolute path to a policy directory which is
+        configured in policy_dirs oslo.policy option.
+        """
+        self.policyd = policyd
+
+    def __call__(self):
+        # make it if it's not there
+        mkdir(self.policyd)
+
+        ctxt = {}
+        # if a charm uses this context it should define this config option
+        extra_policy = config('extra-policy')
+        # some oslo.policy versions error out if an empty file is used
+        ctxt['extra_policy'] = extra_policy if extra_policy else '{}'
+        return ctxt
--- a/charmhelpers/contrib/openstack/templates/section-extra-policy
+++ b/charmhelpers/contrib/openstack/templates/section-extra-policy
@@ -1,0 +1,1 @@
+{{ extra_policy }}


charmhelpers/contrib/openstack/context.py
```diff
@@ -1779,3 +1779,24 @@ def __call__(self):
...
+    """A context to render extra policy for a service"""
+
+    def __init__(self, policyd):
+        """
+        @param policyd: an absolute path to a policy directory which is
+        configured in policy_dirs oslo.policy option.
+        """
+        self.policyd = policyd
+        # make it if it's not there
+        mkdir(policyd)
```
ajkavanagh: Let's not put `mkdir(...)` here.  Despite the fact I hate that these contexts have side-effects (rather than just being contexts), this ought probably to be in the `__call__()` as now `mkdir` would need to be mocked *before* the file that contains the call to instantiate this context is imported.  Obviously, it could be in the form `ExtraPolicyContext()()`, but let's at least have a chance.

dshcherb: I hate this behavior too but I could not find any other place that would make sense for multiple classic charms.

I can move this to __call__().

ajkavanagh: I wasn't getting at you :)  ... it's a common pattern in the charms and tricky to fix - however, in the `__call__` at least we can refactor charms to move the call to `__call__()` out of the load-time code (eventually). Thanks for fixing!

dshcherb: No negativity implied :^)

Just wanted to say that I had a similar impression about contexts being more than contexts which is not obvious when such a change is needed.

dshcherb:
Pull-request updated, HEAD is now b2a94f64cc99565e02be0856acdccddb623cc03d

dshcherb:
One feature I have not addressed with this PR is drop-in syntax such as:

"domain_id:{{ admin_domain_id }}"

This would imply rendering a temporary template file visible by loaders created in charm-helpers and rendering that with a charm-specific context (other than ExtraPolicyContext).

ajkavanagh:
Just a thought: This is starting to feel "design-y" in need, and therefore a use-case is probably useful.  Perhaps you could flesh out how you think it will work in a doc somewhere (here in the repo might be good?) so you can describe how you think the config option would work - I can guess, for example.  However, if you document the use-case with the policy-extra (and a dynamic element) then it might clear up how it's needed.

dshcherb:
Hmm, I think I have a better way of doing this but it requires "render from a string" type of functionality and I will create a separate PR for that. In this case the policy section will go away and I will only need to create a policy.d directory defined per-charm at charm "build-time" in the ExtraPolicyContext.

Usage of context for drop-ins is desirable because we depend on certain context being present in policy files (admin_domain_id, admin_role, service_tenant_id). If I have to deploy a large bundle with a policy drop-in pre-defined I'd rather avoid waiting until the deployment ends to then manually retrieve IDs and update extra-policy post-deployment.

Here's a sample of policy.json used in keystone:

```
{
    "admin_required": "role:{{ admin_role }}",
    "cloud_admin": "rule:admin_required and (is_admin_project:True or domain_id:{{ admin_domain_id }} or project_id:{{ service_tenant_id }})",
    "service_role": "role:service",
...
    "identity:get_region": "",
    "identity:list_regions": "",
    "identity:create_region": "rule:cloud_admin",
    "identity:update_region": "rule:cloud_admin",
    "identity:delete_region": "rule:cloud_admin",

    "identity:get_service": "rule:admin_required",
```

I may want to define another rule that would use admin_domain_id or other variables from the keystone context in a drop-in file.

dshcherb:
https://github.com/juju/charm-helpers/pull/87 (render from_string)
https://github.com/juju/charm-helpers/pull/88 (extra policy context)

PR #87, opened 2018-01-08
Add from_string template rendering capability

dshcherb:
* sometimes there is a need to render a drop-in file based on a charm
config option;
* this drop-in file may be a jinja2 template itself by-design which may
depend on a certain context specified in a charm;
* jinja2 allows templates to be loaded from memory using from_string
method of a given Environment object;
* charm-helpers currently do not allow this but it is fairly
straightforward to extend.

Sample usage:

```
BASE_RESOURCE_MAP = OrderedDict([
...
    (EXTRA_POLICY_FILE, {
        'contexts': [keystone_context.KeystoneContext(),
                     context.ExtraPolicyContext(policyd=POLICY_DIR)],
        'services': BASE_SERVICES,
        'config_template': config('extra-policy')
    }),
...

def resource_map():
...
    return resource_map

def register_configs():
...
    configs = templating.OSConfigRenderer(templates_dir=TEMPLATES,
                                          openstack_release=release)
    for cfg, rscs in resource_map().iteritems():
        configs.register(
            cfg, rscs['contexts'],
            rscs.get('config_template')
        )
...
```

See pad.lv/1741723 and https://git.io/vNLDG

### DIFF
--- a/charmhelpers/contrib/openstack/templating.py
+++ b/charmhelpers/contrib/openstack/templating.py
@@ -93,7 +93,8 @@ class OSConfigTemplate(object):
     Associates a config file template with a list of context generators.
     Responsible for constructing a template context based on those generators.
     """
-    def __init__(self, config_file, contexts):
+
+    def __init__(self, config_file, contexts, config_template=None):
         self.config_file = config_file
 
         if hasattr(contexts, '__call__'):
@@ -102,6 +103,8 @@ class OSConfigTemplate(object):
             self.contexts = contexts
 
         self._complete_contexts = []
+
+        self.config_template = config_template
 
     def context(self):
         ctxt = {}
@@ -123,6 +126,11 @@ class OSConfigTemplate(object):
             return self._complete_contexts
         self.context()
         return self._complete_contexts
+
+    @property
+    def is_string_template(self):
+        """:returns: Boolean if this instance is a template initialised with a string"""
+        return self.config_template is not None
 
 
 class OSConfigRenderer(object):
@@ -148,6 +156,10 @@ class OSConfigRenderer(object):
                          contexts=[context.IdentityServiceContext()])
         configs.register(config_file='/etc/haproxy/haproxy.conf',
                          contexts=[context.HAProxyContext()])
+        configs.register(config_file='/etc/keystone/policy.d/extra.cfg',
+                         contexts=[context.ExtraPolicyContext()
+                                   context.KeystoneContext()],
+                         config_template=hookenv.config('extra-policy'))
         # write out a single config
         configs.write('/etc/nova/nova.conf')
         # write out all registered configs
@@ -218,14 +230,23 @@ class OSConfigRenderer(object):
             else:
                 apt_install('python3-jinja2')
 
-    def register(self, config_file, contexts):
+    def register(self, config_file, contexts, config_template=None):
         """
         Register a config file with a list of context generators to be called
         during rendering.
+        config_template can be used to load a template from a string instead of
+        using template loaders and template files.
+        :param config_file (str): a path where a config file will be rendered
+        :param contexts (list): a list of context dictionaries with kv pairs
+        :param config_template (str): an optional template string to use
         """
-        self.templates[config_file] = OSConfigTemplate(config_file=config_file,
-                                                       contexts=contexts)
-        log('Registered config file: %s' % config_file, level=INFO)
+        self.templates[config_file] = OSConfigTemplate(
+            config_file=config_file,
+            contexts=contexts,
+            config_template=config_template
+        )
+        log('Registered config file: {}'.format(config_file),
+            level=INFO)
 
     def _get_tmpl_env(self):
         if not self._tmpl_env:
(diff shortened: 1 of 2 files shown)


dshcherb:
Needed for https://github.com/juju/charm-helpers/pull/88 and https://review.openstack.org/#/c/531614/ to work.

dshcherb:
Pull-request updated, HEAD is now 42694ffef90d4bb08181936ebd71a10966bb806d

charmhelpers/contrib/openstack/templating.py
```diff
@@ -124,6 +127,20 @@ def complete_contexts(self):
         self.context()
         return self._complete_contexts
 
+    @property
+    def config_file(self):
+        return self._config_file
+
+    @property
+    def from_string(self):
```
ajkavanagh: For clarify, I would name this `def is_string_template(self)` as that's what the property is really reporting.  It would also, I think (but open to persuasion!) be simpler as:

```python
    @property
    def is_string_template(self):
        """:returns: Boolean if this instance is a template initialised with a string"""
        return self._config_template is not None
```

charmhelpers/contrib/openstack/templating.py
```diff
@@ -218,13 +235,18 @@ def __init__(self, templates_dir, openstack_release):
             else:
                 apt_install('python3-jinja2')
 
-    def register(self, config_file, contexts):
+    def register(self, config_file, contexts, config_template=None):
         """
         Register a config file with a list of context generators to be called
         during rendering.
+        config_template can be used to load a template from a string instead of
+        using template loaders and template files.
```
ajkavanagh: Opportunity here to add params and their types to the docstring.  Partiularly for the `config_template` param which is actually an string that gets converted to an OSConfigTemplate.

charmhelpers/contrib/openstack/templating.py
```diff
@@ -238,29 +260,48 @@ def _get_template(self, template):
         log('Loaded template from %s' % template.filename, level=INFO)
         return template
 
+    def _get_template_from_string(self, ostmpl):
```
ajkavanagh: Missing docstring

charmhelpers/contrib/openstack/templating.py
```diff
@@ -238,29 +260,48 @@ def _get_template(self, template):
...
+        self._get_tmpl_env()
+        template = self._tmpl_env.from_string(ostmpl.config_template)
+        log('Loaded a template from a string for %s' % (
```
ajkavanagh: Please can we use newstyle string formats:

```python
    log('... {}'.format(ostmpl.config_file))
```

charmhelpers/contrib/openstack/templating.py
```diff
@@ -238,29 +260,48 @@ def _get_template(self, template):
...
-
-        _tmpl = os.path.basename(config_file)
-        try:
-            template = self._get_template(_tmpl)
-        except exceptions.TemplateNotFound:
-            # if no template is found with basename, try looking for it
-            # using a munged full path, eg:
-            #   /etc/apache2/apache2.conf -> etc_apache2_apache2.conf
-            _tmpl = '_'.join(config_file.split('/')[1:])
+
```
ajkavanagh: As the `OSConfigRenderer()` class' behaviour is changing, it would be a good idea to add a comment to it so that the new functionality is exposed.  e.g. how to use it, etc.

charmhelpers/contrib/openstack/templating.py
```diff
@@ -238,29 +260,48 @@ def _get_template(self, template):
...
+        ostmpl = self.templates[config_file]
+        ctxt = ostmpl.context()
+
+        if ostmpl.from_string:
+            template = self._get_template_from_string(ostmpl)
+            log('Rendering from a string template: %s' % config_file,
```
ajkavanagh: New style format please.

charmhelpers/contrib/openstack/templating.py
```diff
@@ -238,29 +260,48 @@ def _get_template(self, template):
...
-                    (self.templates_dir, os.path.basename(config_file), _tmpl),
-                    level=ERROR)
-                raise e
-
-        log('Rendering from template: %s' % _tmpl, level=INFO)
+            except exceptions.TemplateNotFound:
+                # if no template is found with basename, try looking
+                # for it using a munged full path, eg:
+                # /etc/apache2/apache2.conf -> etc_apache2_apache2.conf
+                _tmpl = '_'.join(config_file.split('/')[1:])
```
ajkavanagh: This assumes that the config_file always starts with a `/`.  Is that the case?  Can we stick an assert / if statement to verify as otherwise this may do odd things.

charmhelpers/contrib/openstack/templating.py
```diff
@@ -238,29 +260,48 @@ def _get_template(self, template):
...
+                try:
+                    template = self._get_template(_tmpl)
+                except exceptions.TemplateNotFound as e:
+                    log('Could not load template from %s by %s or %s.' %
```
ajkavanagh: New style format string, please.

charmhelpers/contrib/openstack/templating.py
```diff
@@ -238,29 +260,48 @@ def _get_template(self, template):
...
+                        (
+                            self.templates_dir,
+                            os.path.basename(config_file),
+                            _tmpl
+                        ),
+                        level=ERROR)
+                    raise e
+
+            log('Rendering from template: %s' % config_file, level=INFO)
```
ajkavanagh: New style format string, please.

charmhelpers/contrib/openstack/templating.py
```diff
@@ -124,6 +127,20 @@ def complete_contexts(self):
         self.context()
         return self._complete_contexts
 
+    @property
```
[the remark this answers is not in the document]
dshcherb: I can remove that but this is just for visual appearance - by looking at this I do not need to grep in different codebases to know that this might be used outside the class.

charmhelpers/contrib/openstack/templating.py
dshcherb: Agreed.

charmhelpers/contrib/openstack/templating.py
dshcherb: ack

charmhelpers/contrib/openstack/templating.py
dshcherb: Merely re-formatted this part to be included in the if statement branch. I think if changes are needed here it should be in a separate PR.

charmhelpers/contrib/openstack/templating.py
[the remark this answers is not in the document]
ajkavanagh: I tend to think this pattern just adds noise to the code - it'll be simpler without it and you can avoid writing a docstring!

charmhelpers/contrib/openstack/templating.py
ajkavanagh: Fair enough, ignore this one.

ajkavanagh:
I'm not seeing any tests for the new functionality?  Please could you add some, and ensure that they work with py2 and py3.  Thanks.

dshcherb:
Pull-request updated, HEAD is now eb68a833cf4e9e2a77c697153754256ade706eb8

dshcherb:
Pull-request updated, HEAD is now 21872915b88af78ea45f4b34f08ab987e567973d

dshcherb:
Pull-request updated, HEAD is now da839771f8e2dba9200d10f60ca29fadb2d8a7b7

PR #88, opened 2018-01-08
add extra policy context

dshcherb:
Mainly serves as a single place to create a custom policy directory if
it is not already there.

See pad.lv/1741723 and https://git.io/vNLDG

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -1784,3 +1784,30 @@ class MemcacheContext(OSContextGenerator):
                     ctxt['memcache_server_formatted'],
                     ctxt['memcache_port'])
         return ctxt
+
+
+class EnsureDirContext(OSContextGenerator):
+    '''
+    Serves as a generic context to create a directory as a side-effect.
+
+    Useful for software that supports drop-in files (.d) in conjunction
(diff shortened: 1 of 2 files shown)


dshcherb:
To be used with https://github.com/juju/charm-helpers/pull/87

charmhelpers/contrib/openstack/context.py
```diff
@@ -1775,3 +1775,19 @@ def __call__(self):
                     ctxt['memcache_server_formatted'],
                     ctxt['memcache_port'])
         return ctxt
+
+
+class ExtraPolicyContext(OSContextGenerator):
```
ajkavanagh: I may be missing something here, but the This Context's purpose purely seems to be for its side-effect?  I'm guessing that a template also uses 'policyd' at the top level to indicate the directory to whatever is reading the policies.

In this case, therefore, I think we need an explanation of how this policy is to be used with an example in the class' docstring, please.

ajkavanagh:
I'm not seeing any tests with this change.  Please could you add unit tests that demonstrate how it is used.  (It'll be a very simple test, but it helps to 'fix' the usage pattern).

dshcherb:
Pull-request updated, HEAD is now a6e54693536af8dcd48720fb38ea0856d6589217

PR #90, opened 2018-01-15
Fix credential gathering for haproxy NRPE checks

devec0:
Addresses LP#1743287 and LP#1713165

Fixes haproxy checks to correctly gather credentials from the haproxy config file for use when checking the status of backends via the admin endpoint with NRPE checks


javacruft:
Fixed in #95

PR #93, opened 2018-01-17
Fix Swift package version check

dosaboy:
Related-Bug: #1743847


coreycb:
LGTM. I really don't understand the separate path for swift so I'm ok removing it and clearly there was a bug in it.

PR #97, opened 2018-01-19
Handle no network binding exception gracefully

thedac:
When network_get_primary_address is called with a binding name that
does not exist a subprocess.CalledProcessError is thrown.

There are times when get_relation_ip will be called with a binding that
may or may not exist. It would be better if this exception is handled
gracefully.

An example is the HAProxyContext in openstack-dahsboard. The dashboard
charm does not have the typical extra bindings: admin, internal, public.
Rather than make this an exception, make get_relation_ip handle a
non-existent binding gracefully.

This change updates network_get_primary_address to throw a
NoNetworkBinding exception when a binding does not exist.
get_relation_ip will then handle this exception by using the default
address.

### DIFF
--- a/charmhelpers/contrib/network/ip.py
+++ b/charmhelpers/contrib/network/ip.py
@@ -27,6 +27,7 @@ from charmhelpers.core.hookenv import (
     network_get_primary_address,
     unit_get,
     WARNING,
+    NoNetworkBinding,
 )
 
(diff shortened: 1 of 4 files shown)


javacruft:
I'll take a look - another alternative might be #98

javacruft:
Testing with the dashboard charm now

PR #98, opened 2018-01-20
Support use of HAProxy context in dashboard charm

javacruft:
The Openstack Dashboard charm only has a public network binding,
which means the default HAProxy behaviour to inspect public,
internal and admin bindings breaks; support override of the
list of bindings to inspect during creation of the context.


javacruft:
Testing under https://review.openstack.org/#/c/535759/

PR #99, opened 2018-01-22
Account for password field name change in PXC 5.7

coreycb:
The database field name for password has been changed to
authentication_string in PXC 5.7. PXC 5.7 was introduced in
Ubuntu 18.04 (Bionic).


coreycb:
NOTE: I haven't uploaded the PXC 5.7 package yet. It still needs a review.

PR #102, opened 2018-01-28
add string template rendering to core/templating

dshcherb:
Another change to allow rendering a template from a specified string but
now to core/templating.py which is used by all reactive OpenStack
charms.

This change is required to allow a charm writer to support specifying
config template parts in config options or render config files based on
templates passed via relation data from other charms.

The main use-case is drop-in files which contain custom logic not
possible to predict or model in the charm code by providing relevant
config options.

For example, RBAC rules are subjective and some operators have
to implement policy based on those subjective rules which may
differ from defaults. If those rules cannot be provided via an API
to a given service and are stored in config files this is the only way
to provide desired customization options.

Related to pad.lv/1741723
In addition to https://github.com/juju/charm-helpers/pull/87 but for
reactive charms which mainly use this render function.

### DIFF
--- a/charmhelpers/core/templating.py
+++ b/charmhelpers/core/templating.py
@@ -20,7 +20,8 @@ from charmhelpers.core import hookenv
 
 
 def render(source, target, context, owner='root', group='root',
-           perms=0o444, templates_dir=None, encoding='UTF-8', template_loader=None):
+           perms=0o444, templates_dir=None, encoding='UTF-8',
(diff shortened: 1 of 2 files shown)


ajkavanagh:
Can we edit your first comment to indicate the bug this relates and and any other information pertinent to the change; this will become part of the commit message when it's merged and it's useful to have the information.

dshcherb:
@ajkavanagh done, ta.

PR #105, opened 2018-02-01
Bug/lp 1745985/apache2 is not reloaded when updating ssl certificate

tytus-kurek:


charmhelpers/contrib/openstack/context.py
```diff
@@ -800,6 +801,8 @@ def configure_cert(self, cn=None):
         write_file(path=os.path.join(ssl_dir, key_filename),
                    content=b64decode(key))
 
+        service_restart('apache2')
```
fnordahl: Any reason not to do a service_reload here instead?

This will in effect call apachectl graceful and issuing a reload without impacting any on-going connections.

barryprice:
I don't think this will do what you want.

Apache [needs to be fully restarted](https://serverfault.com/questions/590868/will-a-graceful-reload-of-apache-configuration-require-the-ssl-cert-password) in order to pick up new keys/certificates.

[It looks like service_reload will just do a reload](https://github.com/juju/charm-helpers/blob/master/charmhelpers/core/host.py).

You might want to consider a graceful restart here to avoid service disruption.

fnordahl:
> I don't think this will do what you want.
> 
> Apache needs to be fully restarted in order to pick up new keys/certificates.
> 
> It looks like service_reload will just do a reload.
> 
> You might want to consider a graceful restart here to avoid service disruption.

@barryprice The Apache init scripts calls apachectl graceful which indeed does a restart though gracefully.

This does reload the SSL certificate if it has changed.

fnordahl:
To be clear the reload function of apache2 init script calls apachectl graceful which means calling service_reload is the right thing to do here.

barryprice:
Ah sorry about that, you're absolutely right.

PR #110, opened 2018-02-05
Use volumev3 for Openstack >= Pike (#65) (#66)

wolsen:
The InternalEndpointContext() should select volume
api version based on which version of Openstack is
used.

(cherry picked from commit 3aa436da7151770e104df7d2ff013fbca3cb4367)

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -101,6 +101,8 @@ from charmhelpers.contrib.openstack.utils import (
     git_determine_python_path,
     enable_memcache,
     snap_install_requested,
+    CompareOpenStackReleases,
+    os_release,
 )
 from charmhelpers.core.unitdata import kv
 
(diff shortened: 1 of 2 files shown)


ajkavanagh:
I know this is a cherry pick, but the original code looks like it's "gone in the wrong place".  i.e. volume API decisions in code that is only trying to 'pick up' the endpoint config option.  Is there still time to undue the main change and put it somewhere that makes more sense (even if it's a new context class)?

wolsen:
@ajkavanagh A fair point. I too thought it was a bit off when I first saw it, since its conflating visibility of endpoints and versions of endpoints.

I'm happy to move the api version context bit into its own VolumeAPIContext. Naturally, we'll also need to update the corresponding charms in [LP #1733566](https://bugs.launchpad.net/charm-nova-compute/+bug/1733566); which I'm also happy to do.

Ideally, this would come along by way of the [service-discovery spec](https://specs.openstack.org/openstack/charm-specs/specs/queens/approved/service-discovery.html) as it seems a natural fit. Yet, progress is seemingly slow or non-existent for the service-discovery work unfortunately.

ajkavanagh:
@wolsen yes, your suggestion makes a lot of sense.  I like the separation of concerns by moving it to its own context.  And yes, service-discovery would be the best fit.  Maybe we add a FIXME comment to the code for a later fix when it becomes available?

wolsen:
@ajkavanagh great, thanks for confirming. The FIXME sounds reasonable. I'll close this PR and file a new one with agreed upon changes

PR #113, opened 2018-02-16
Allow passing of expected number of EPs

thedac:
Cinder has multiple api versions advertised in the Keystone catalog.
The endpoints test assumes 3 endpoints (public, internal, admin). Allow
passing of the expected number of endpoints. In the case of cinder, 6.


gnuoy:
LGTM

PR #116, opened 2018-02-21
Add egress_subnets helper to access egress-subnets on a relation

stub42:
Access egress-subnets on a relation, with fallbacks. Required for charms maintaining IP access control lists that need to support cross model relations.

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -27,6 +27,7 @@ import glob
 import os
 import json
 import yaml
+import re
 import subprocess
 import sys
 import errno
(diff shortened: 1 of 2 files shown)


charmhelpers/core/hookenv.py
```diff
@@ -1220,3 +1221,37 @@ def ingress_address(rid=None, unit=None):
...
+    Returns a stable list of subnets in CIDR format.
+    eg. ['192.168.1.0/24', '2001::F00F/128']
+
+    If egress-subnets is not available, falls back to using the published
+    ingress-address, or finally private-address.
+
+    :param rid: string relation id
+    :param unit: string unit name
+    :side effect: calls relation_get
+    :return: list of subnets in CIDR format. eg. ['192.168.1.0/24', '2001::F00F/128']
```
ajkavanagh: Is this not an opportunity to leverage a type?  Python has ip_address.IPv4Network and ip_address.IPv6Network which, if returned, would add additional information over bare strings that could be used by the consuming functions.  e.g. making decisions on address classes, private networks, etc.

Also, by using the classes, you can leverage additional checking of the classes do on whether the address strings passed on egress are valid.  If the user of the function wants a str, then str(returned_object) gives up the string representation.  See https://docs.python.org/3/library/ipaddress.html (works in Py27 too).

stub42: We could use Python's IP network types, but this would be surprising since all the other IP address relation helpers in hookenv use strings for addresses. Or do you think this is acceptable here, because it is the first(?) helper dealing with subnets rather than single addresses?

charmhelpers/core/hookenv.py
```diff
@@ -1220,3 +1221,37 @@ def ingress_address(rid=None, unit=None):
     settings = relation_get(rid=rid, unit=unit)
     return (settings.get('ingress-address') or
             settings.get('private-address'))
+
+
+def egress_subnets(rid=None, unit=None):
+    """
+    Retrieve the egress-subnets from a relation. This function is to
+    be used on the providing sude of the relation, and provides the
```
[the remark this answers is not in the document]
stub42: Fixed

charmhelpers/core/hookenv.py
ajkavanagh: That's a really good point.  I guess my initial thought is: "let's get more semantic information into the return type."  Your point about the other functions just returning bare strings is relevant though.  I reckon we could have our cake and eat it: I'm assuming that the user, the first time they go looking for the function, will read the docstring; (and if using a "reasonable" editor, it will tell them what the docstring is, and what the function takes and returns), so they'll see it's not a bare string.

Alternatively, when a reader is reading code that uses it, it may not be apparent that it's an object rather than a string, especially since the `__str__()` method will give up a string.

Personally, I'd implement it using the ipaddress classes, but I can understand if you say it violates the principle of least surprise due to the rest of the code not being as type supportive. :)

stub42: (I'm leading towards strings here for legacy charms, and using proper types in charms.reactive, adding methods to the Endpoint class)

stub42: ipaddress is not builtin to Python 2.7, and using it would add a new dependency. I don't think it is worth the disruption.

ajkavanagh: It's not?  I seem to have a backport installed; my bad.  In that case, yet, I agree that for legacy charms we may as well stick with the bare strings.  I do like the idea of using in in charms.reactive as that's Py3.

PR #119, opened 2018-02-23
Service catalogue validators to convert to v3

gnuoy:
Update the service catalogue validator methods to accept v2 formatted
data and convert to v3 as necessary and call the correct subsequent
validator.

### DIFF
--- a/charmhelpers/contrib/openstack/amulet/utils.py
+++ b/charmhelpers/contrib/openstack/amulet/utils.py
@@ -51,11 +51,12 @@ ERROR = logging.ERROR
 NOVA_CLIENT_VERSION = "2"
 
 OPENSTACK_RELEASES_PAIRS = [
-     'trusty_icehouse', 'trusty_kilo', 'trusty_liberty',
-     'trusty_mitaka', 'xenial_mitaka', 'xenial_newton',
-     'yakkety_newton', 'xenial_ocata', 'zesty_ocata',
-     'xenial_pike', 'artful_pike', 'xenial_queens',
-     'bionic_queens']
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/amulet/utils.py
```diff
@@ -63,7 +63,34 @@ def __init__(self, log_level=ERROR):
...
+               'region': region,
+               'adminurl': adminurl,
+               'internalurl': internalurl,
+               'publicurl': publicurl,
+               'service_id': service_id}
+
+           """
+        validation_function = self.validate_v2_endpoint_data
+        # 11 => xenial_queens
+        if openstack_release and openstack_release >= 11:
```
javacruft: is it possible to use the class data rather than the literal here?

charmhelpers/contrib/openstack/amulet/utils.py
```diff
@@ -141,7 +168,86 @@ def validate_v3_endpoint_data(self, endpoints, admin_port, internal_port,
...
+                       'region': region.
+                       'publicURL': publicURL,
+                       'internalURL': internalURL
+                   }],
+           }
+
+           """
+        validation_function = self.validate_v2_svc_catalog_endpoint_data
+        # 11 => xenial_queens
+        if openstack_release and openstack_release >= 11:
```
javacruft: same comment as above

gnuoy:
James, I think I have addressed you comments. Also, I have tested with a neutron-gateway trusty-icehouse deploy and a xenial-queens deploy.

PR #127, opened 2018-03-08
core: call relation_get with kwargs in the same order.

### LINKED ISSUE #126: @cached mechanism broken in bionic (python 3.6)
Hi,

The `@cached` mechanism in `core/hookenv.py` is broken. The key for the `cache` dict is created as follows (from https://github.com/juju/charm-helpers/blob/master/charmhelpers/core/hookenv.py#L70) : 

```
key = str((func, args, kwargs))
```

But there's a subtle difference in how `str()` called on a dict (`kwargs`, here) behaves between xenial and bionic.

On xenial : 
```
 >>> a = {'rid': 'general-info:2', 'unit': 'ubuntu/0'}
>>> str(a)
"{'rid': 'general-info:2', 'unit': 'ubuntu/0'}"
>>> b = {'unit': 'ubuntu/0', 'rid': 'general-info:2'}
>>> str(b)
"{'rid': 'general-info:2', 'unit': 'ubuntu/0'}"
>>> str(a) == str(b)
True
```

On bionic : 
```
>>> a = {'rid': 'general-info:2', 'unit': 'ubuntu/0'}
>>> str(a)
"{'rid': 'general-info:2', 'unit': 'ubuntu/0'}"
>>> b = {'unit': 'ubuntu/0', 'rid': 'general-info:2'}
>>> str(b)
"{'unit': 'ubuntu/0', 'rid': 'general-info:2'}"
>>> str(a) == str(b)
False
```

In other words, on xenial, the orders of the keys in the dict doesn't matter, but on bionic it does.

So for example, the `relation_get()` method being cached, if I call `relation_get(rid=foo, unit=bar)` and `relation_get(unit=bar, rid=foo)`, on xenial they will have the same cache key, but on bionic they won't.

This shouldn't result in bugs however, just in slower calls. There is a twist however : some functions modify the object returned by the cached function call, which modify the object in the cache itself.

For example : 
```
@cached
def relation_for_unit(unit=None, rid=None):
    """Get the json represenation of a unit's relation"""
    unit = unit or remote_unit()
    relation = relation_get(unit=unit, rid=rid)
    for key in relation:
        if key.endswith('-list'):
            relation[key] = relation[key].split()
    relation['__unit__'] = unit
return relation
```
(taken from https://github.com/juju/charm-helpers/blob/master/charmhelpers/core/hookenv.py#L466)

This will add a `__unit___` key to the dict that is stored in the cache.

Now consider the following call, taken from https://github.com/juju/charm-helpers/blob/master/charmhelpers/core/services/helpers.py#L119 :
```
reldata = hookenv.relation_get(rid=rid, unit=unit)
```

On xenial, because `str(a) == str(b)` (see the first block of code), this call and `relation_get(unit=unit, rid=rid)` (kwargs in different order) will have the same key in the cache, and so will use the same cached object, containing `__unit__`. However on bionic, because `str(a)!=str(b)`, these 2 calls will not have the same key, and so will not use the same object from the cache. So the `reldata` object above will _never_ have a `__unit__` key in it.

The short term fix I'm going to propose here is to have all calls to `relation_get()` in charmhelpers have the same order for kwargs.

For the long term, this caching technique must be changed. Perhaps using https://docs.python.org/dev/library/functools.html#functools.lru_cache ? I'll leave that up to you.

axinojolais:
Fixes #126

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -430,8 +430,8 @@ def relation_set(relation_id=None, relation_settings=None, **kwargs):
 
 def relation_clear(r_id=None):
     ''' Clears any relation data already set on relation r_id '''
-    settings = relation_get(rid=r_id,
-                            unit=local_unit())
+    settings = relation_get(unit=local_unit(),
+                            rid=r_id)
     for setting in settings:
         if setting not in ['public-address', 'private-address']:
             settings[setting] = None
@@ -640,8 +640,8 @@ def is_relation_made(relation, keys='private-address'):
         for unit in related_units(r_id):
             context = {}
             for k in keys:
-                context[k] = relation_get(k, rid=r_id,
-                                          unit=unit)
+                context[k] = relation_get(k, unit=unit,
+                                          rid=r_id)
             if None not in context.values():
                 return True
     return False
@@ -1217,6 +1217,6 @@ def ingress_address(rid=None, unit=None):
     :side effect: calls relation_get
     :return: string IP address
     """
-    settings = relation_get(rid=rid, unit=unit)
+    settings = relation_get(unit=unit, rid=rid)
     return (settings.get('ingress-address') or
             settings.get('private-address'))
--- a/charmhelpers/core/services/helpers.py
+++ b/charmhelpers/core/services/helpers.py
@@ -116,7 +116,7 @@ class RelationContext(dict):
         ns = self.setdefault(self.name, [])
         for rid in sorted(hookenv.relation_ids(self.name)):
             for unit in sorted(hookenv.related_units(rid)):
-                reldata = hookenv.relation_get(rid=rid, unit=unit)
+                reldata = hookenv.relation_get(unit=unit, rid=rid)
                 if self._is_ready(reldata):
                     ns.append(reldata)
 


axinojolais:
kwargs can have nested dicts, so I didn't find a way to cover all possible cases :| This is the sanest solution I could come up with that fixes my immediate problem. I'm fully aware this is not a complete fix. But it works, and doesn't break stuff :)

johnsca:
I think the correct fix would be to change how the cache key is generated.  I propose:

```python
key = json.dumps((str(func), args, kwargs), sort_keys=True)
```

axinojolais:
@ajkavanagh by the way, sorry, my first comment was not clear. As I said at the end of the related issue, this is intended to be a partial, short-term fix for the symptoms which impact me.

I'll test @johnsca's changes tomorrow. Thanks for the quick reply !

johnsca:
One potential issue with my suggestion would be if a value is passed in that's not JSON-serializable.  But I don't think that is too much of a concern considering this is really only used to save on subprocess calls, so we shouldn't be passing in anything crazy.

I would be +1 to the `functools.lru_cache` suggestion, except that it's not available in Python 2, which charm-helpers still supports, AFAIK.

johnsca:
We should also fix the issue with modified return values by having `@cached` return a `copy.deepcopy()` of the cache value instead.

ajkavanagh:
I think the `json.dumps(..., sort_keys=True)` is the correct approach; we've used it extensively in relations to get around dictionary key order changing frequently when serialising to strings. The `deepcopy()` is an interesting issue as the thing being cached might not be deep-copyable?  (I honestly can't think of an example, though!)

@axinojolais thanks for the clarification; I understand a short-term solution, but we can't put that into the charmhelpers library ... that should be patched into the charm if a very short term solution is needed, until it gets fixed here.  Then the next sync/build would then (hopefully) pull a version of c-h that works.

johnsca:
@ajkavanagh I think in this case, the only return values being cached are parsed outputs from the Juju CLI, so we should be fine to `deepcopy` them.

ajkavanagh:
@johnsca Yes, I think that's generally true, too.  +1 on deepcopy().

stub42:
Fixing the cache keys can be done easily and now. Returning a deepcopy of the results is potentially breaking and should be a separate PR.

axinojolais:
I just tested and can confirm that using
```
key = json.dumps((str(func), args, kwargs), sort_keys=True)
```
fixes the problem I'm having.

johnsca:
Closing in favor of #132

axinojolais:
Heh, was about to update my PR with your changes. Thanks !

PR #130, opened 2018-03-13
HAProxyContext: deploy configs that work at boot time on trusty

jetpackdanger:
admin.sock was added to the default config file so that haproxy
would always start, even without any listeners defined, which
prevents systemd from marking the unit as failed[1].  On xenial,
/var/run/haproxy is created at boot time by the init script,
since /var/run (really /run) is a tmpfs.  On trusty, this seems
to work, because the charm also creates /var/run/haproxy, but
since the trusty init script does not create /var/run/haproxy,
following a reboot, haproxy will not start until charm hooks have
run and created the missing directory.  Since the admin socket
does not otherwise exist on trusty haproxy deploys, and we already
create a stats listener on 127.0.0.1:8888, the simplest thing to
do is to only add the directive if may be expected or needed.

Fixes: https://bugs.launchpad.net/charm-helpers/+bug/1755061
Footnotes:
 1: https://anonscm.debian.org/cgit/pkg-haproxy/haproxy.git/commit/?id=8b77d5e324bb11466a3e10cd4885234f9cbf8a61

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -62,6 +62,7 @@ from charmhelpers.core.host import (
     lsb_release,
     CompareHostReleases,
     is_container,
+    init_is_systemd,
 )
 from charmhelpers.contrib.hahelpers.cluster import (
     determine_apache_port,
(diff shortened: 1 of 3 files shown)


javacruft:
Should be fixed in the haproxy packaging in trusty rather than the charm (as discussed in the bug report)

PR #135, opened 2018-03-16
Fix cached not JSON serializable errors

thedac:
After the change at 2aab46ca4ae95c6c5ec8d28b3b39b768f034c179, OpenStack
charms are failing with errors similar to this:
TypeError: <charm.openstack.designate.DesignateConfigurationAdapter
object at 0x7f7e00f83c88> is not JSON serializable

This change guards the json.dumps in a try/except block. Do the former
behavior if json.dumps fails.

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -68,7 +68,10 @@ def cached(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         global cache
-        key = json.dumps((str(func), args, kwargs), sort_keys=True)
+        try:
+            key = json.dumps((str(func), args, kwargs), sort_keys=True)
+        except TypeError:
+            key = str((func, args, kwargs))
         try:
             return cache[key]
         except KeyError:


charmhelpers/core/hookenv.py
```diff
@@ -68,7 +68,10 @@ def unit_get(attribute):
     @wraps(func)
     def wrapper(*args, **kwargs):
         global cache
-        key = json.dumps((str(func), args, kwargs), sort_keys=True)
+        try:
+            key = json.dumps((str(func), args, kwargs), sort_keys=True)
+        except TypeError:
+            key = str((func, args, kwargs))
```
ajkavanagh: We should probably modify the test at tests/core/test_hookenv.py +1176 to activate this piece of code?  i.e. pass something that isn't JSON serialisable and check that it still works as intended.

We also have the odd, but perhaps almost impossible to hit, potential issue of the json serialisation of an object being the same as a str-able set of parameters and thus different things ending up being cached.  Probably not something to worry about!

thedac: I'll fix up the test on Monday. We can hold till then. At the time I thought this was a field emergency. But the workaround was acceptable.

johnsca:
I proposed #136 as an alternative to this PR which preserves the key ordering fix.  It does include the test improvement, as well.

johnsca:
Closing in favor of #136

PR #138, opened 2018-03-21
Handle non-zero unit numbered leader

thedac:
When the distributed_wait function is called and the leader does not
happen to have 0 unit number, the leader winds up waiting while the
non-leader modulo 0 does not. Also when there are units > modulo
there may be conflicts with the leader when multiple units get modulo
0 and do not wait. Neither of these are desired behavior.

This change makes sure the leader, regardless of modulo, never waits.
And forces non-leaders which happen to get modulo 0 to wait.

For applications where the original behavior is desirable,
modulo_distribution can be used directly.

### DIFF
--- a/charmhelpers/contrib/hahelpers/cluster.py
+++ b/charmhelpers/contrib/hahelpers/cluster.py
@@ -386,6 +386,12 @@ def distributed_wait(modulo=None, wait=None, operation_name='operation'):
     if wait is None:
         wait = config_get('known-wait')
     calculated_wait = modulo_distribution(modulo=modulo, wait=wait)
+    if juju_is_leader():
+        # The leader should never wait
+        calculated_wait = 0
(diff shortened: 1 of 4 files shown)


thedac:
Robert,

I updated to use modulo*wait. That makes sense.

Regarding your point about non-consecutive unit numbers, this has always been a workaround to a missing feature in juju. The juju team should be working on a way to do distributed operations per our most recent sprint. Until then, the individual unit will always have to guess because it has no visibility into the whole model.

bbcmicrocomputer:
(As discussed on irc) it makes sense to address the non-sequential unit numbers issue in the modulo_distribution function as a separate set of PRs.

I've raised #142 to track that.

thedac:
Addressed coreycb's requests.

coreycb:
LGTM

PR #141, opened 2018-03-26
storage: Add create_logical_volume helper

javacruft:
Add basic LVM helper for creation of logical volumes.


charmhelpers/contrib/storage/linux/lvm.py
```diff
@@ -151,3 +151,31 @@ def extend_logical_volume_by_device(lv_name, block_device):
     '''
     cmd = ['lvextend', lv_name, block_device]
     check_call(cmd)
+
+
+def create_logical_volume(lv_name, volume_group, size=None):
+    '''
```
ChrisMacNaughton: If we can raise (subprocess.CalledProcessError) I believe that we should document that here

PR #147, opened 2018-04-10
Cache config-get data

fnordahl:
The config() function is currently decorated with the @cached
decorator, however, in a typical charm this will not have
much effect as the charm will make every call to config()
with different arguments, subsequently bypassing the cache.

This commit caches the output from config-get --all and
returns the Config object or each individual value the
caller might ask for.

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -354,21 +354,30 @@ class Config(dict):
             self.save()
 
 
-@cached
+_cache_config = None
+
+
 def config(scope=None):
-    """Juju charm configuration"""
-    config_cmd_line = ['config-get']
-    if scope is not None:
-        config_cmd_line.append(scope)
-    else:
-        config_cmd_line.append('--all')
-    config_cmd_line.append('--format=json')
+    """
+    Get the juju charm configuration (scope==None) or individual key,
+    (scope=str).  The returned value is a Python data structure loaded as
+    JSON from the Juju config command.
+
+    :param scope: If set, return the value for the specified key.
+    :type scope: Optional[str]
+    :returns: Either the whole config as a Config, or a key from it.
+    :rtype: Any
+    """
+    global _cache_config
+    config_cmd_line = ['config-get', '--all', '--format=json']
     try:
-        config_data = json.loads(
-            subprocess.check_output(config_cmd_line).decode('UTF-8'))
+        if _cache_config is None:
+            config_data = json.loads(
+                subprocess.check_output(config_cmd_line).decode('UTF-8'))
+            _cache_config = Config(config_data)
         if scope is not None:
-            return config_data
-        return Config(config_data)
+            return _cache_config.get(scope)
+        return _cache_config
     except ValueError:
         return None
 
--- a/tests/core/test_hookenv.py
+++ b/tests/core/test_hookenv.py
@@ -318,15 +318,18 @@ class HelpersTest(TestCase):
             hookenv.log('foo', level)
             mock_call.assert_called_with(['juju-log', '-l', level, 'foo'])
 
+    @patch('charmhelpers.core.hookenv._cache_config', None)
+    @patch('charmhelpers.core.hookenv.charm_dir')
     @patch('subprocess.check_output')
-    def test_gets_charm_config_with_scope(self, check_output):
-        config_data = 'bar'
-        check_output.return_value = json.dumps(config_data).encode('UTF-8')
+    def test_gets_charm_config_with_scope(self, check_output, charm_dir):
+        check_output.return_value = json.dumps(dict(baz='bar')).encode('UTF-8')
+        charm_dir.return_value = '/nonexistent'
 
         result = hookenv.config(scope='baz')
 
         self.assertEqual(result, 'bar')
-        check_output.assert_called_with(['config-get', 'baz', '--format=json'])
+        check_output.assert_called_with(['config-get', '--all',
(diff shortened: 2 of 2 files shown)


fnordahl:
This commit shaves off 2 - 5 minutes on a simple clustered Keystone deploy, some statistically insignificant data:

Model contains 3-node keystone cluster, mysql and two instances of glance.

Deployment runtime measured on non-leader unit from execution of first hook after `install` hook.

Deployment deemed completed on execution of first `update-status` hook without any subsequent executions of `cluster-relation-changed` hooks.

Baseline: unmodified stable/18.02, juju model-config logging-config=<root>=WARNING;unit=DEBUG
1. run
  * start `2018-04-10 11:01:21`
  * end   `2018-04-10 11:21:38`
  * time  20m17s (1217s)
2. run
  * start `2018-04-10 11:56:07`
  * end   `2018-04-10 12:22:04`
  * time  26m57s (1617s)
 3. run
  * start `2018-04-10 12:35:18`
  * end   `2018-04-10 12:58:47`
  * time  23m29s (1409s)
 
 Modified stable/18.02 (this change) link to branch, juju model-config logging-config=<root>=WARNING;unit=DEBUG
 1. run
   * start `2018-04-10 13:18:47`
   * end   `2018-04-10 13:36:23`
   * time  18m24s (1104s)
 2. run
   * start `2018-04-10 13:47:16`
   * end   `2018-04-10 14:09:28`
   * time  22mN12 (1332s)
 3. run
   * start `2018-04-10 14:20:25`
   * end   `2018-04-10 14:34:36`
   * time  14mN11 (850s)

charmhelpers/core/hookenv.py
```diff
@@ -50,6 +50,7 @@
 MARKER = object()
 
 cache = {}
+cache_config = None
```
ajkavanagh: This should be moved nearer to the "def config" -- otherwise, it's not really clear what it is.

I'd also name is `_cache_config` to make it clear that it is local.

charmhelpers/core/hookenv.py
```diff
@@ -357,18 +358,16 @@ def _implicit_save(self):
 @cached
```
ajkavanagh: I'd drop this decorator as it isn't technically needed any more.

charmhelpers/core/hookenv.py
```diff
@@ -357,18 +358,16 @@ def _implicit_save(self):
...
-        config_cmd_line.append(scope)
-    else:
-        config_cmd_line.append('--all')
-    config_cmd_line.append('--format=json')
+    global cache_config
+    config_cmd_line = ['config-get', '--all', '--format=json']
     try:
-        config_data = json.loads(
-            subprocess.check_output(config_cmd_line).decode('UTF-8'))
+        if not cache_config:
```
ajkavanagh: This needs to be `if cache_config is None:`

This stops the false/true check for "{}" and makes it explicit that this will only happen once (unless json.loads fails, that is)

charmhelpers/core/hookenv.py
```diff
@@ -357,18 +358,16 @@ def _implicit_save(self):
 @cached
 def config(scope=None):
     """Juju charm configuration"""
```
ajkavanagh: Opportunity to fill out this docstring.  What is scope (it's a str) and what does the function return depending on what scope is if it is set.

fnordahl:
@ajkavanagh Thank you Alex, I have updated the PR on basis of your feedback.

charmhelpers/core/hookenv.py
```diff
@@ -354,21 +354,27 @@ def _implicit_save(self):
...
-    config_cmd_line = ['config-get']
-    if scope is not None:
-        config_cmd_line.append(scope)
-    else:
-        config_cmd_line.append('--all')
-    config_cmd_line.append('--format=json')
+    """
+    Juju charm configuration
+
+    :param scope: string. (Optional) Return value for specified key
```
ajkavanagh: So we can do these docstring comments as:

```python
def config(scope=None):
    """
    Get the juju charm configuration (scope==None) or individual key, 
    (scope=str).  The returned value is a Python data structured loaded as 
    JSON from the Juju config command.

    :param scope: If Set, return the value for the specified key.
    :type scope: Optional[str]
    :returns: Either the whole config as a Config, or a key from it.
    :rtype: Any
    """
```

This is basically using `typing` to specify the param and return types, and Sphinx renders the docstring in a really nice way.

fnordahl: This is wonderful, I was unaware of how elaborately Sphinx made use of the doc-string for rich formatting. Thank you!

tests/core/test_hookenv.py
```diff
@@ -336,25 +338,25 @@ def test_gets_charm_config_with_scope(self, check_output):
...
-        check_output.return_value = b''
-
-        result = hookenv.config(scope='baz')
+        result = hookenv.config(scope='foo')
 
         self.assertEqual(result, None)
-        check_output.assert_called_with(['config-get', 'baz', '--format=json'])
+        # cache is populated on first call to hookenv.config() in
+        # test_gets_charm_config_with_scope()
+        self.assertFalse(check_output.called)
```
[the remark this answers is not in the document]
fnordahl: Ah, very good point @johnsca and thank you for enlightening me about possibility of patching the global variable directly, had not thought of that.

I can then leave the data used in existing unit tests unmodified which helps validating non-breakage of the change.

Will add a few new to validate the caching behavior too.

PR #152, opened 2018-04-17
Don't error/raise an exception on port already open

hloeung:
In some cases, we have multiple units deployed to the same machine. An
example of this is nrpe subordinates related to different services
deployed to the same machine. In these special situations, open-port
fails per LP#1750490 and as follows:

| ERROR cannot open 5666/tcp (unit "nrpe-ceph/0"): conflicts with existing 5666/tcp (unit "nrpe-ceph-mon/0")

This fix ignores that error and allows hooks to continue to run.

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -657,12 +657,19 @@ def _port_op(op_name, port, protocol="TCP"):
     else:
         _args.append('{}/{}'.format(port, protocol))
     try:
-        subprocess.check_call(_args)
-    except subprocess.CalledProcessError:
+        subprocess.check_output(_args, stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError as e:
         # Older Juju pre 2.3 doesn't support ICMP
         # so treat it as a no-op if it fails.
-        if not icmp:
-            raise
+        if icmp:
+            return
+        # Work around LP#1750490. We also can't use 'opened-ports' because
+        # that doesn't include ports for other units per LP#1427770.
+        if 'conflicts with existing' in str(e.output):
+            log('Port {}/{} already open, continuing'.format(port, protocol))
+            return
+        log('Error opening port {}/{}'.format(port, protocol))
+        raise
 
 
 def open_port(port, protocol="TCP"):


stub42:
Ignoring the error pushes the problems further down the track. Now the service being managed by the charm may fail since it will attempt to start bound to a port already in use. Now the charm can later close the port, breaking the co-hosted units which think the port is open.

I think this problem would be better handled by making open_port return a more informative error message, so the charm is able to handle the issue as it chooses. Most charms will not be able to handle port conflicts, so it is better if they fail sooner. In nrpe's case, it can ignore an open-port already-open failure, assume that any other units dealing with port 5666 are also nrpe units (and not, say, Apache with weird config), and that charms are never removed once deployed and the close-port races can be ignored.

hloeung:
Okay, these points all make sense. I'll close this PR off and sit on this for now.

BlackDex:
Oke, why is this closed? This is an issue which still exists as far as i know.
Running nova-compute and ceph-osd on the same host and creating a relation between nrpe and nova-compute and nrpe and ceph-osd will cause this same error since the port is already opened by one of the subs.

I now have to choose if i want it to link to nova-compute or ceph-osd.

stub42:
It is closed because two charms on the same unit both attempting to control the same port is a conflict. If open-port silently succeeded, it just defers the problem until later when one charm closes the port and breaks the other. The charms need to be redesigned to cooperate if they need to be smashed together into the same container, or better yet reworked so they don't need to smashed together and they all get a clean container.

BlackDex:
This is not possible for some stuf.
ceph-osd's cant be in a container even as nova-compute.
And i think it is wrong to have the nova-compute and ceph-osd charm to change stuff since nrpe needs to be linked to one ore the other.
What if there is a nova-compute system which isn't an ceph-osd system.
If i only link it to ceph-osd that specific system would not have nrpe checks.

I think the NRPE charm should check if there is already an other NRPE charm running, if so, it should just add its config according to the parent charm it is linked to.

i agree that just ignoring that port 5666 is already open is a bad idea since you then do not check if that is used for nrpe or something else.

Letting the parent charms fix this is not the solution, since that is not how subordinates should work i think.

ajkavanagh:
@BlackDex I still think it's an anti-pattern to have two charms (subordinate or not) try to control the same port.  If we have a situation where we want to have both ceph-osd and nova-computer and their nrpe subordinates on the same host, I'd probably argue that the nrpe charms need to be modified.  The simplest is to modify the charms to try to open the ports and catch the error if it doesn't work and log that the port is probably open.  i.e. the charm acknowledges that there might be an error because other subs also want to open the port, and handle it there in the charm.

BlackDex:
@ajkavanagh I totally agree about two charms controlling the same port. It is good that this causes an error to prevent unexpected behavior.

But in this case it is the same subordinate application just wanting to add more checks to it if there already is an other subordinate charm installed.
I can imagine that this would also take some special care for removing the relation between nrpe and it's master. Since it should not uninstall anything of nrpe, since there is still an other nrpe subordinate running.

I think the NRPE is kind of a different/unique subordinate-charm in this case, since i can't kinda think of any other charm which has the same use-case.

PR #154, opened 2018-04-18
Moved existing interface check out of add_ovsbridge_linuxbridge

mskalka:
Restructures how `add_ovsbridge_linuxbridge` checks for an existing interface and ensures that the OVS bridge is actually created. Also adds a pass-through for the `promisc` arg.

This allows `interface_exists` to be called outside of `add_ovsbridge_linuxbridge`, making it useful for other charms which already import the OVS functions, like Neutron Gateway. 

This is part of a larger push to allow the Neutron Gateway charm to utilize `add_ovsbridge_linuxbridge` to hook into an existing linux bridge for its `data-port` config option, which is desirable for many deployments of the charm where the number of network interfaces is limited.

### DIFF
--- a/charmhelpers/contrib/network/ovs/__init__.py
+++ b/charmhelpers/contrib/network/ovs/__init__.py
@@ -77,30 +77,22 @@ def del_bridge_port(name, port):
     subprocess.check_call(["ip", "link", "set", port, "promisc", "off"])
 
 
-def add_ovsbridge_linuxbridge(name, bridge):
+def add_ovsbridge_linuxbridge(name, bridge, promisc=False):
     ''' Add linux bridge to the named openvswitch bridge
     :param name: Name of ovs bridge to be added to Linux bridge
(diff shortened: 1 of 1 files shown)


charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -77,30 +77,22 @@ def del_bridge_port(name, port):
...
-            apt_install('python3-netifaces', fatal=True)
-        import netifaces
 
     ovsbridge_port = "veth-" + name
     linuxbridge_port = "veth-" + bridge
     log('Adding linuxbridge {} to ovsbridge {}'.format(bridge, name),
         level=INFO)
-    interfaces = netifaces.interfaces()
-    for interface in interfaces:
-        if interface == ovsbridge_port or interface == linuxbridge_port:
```
johnsca: This seems to remove the check for `linuxbridge_port` already existing.  Is that intended?

PR #155, opened 2018-04-20
core/services : fix handling of ports

axinojolais:
We need to turn the generator into a list since the code iterates through it multiple times

### DIFF
--- a/charmhelpers/core/services/base.py
+++ b/charmhelpers/core/services/base.py
@@ -307,7 +307,9 @@ class PortManagerCallback(ManagerCallback):
     """
     def __call__(self, manager, service_name, event_name):
         service = manager.get_service(service_name)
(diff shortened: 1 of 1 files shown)


charmhelpers/core/services/base.py
```diff
@@ -307,7 +307,9 @@ class PortManagerCallback(ManagerCallback):
     """
     def __call__(self, manager, service_name, event_name):
         service = manager.get_service(service_name)
-        new_ports = service.get('ports', [])
+        # turn this generator into a list,
+        # as we'll be going over it multiple times
+        new_ports = [port for port in service.get('ports', [])]
```
stub42: list(service.get('ports', [])) works too

axinojolais:
Yeah list() is better I'll amend my commit

PR #156, opened 2018-04-20
Make get_os_codename_install_source() independent of the series where  it's executed

freyes:
When upgrading the ubuntu series (e.g from Trusty to Xenial)
get_os_codename_install_source() will try to figure out the OpenStack
version splitting the value in openstack-origin (or source) by the
ubuntu series, for cases where the configured value is trusty-mitaka and
the series running is xenial this split gets broken, spliting by '-'
make it reliable.

Fixes LP: #1765805


freyes:
I haven't tested this extensively yet, so far only in a ceph-osd unit upgrading it from trusty to xenial, but I would like to get feedback on this approach, which seems to be harmless, but I may be missing some historical context of why this is being done the way it is. According to `git blame` this hasn't changed since 2013.

PR #161, opened 2018-04-24
Add logging of ValueError Exception in config()

### LINKED ISSUE #160: config() function should log message from JSON parser when parsing fails
Debugging charm issues with reading config from juju config-get is currently hard as there is no log message for the failure event.

Example Traceback from hacluster charm:
```
2018-04-24 12:03:01 INFO config-changed Traceback (most recent call last):
2018-04-24 12:03:01 INFO config-changed   File "/var/lib/juju/agents/unit-hacluster-rados-1/charm/hooks/config-changed", line 481, in <module>
2018-04-24 12:03:01 INFO config-changed     hooks.execute(sys.argv)
2018-04-24 12:03:01 INFO config-changed   File "/var/lib/juju/agents/unit-hacluster-rados-1/charm/hooks/charmhelpers/core/hookenv.py", line 800, in execute
2018-04-24 12:03:01 INFO config-changed     self._hooks[hook_name]()
2018-04-24 12:03:01 INFO config-changed   File "/var/lib/juju/agents/unit-hacluster-rados-1/charm/hooks/config-changed", line 157, in config_changed
2018-04-24 12:03:01 INFO config-changed     update_nrpe_config()
2018-04-24 12:03:01 INFO config-changed   File "/var/lib/juju/agents/unit-hacluster-rados-1/charm/hooks/config-changed", line 450, in update_nrpe_config
2018-04-24 12:03:01 INFO config-changed     nrpe_setup = nrpe.NRPE(hostname=hostname)
2018-04-24 12:03:01 INFO config-changed   File "/var/lib/juju/agents/unit-hacluster-rados-1/charm/hooks/charmhelpers/contrib/charmsupport/nrpe.py", line 244, in __init__
2018-04-24 12:03:01 INFO config-changed     self.nagios_context = self.config['nagios_context']
2018-04-24 12:03:01 INFO config-changed TypeError: 'NoneType' object has no attribute '__getitem__'
```

It is here pretty clear that the following events occur:
- config-get produces invalid output
- This causes a subsequent json.loads() call to raise a ValueError
- This causes the config() function to return None
- Which results in the Traceback since the code in charmhelpers/contrib/charmsupport/nrpe.py attempts to use None as a Dictionary.

However when the user ran the config-get manually the output looks just fine. Having a log call sharing the details of why the ValueError was raised from json.loads() would help immensely in these types of issues.

fnordahl:
Debugging charm issues with reading config from juju config-get is
currently hard as there is no log message for the failure event.

Fixes #160


charmhelpers/core/hookenv.py
```diff
@@ -370,6 +370,12 @@ def config(scope=None):
     """
     global _cache_config
     config_cmd_line = ['config-get', '--all', '--format=json']
+    try:
```
ajkavanagh: This looks a little ugly:  Why not just do

```python
exc_json = json.decoder.JSONDecodeError if six.PY3 else ValueError
```

charmhelpers/core/hookenv.py
```diff
@@ -378,7 +384,10 @@ def config(scope=None):
         if scope is not None:
             return _cache_config.get(scope)
         return _cache_config
-    except ValueError:
+    except (exc_json, UnicodeDecodeError) as e:
+        log('Unable to parse output from config-get: config_cmd_line="{}" '
+            'message="{}"'
+            ''.format(config_cmd_line, str(e)), level=ERROR)
```
ajkavanagh: You don't need the `''` before the `.format(...)`

tests/core/test_hookenv.py
```diff
@@ -362,6 +362,67 @@ def test_gets_config_without_scope(self, check_output, charm_dir):
...
+
+        result = hookenv.config(scope='invalid')
+
+        self.assertEqual(result, None)
+        cmd_line = ['config-get', '--all', '--format=json']
+        check_output.assert_called_with(cmd_line)
+        log.assert_called_with(
+            'Unable to parse output from config-get: '
+            'config_cmd_line="{}" message="{}"'
+            ''.format(str(cmd_line),
```
ajkavanagh: The `''` as above.

charmhelpers/core/hookenv.py
fnordahl: Indeed, however it seems like the exception thrown by json changed from ValueError to JSONDecodeError between Python3.4 and Python3.5.

The PY3 test suite run on Travis uses Python3.4

Thus a binary test for PY2 vs. PY3 is not enough.

charmhelpers/core/hookenv.py
fnordahl: Thx!

tests/core/test_hookenv.py
fnordahl: Thx!

PR #163, opened 2018-04-26
Update pool creation to set app-name

ChrisMacNaughton:


javacruft:
Approach looks OK - would be nice to see a negative test against the ceph_version check.

Some unit test failures against py3.

PR #171, opened 2018-05-08
vaultlocker: Use secret_id's

javacruft:
Enable use of approle secret_id's with vaultlocker; the actual
secret_id is response-wrapped by the vault charm, and is accessed
directly by the consuming application unit using a one-shot token
with a 1hr TTL.

This ensures the secret part of the approle usage is not passed
via charm relation data.

### DIFF
--- a/charmhelpers/contrib/openstack/vaultlocker.py
+++ b/charmhelpers/contrib/openstack/vaultlocker.py
@@ -21,6 +21,7 @@ import charmhelpers.contrib.openstack.context as context
 import charmhelpers.core.hookenv as hookenv
 import charmhelpers.core.host as host
 import charmhelpers.core.templating as templating
+import charmhelpers.core.unitdata as unitdata
 
 VAULTLOCKER_BACKEND = 'charm-vaultlocker'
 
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/vaultlocker.py
```diff
@@ -48,11 +52,30 @@ def __call__(self):
                     unit=unit,
                     rid=relation_id
                 )
+                token = hookenv.relation_get(
+                    '{}_token'.format(hookenv.local_unit()),
+                    unit=unit,
+                    rid=relation_id
+                )
+                if vault_url and role_id and token:
```
gnuoy: I think all([vault_url, role_id, token]) would be neater.

charmhelpers/contrib/openstack/vaultlocker.py
```diff
@@ -48,11 +52,30 @@ def __call__(self):
                     unit=unit,
                     rid=relation_id
                 )
+                token = hookenv.relation_get(
```
gnuoy: There are three separate relation_gets here, why not retrieve all the relation data into a dict and then access that?

ajkavanagh:
Contexts are doing a lot work during initialisation due to the way we contexts are initialised at load time.  I'm wondering if we can do something with the base class to make it lazier and not evaluate the context until it's actually accessed?

PR #173, opened 2018-05-09
Add support for certs relation in OpenStack charms

gnuoy:
* Update https() to check the certificates relation to turn on https.
* Add charmhelpers.contrib.openstack.cert_utils for handeling certificates
  management using the certificates relation
* Update ApacheSSLContext to check for certificates relation. If detected
  create the context but avoid the side effects.


charmhelpers/contrib/openstack/cert_utils.py
```diff
@@ -0,0 +1,229 @@
...
+                # There is network address with no corresponding hostname.
+                # Add the ip to the hostname cert to allow for this.
+                req.add_hostname_cn_ip(addresses)
+        except NoNetworkBinding:
+            log("Skipping request for certificate for ip in {} space, no "
+                "local address found".format(net_type), WARNING)
+    return req.get_request()
+
+
+def create_ip_cert_links(ssl_dir, custum_hostname_link=None):
```
javacruft: custom_hostname_link? type methinks

charmhelpers/contrib/openstack/cert_utils.py
```diff
@@ -0,0 +1,229 @@
...
+    """Create symlinks for SAN records
+
+    :param ssl_dir: str Directory to create symlinks in
+    :param custum_hostname_link: str Additional link to be created
```
javacruft: Same as above

javacruft:
Unit tests also need some work still

charmhelpers/contrib/openstack/context.py
```diff
@@ -871,23 +872,31 @@ def __call__(self):
         if not self.external_ports or not https():
             return {}
 
-        self.configure_ca()
+        use_keystone_ca = True
+        for rid in relation_ids('certificates'):
+            if related_units(rid):
+                use_keystone_ca = False
```
javacruft: Shortcut here might be

use_keystone_ca = not relation_ids('certificates')

I think just checking for relation presence is OK

PR #182, opened 2018-05-22
Add 'ovs-vsctl set' interface function

nickolaev:
The added functions allow direct amnipulation of the OVSDB tables
and setting specific values.
 * set_value - set entries in an arbitrary table
 * set_value_Open_vSwitch - a shortcut to Open_vSwitch table

Signed-off-by: Nikolay Nikolaev <nikolay.nikolaev@canonical.com>

### DIFF
--- a/charmhelpers/contrib/network/ovs/__init__.py
+++ b/charmhelpers/contrib/network/ovs/__init__.py
@@ -134,6 +134,20 @@ def set_manager(manager):
                            'ssl:{}'.format(manager)])
 
 
+def set_Open_vSwitch_column_value(column_value):
+    """
+    Calls ovs-vsctl and sets the 'column_value' in the Open_vSwitch table.
+
+    :param column_value:
+            See http://www.openvswitch.org//ovs-vswitchd.conf.db.5.pdf for
+            details of the relevant values.
+    :type str
+    :raises CalledProcessException: possibly ovsdb-server is not running
+    """
+    log('Setting {} in the Open_vSwitch table'.format(column_value))
+    subprocess.check_call(['ovs-vsctl', 'set', 'Open_vSwitch', '.', column_value])
+
+
 CERT_PATH = '/etc/openvswitch/ovsclient-cert.pem'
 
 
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -134,6 +134,16 @@ def set_manager(manager):
                            'ssl:{}'.format(manager)])
 
 
+def set_value(table, arg1, arg2):
+    ''' Set the controller for the local openvswitch '''
```
ajkavanagh: Please could we use fully formed docstrings.

e.g. in this case (using double ", not single ')

"""Sets a record column=value for a table in the ovs configuration.
    Please see (some manual or reference for what these arguments can be.

    :param table: <what this is>
    :type table: str
    :param arg1: <what it is, or reference where to find them.
    :type arg1: str (???)
    :param arg2: <what is it ...>
    :type arg2: str <....>
    :raises CalledProcessException: <why might ovs-vsctl fail?>
    """

Are 'arg1' and 'arg2' really good names?  I'm thinking `record` and `column_value` might be the closest.  However, it's really a very large API surface, so having this as a general command is probably not useful.

Also, `subprocess.check_call(..)` can raise `CalledProcessException` if the command goes wrong, so it that should also go in the docstring as above.

charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -134,6 +134,16 @@ def set_manager(manager):
...
+    log('Setting table {} {} {}'.format(table, arg1, arg2))
+    subprocess.check_call(['ovs-vsctl', 'set', table, arg1, arg2])
+
+
+def set_value_Open_vSwitch(arg):
```
ajkavanagh: This also needs a docstring; however, if this is all that `set_value` above is being used for, then I'd ditch `set_value(...)` and merge the functionality into this one and just have a specific function.

nickolaev:
I have simplified it to a single function and added the proper docstrings as suggested. Please advice if what should be the next steps.

ajkavanagh:
Thanks @nickolaev, nice patch!

PR #185, opened 2018-06-05
Add "select" function to "MySQLHelper" class

tytus-kurek:
This patchset adds "select" function to "MySQLHelper" class which
can be used to execute SQL select query and return the results.

Closes-Bug: 1775182


charmhelpers/contrib/database/mysql.py
```diff
@@ -163,6 +163,17 @@ def execute(self, sql):
         finally:
             cursor.close()
 
+    def select(self, sql):
+        """Execute arbitrary SQL select query against the database
+         and return the results."""
```
ajkavanagh: Please could we do full doctring comments.  In this case something like:

```python
"""Execute ...

    :param sql: the sql string to execute.
    :type sql: str
    :returns: list of strings? <-- what does it return?
    :rtype: [str]
    :raises: <-- which Execption classes can it raise?
"""

ajkavanagh:
@tytus-kurek unfortunately you've got a pep8 failure which is just white space formatting.

tytus-kurek:
@ajkavanagh sorry for the mess around it :).

ajkavanagh:
Thanks @tytus-kurek for the patch and changes.

PR #186, opened 2018-06-06
ovs: long interface names and existing wiring

javacruft:
Linux interface names are a maximum of 15 characters; ensure that
when wiring an ovs bridge to a linux bridge, the generated veth
pair naming is below this limit by using parts of the hash of
the bridge name.

Any existing use of this feature will remain unchanged as this
is only used when the generated bridge name is > 15 characters.

This commit also fixes an issue when a linuxbridge was previously
directly wired into ovs bridge - this was supported in older
code versions, and we need to ensure we don't break users who
have leveraged this feature on upgrades.

### DIFF
--- a/charmhelpers/contrib/network/ovs/__init__.py
+++ b/charmhelpers/contrib/network/ovs/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 ''' Helpers for interacting with OpenvSwitch '''
+import hashlib
 import subprocess
 import os
 import six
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -92,8 +93,20 @@ def add_ovsbridge_linuxbridge(name, bridge):
             apt_install('python3-netifaces', fatal=True)
         import netifaces
 
-    ovsbridge_port = "veth-" + name
-    linuxbridge_port = "veth-" + bridge
+    if len(bridge) <= 10 and len(name) <= 10:
```
dshcherb: nit:
MAX_KERNEL_INTERFACE_NAME_LEN = 15
VETH_PREFIX = 'veth-'
BRIDGE_NAME_LEN_LIMIT = MAX_KERNEL_INTERFACE_NAME_LEN - len(VETH_PREFIX)
if len(bridge) <= BRIDGE_NAME_LEN_LIMIT and len(name) <= BRIDGE_NAME_LEN_LIMIT:

PR #187, opened 2018-06-06
Expose RabbitMQ port in the context class

freyes:
The nova-consoleauth pacemaker agent verifies if a connection was made
to the rabbitmq port to determine if the process started successfully,
so this new property will be used to configure pacemaker.

Related-Bug: LP: #1753432


ajkavanagh:
@freyes I updated the branch to master (as a precursor to merging the PR), but sadly one of the tests is now failing; you should be able to pull those commits to get it to be the same, and hopefully the fix will be simple.

freyes:
> @freyes I updated the branch to master (as a precursor to merging the PR), but sadly one of the tests is now failing; you should be able to pull those commits to get it to be the same, and hopefully the fix will be simple.

Between my proposal and now a new test was added, I updated that test (the fixture specifically) to include rabbit's port.

charmhelpers/contrib/openstack/context.py
```diff
@@ -489,7 +489,7 @@ def __call__(self):
             ha_vip_only = False
             self.related = True
             transport_hosts = None
-            rabbitmq_port = '5672'
+            rabbitmq_port = 5672
```
ajkavanagh: Why did this change from a string to a number?  I'm worried that, as this is a library function, the change in type may have consequences for other code that might be depending on it being a string?

PR #189, opened 2018-06-20
Add "host" option to "connect" method

tytus-kurek:
This patchset allows specifying the host against which the "connect"
method is executed.

Needed-By: I94710bff17091516875c81ca769d8078ef5efd10
Partial-Bug: 1776171


tytus-kurek:
Could anyone please review and merge it? This is blocking https://review.openstack.org/#/c/576792/

ajkavanagh:
Hi @tytus-kurek, please could you rebase the branch against master.  Also, there's no test for the change, and this would, I think be useful.

charmhelpers/contrib/database/mysql.py
```diff
@@ -78,9 +78,11 @@ def __init__(self, rpasswdf_template, upasswdf_template, host='localhost',
         self.delete_ondisk_passwd_file = delete_ondisk_passwd_file
         self.connection = None
 
-    def connect(self, user='root', password=None):
-        log("Opening db connection for %s@%s" % (user, self.host), level=DEBUG)
-        self.connection = MySQLdb.connect(user=user, host=self.host,
+    def connect(self, user='root', password=None, host=None):
+        if not host:
```
ajkavanagh: As `host` is None in this test, the correct python code is:

```python
if host is not None:
    host = self.host
```

ajkavanagh: This now needs a test, as it's overriding the `self.host`, and thus the contract needs to be enforced in the future.  Thanks.

tytus-kurek: Hi Alex,

Shouldn't it be:

```
if host is None:
    host = self.host
```

so that it can either be specified or defaults to  `self.host` if not specified?

ajkavanagh: @tytus-kurek yes, sorry, that's a typo on my part.  Well caught!  Whilst these are equivalent:

```python
if not host:
    ...

if host is None:
    ...
```

The latter is more correct if we are using `None` as a sentinel to indicate that a parameter has not been passed.

ajkavanagh:
@tytus-kurek please could you also merge or rebase your branch against master.

tytus-kurek:
I am afraid I broke my local git environment. Let me close this PR and create a new one.

PR #192, opened 2018-06-27
Add goal_state_filter_self_relations()

fnordahl:
When Juju `goal-state` lists expected relations the list
includes your own application.

In some cases a charm is only interested in what other
applications will be joining the relation and it is useful
to have a common way of doing that.

### DIFF
--- a/tests/core/test_hookenv.py
+++ b/tests/core/test_hookenv.py
@@ -1371,6 +1371,34 @@ class HelpersTest(TestCase):
         self.assertEqual(result, expect)
         check_output_.assert_called_with(['goal-state', '--format=json'])
 
+    @patch('charmhelpers.core.hookenv.goal_state')
+    @patch('charmhelpers.core.hookenv.application_name')
+    def test_goal_state_filter_self_relations(self, application_name_,
+                                              goal_state_):
+        application_name_.return_value = 'keystone'
+        goal_state_.return_value = {
+            'units': {'keystone/0': {}, 'keystone/1': {}},
+            'relations': {
+                'identity-service': {
+                    'glance': {},
+                    'keystone': {},
+                },
+            },
+        }
+        expect = {
+            'units': {'keystone/0': {}, 'keystone/1': {}},
+            'relations': {
+                'identity-service': {
+                    'glance': {},
+                },
+            },
+        }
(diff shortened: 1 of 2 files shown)


fnordahl:
Indeed @stub42, that sounds reasonable to me, thank you for the feedback.

tests/core/test_hookenv.py
```diff
@@ -1370,6 +1370,32 @@ def test_goal_state(self, check_output_):
...
+        }
+        expect = {
+            'units': {'keystone/0': {}, 'keystone/1': {}},
+            'relations': {
+                'identity-service': {
+                    'glance/0': {},
+                },
+            },
+        }
+        check_output_.return_value = json.dumps(data).encode('UTF-8')
```
stub42: Personally, I think it would have been better to mock charmhelpers.core.hookenv.goal_state instead of subprocess.check_output. The way you have it, the tests are more fragile (eg. someone updating goal_state use different or tweaked subprocess calls would break these tests), and it is more complicated (goal_state.return_value = data, vs. serializing your data structure).

tests/core/test_hookenv.py
```diff
@@ -1370,6 +1370,32 @@ def test_goal_state(self, check_output_):
...
+        result = hookenv.goal_state_filter_self_relations()
+
+        self.assertEqual(result, expect)
+        check_output_.assert_called_with(['goal-state', '--format=json'])
```
stub42: This test isn't necessary, and doesn't belong here. The test is testing the goal_state_filter_self_relations implementation; test of the goal_state implementation should be separate.

tests/core/test_hookenv.py
johnsca: Another approach would be to use functional composition instead and have `filter_self_relations` take the results of `goal_state` as input.  This would actually make the function name more accurate, but would make calling it even more verbose (though, again more explicit and self-documenting).

fnordahl:
Hold merging this until we know the outcome of [bug 1794739](https://bugs.launchpad.net/juju/+bug/1794739)

fnordahl:
Juju will now itself filter yourself from expected applications and units to join a relation in `goal-state` output. Closing this PR as it is obsolete.

PR #195, opened 2018-07-15
write_file: ensure perms always observed

javacruft:
When the content of a file has not changed, write_file
only asserts that the ownership of the file is correct.

Extend this to ensure that permissions changes are also
applied.


johnsca:
Closing in favor of #231

PR #204, opened 2018-08-23
Enable the admin socket and log file for librbd client

taodd:
Need to sync it on related charm openstack project

### DIFF
--- a/charmhelpers/contrib/openstack/templates/ceph.conf
+++ b/charmhelpers/contrib/openstack/templates/ceph.conf
@@ -22,3 +22,5 @@ rbd default features = {{ rbd_features }}
 {{ key }} = {{ value }}
 {% endfor -%}
 {%- endif %}
+log file = /var/log/ceph/$name.$pid.log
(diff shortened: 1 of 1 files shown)


taodd:
looks like cinder-volume use too much rados client. which will generate too much ceph client log file.
close it for now..

taodd:
maybe I'm wrong, I might have misunderstood about the $pid, maybe the $pid is refer to cinder-volume process, not the thread.  I will need to do some experiment.

dosaboy:
Ive tried enabling this in my local env on a cinder node. I then created 100 (rbd) volumes and there were 0 logfiles in /var/log/ceph which suggests that librbd does not support client logging. As a side test i ran an rbd cli command and it created /var/log/ceph/client.admin.550767.log.

taodd:
Two problems
1. AppArmor might stop the cinder/nova/glance creating the admin-socket and log file
2. As cinder and glance should only be using 1 librbd client, so there should be only one log file and admin socket, but for Nova, it should be one instance one log file/admin socket. If there is a host contains many instances, there would be too many of them.

PR #209, opened 2018-09-05
Add helper for apt autoremove

javacruft:


charmhelpers/fetch/ubuntu.py
```diff
@@ -248,6 +248,14 @@ def apt_purge(packages, fatal=False):
     _run_apt_command(cmd, fatal)
 
 
+def apt_autoremove(purge=True, fatal=False):
```
coreycb: Should this optionally allow specifying a package name?

javacruft: apt-get autoremove does not support package names so I don't think so:

```
autoremove (apt-get(8))

   autoremove is used to remove packages that were automatically installed to satisfy dependencies for other packages and are
   now no longer needed as dependencies changed or the package(s) needing them were removed in the meantime.

   You should check that the list does not include applications you have grown to like even though they were once installed just
   as a dependency of another package. You can mark such a package as manually installed by using apt-mark(8). Packages which
   you have installed explicitly via install are also never proposed for automatic removal.
```

PR #212, opened 2018-09-13
Add helpers for common series upgrade tasks

thedac:
For many of the OpenStack API charms the tasks for series upgrade
prepare and complete are the same between charms. These helpers run
those common tasks.


gnuoy:
LGTM

PR #218, opened 2018-09-18
NSG logging: remove unnecessary variable from context

exceptorr:
In scope of https://review.openstack.org/#/c/602780/6/hooks/neutron_ovs_context.py review I'd like to remove unnecessary context variable and make that option configurable via charm config itself.


ChrisMacNaughton:
Why is this PR showing the first commit adding the two relation options, as well as the second one removing one of them?

PR #219, opened 2018-09-18
Remove unnecessary charm relation option

exceptorr:
In scope of https://review.openstack.org/#/c/602780/6/hooks/neutron_ovs_context.py review I'd like to remove unnecessary context variable and make that option configurable via charm config itself.


ChrisMacNaughton:
LGTM

PR #226, opened 2018-10-01
Add helpers to get expected peer and related units from goal-state

fnordahl:


charmhelpers/core/hookenv.py
```diff
@@ -510,6 +510,35 @@ def related_units(relid=None):
...
+    """
+    if not has_juju_version("2.4.4"):
+        # goal-state existed in 2.4.0, but did not list individual units to
+        # join a relation in 2.4.1 through 2.4.3. (LP: #1794739)
+        raise NotImplementedError("goal_state relation unit count")
+    reltype = reltype or relation_type()
+    _goal_state = goal_state()
+    if reltype == 'cluster':
+        return (key for key in _goal_state['units']
+                if '/' in key and key != local_unit())
```
stub42: Where does the magic string 'cluster' come from? It isn't documented in the docstring. It is going to fail in any case for charms with a relation named 'cluster' (like the Cassandra charm - it might be quite a common name for peer relations).

I think it would be better if you remove this special case, and have this method only return units in the specified relation. A separate expected_units function could return all related units.

charmhelpers/core/hookenv.py
```diff
@@ -998,6 +1027,7 @@ def application_version_set(version):
 
 
 @translate_exc(from_exc=OSError, to_exc=NotImplementedError)
+@cached
```
stub42: Good call adding `@cached`

fnordahl:
> I think this should be two functions, one returning the units related on a particular relation, and the other returning the units related on any relation. The current implementation overloads the reltype argument too much (None == current relation, 'cluster' == all relations, other == specific relation), which is unnecessarily confusing and breaks for relations named 'cluster'.

Not really.  The output from `juju goal-state` does not include the peer relation in the `relations` section at all.

You are right though that assuming the peer relation will be named `cluster` is probably an incorrect assumption.

So I guess a separate function for accessing expected units in the peer relation makes sense.

charmhelpers/core/hookenv.py
fnordahl: The intent was to list peer units, which is what the `units` key in the `goal-state` output contains, not all related units.

I do agree that hiding this under the `cluster` key is odd and that this probably warrants a separate function.

All peer relations of the OpenStack charms are called `cluster` which is where my assumption came from.  Thank you for pointing it out.

fnordahl:
Any thoughts on the updated bits @stub42 ?

charmhelpers/contrib/openstack/ha/utils.py
```diff
@@ -110,12 +111,17 @@ def assert_charm_supports_dns_ha():
...
 
     @returns boolean
     """
-    return config('vip') or config('dns-ha')
+    ha_related_units = []
+    try:
+        ha_related_units = list(expected_related_units(reltype='ha'))
+    except (NotImplementedError, KeyError):
+        pass
+    return len(ha_related_units) > 0 or config('vip') or config('dns-ha')
```
stub42: I don't use the contrib/openstack code, and someone on the OpenStack team should probably confirm if this is OK.

PR #228, opened 2018-10-09
Adding "log" support to Neutron API context

exceptorr:
In scope of https://bugs.launchpad.net/charm-neutron-api/+bug/1787397 I'd like to extend Neutron context to let it support adding "log" parameter for `extension_drivers` option.


javacruft:
Please can you add a unit test to test_os_contexts.py to cover this new codepath.

PR #233, opened 2018-10-18
core/host: fix changing permissions in write_file

mhilton:


ajkavanagh:
@mhilton hi, it looks like the test for this needs updating as well, please.

PR #234, opened 2018-10-25
Make the harden and pausable_restart_on_change lazy

ajkavanagh:
Due to py3 work in some charms, these decorators now take callables
in their parameter list as an option, so that load time decorator usage
doesn't have to force a long function call; it makes the parameters lazy
and only evaluated when the decorated function is called.

Also fix a load of linting errors that had been building up.


charmhelpers/contrib/openstack/utils.py
```diff
@@ -1450,20 +1450,33 @@ def some_hook(...):
...
+        # nonlocal keyword
+        __restart_map_cache = {'cache': None}
+
         @functools.wraps(f)
         def wrapped_f(*args, **kwargs):
             if is_unit_paused_set():
                 return f(*args, **kwargs)
+            if __restart_map_cache['cache'] is None:
+                __restart_map_cache['cache'] = restart_map() \
+                    if callable(restart_map) else restart_map
```
thedac: Have we tested this during an OpenStack upgrade?

ajkavanagh: If you think about how the code *used* to work, it was evaluated when the function was *defined*.  Now it just waits till the function is called.  If the issue is that the restart_map might change between when the function is defined (i.e. at load-up) vs when it is called (as part of the hook or a hook), then it might be a problem.  If not, then there is no issue.

If it's between invocations of the hooks then that is a fresh run of the python code, and it would be re-evaluated.

I've updated the .travis.yml as well.

thedac: During an upgrade it is possible that restart_map will be *called* before and after. It also possible the return values will differ. I will need to test this.

ajkavanagh: Okay, that seems fair enough.  We can remove the caching quite easily.  It might have been an optimisation too far! :)

thedac: Agreed. And we can do that RSN. For now, I have tested it on a couple of charms and it does the right thing. I'll merge it.

PR #240, opened 2018-11-13
Add "host" option to "connect" method

tytus-kurek:
This patchset allows specifying the host against which the "connect"
method is executed.

Needed-By: https://review.openstack.org/#/c/576792/
Partial-Bug: 1776171

### DIFF
--- a/tests/contrib/database/test_mysql.py
+++ b/tests/contrib/database/test_mysql.py
@@ -11,9 +11,32 @@ sys.modules['MySQLdb'] = mock.Mock()
 from charmhelpers.contrib.database import mysql  # noqa
 
 
+try:
+    import MySQLdb
+except ImportError:
+    apt_update(fatal=True)
+    if six.PY2:
+        apt_install(filter_installed_packages(['python-mysqldb']), fatal=True)
+    else:
+        apt_install(filter_installed_packages(['python3-mysqldb']), fatal=True)
+    import MySQLdb
+
+
 class MysqlTests(unittest.TestCase):
     def setUp(self):
         super(MysqlTests, self).setUp()
+
+    @mock.patch.object(MySQLdb, 'connect')
+    def test_connect_host_defined(self, mock_connect):
+        helper = mysql.MySQLHelper('foo', 'bar', host='hostA')
+        helper.connect(user='user', password='password', host='1.1.1.1')
+        mock_connect.assert_called_with('user', '1.1.1.1', 'password')
+
+    @mock.patch.object(MySQLdb, 'connect')
+    def test_connect_host_not_defined(self, mock_connect):
+        helper = mysql.MySQLHelper('foo', 'bar', host='hostA')
+        helper.connect(user='user', password='password')
+        mock_connect.assert_called_with('user', 'localhost', 'password')
 
     @mock.patch.object(mysql.MySQLHelper, 'normalize_address')
     @mock.patch.object(mysql.MySQLHelper, 'get_mysql_password')
--- a/charmhelpers/contrib/database/mysql.py
+++ b/charmhelpers/contrib/database/mysql.py
@@ -78,9 +78,11 @@ class MySQLHelper(object):
         self.delete_ondisk_passwd_file = delete_ondisk_passwd_file
         self.connection = None
 
-    def connect(self, user='root', password=None):
-        log("Opening db connection for %s@%s" % (user, self.host), level=DEBUG)
-        self.connection = MySQLdb.connect(user=user, host=self.host,
+    def connect(self, user='root', password=None, host=None):
+        if host is None:
+            host = self.host
+        log("Opening db connection for %s@%s" % (user, host), level=DEBUG)
+        self.connection = MySQLdb.connect(user=user, host=host,
                                           passwd=password)
 
     def database_exists(self, db_name):


tests/contrib/database/test_mysql.py
```diff
@@ -5,16 +5,49 @@
...
 import shutil
 import tempfile
+import six
+
+
+from charmhelpers.fetch import (
+    apt_install,
+    apt_update,
+    filter_installed_packages,
+)
```
ajkavanagh: Please could you use a qualified import rather than bring the names into the module's namespace.  e.g.

```python
import charmhelpers.fetch as fetch

# the use it as
fetch.apt_install(...)
```

It's preferred to do this (generally) as it makes it obvious *where* a function lives.

tests/contrib/database/test_mysql.py
```diff
@@ -5,16 +5,49 @@
...
 
+try:
+    import MySQLdb
+except ImportError:
+    apt_update(fatal=True)
+    if six.PY2:
+        apt_install(filter_installed_packages(['python-mysqldb']), fatal=True)
+    else:
+        apt_install(filter_installed_packages(['python3-mysqldb']), fatal=True)
+    import MySQLdb
```
ajkavanagh: Why are you trying to install MySQLdb -- it's mocked out above in 18 via `sys.modules`?

tytus-kurek: I believe tests in the previous PR were failing when this part was missing. Let me try to re-patch.

tests/contrib/database/test_mysql.py
tytus-kurek: I simply copied lines 44-48 from `charmhelpers/contrib/database/mysql.py` file. Will re-patch.

ajkavanagh: Ah, it's "old style" vs "new style" tensions! :)  Originally, in Python, the preferred way *was* to do "from xyz import abc".  However, more recently (especially in the openstack charms team) we've been pushing for the changed "import xyz" or "import a.b.c.d as e" and then using the name of the module (or its alias) in front of a foreign module function.  This is to help with maintenance really when reading code.  Thanks!

tytus-kurek:
Yup, it's failing now as expected.

tests/contrib/database/test_mysql.py
tytus-kurek: @ajkavanagh so is it fine to leave it as it is, or is there any other way to refer to MySQLdb?

ajkavanagh: This simply should NOT be necessary for a unit test.  The sys.modules[...] on line 18 mocks in the whole of the module.

ajkavanagh:
@tytus-kurek as I can't push to your repo, this is the patch for your test file to solve the import problem:

```
diff --git a/tests/contrib/database/test_mysql.py b/tests/contrib/database/test_mysql.py
index c52bc62a..a10623c7 100644
--- a/tests/contrib/database/test_mysql.py
+++ b/tests/contrib/database/test_mysql.py
@@ -5,44 +5,29 @@ import unittest
 import sys
 import shutil
 import tempfile
-import six
-import charmhelpers.fetch as fetch
 
 
 sys.modules['MySQLdb'] = mock.Mock()
 from charmhelpers.contrib.database import mysql  # noqa
 
 
-try:
-    import MySQLdb
-except ImportError:
-    fetch.apt_update(fatal=True)
-    if six.PY2:
-        fetch.apt_install(fetch.filter_installed_packages(['python-mysqldb']),
-                          fatal=True)
-    else:
-        fetch.apt_install(fetch.filter_installed_packages(['python3-mysqldb']),
-                          fatal=True)
-    import MySQLdb
-
-
 class MysqlTests(unittest.TestCase):
     def setUp(self):
         super(MysqlTests, self).setUp()
 
-    @mock.patch.object(MySQLdb, 'connect')
-    def test_connect_host_defined(self, mock_connect):
+    def test_connect_host_defined(self):
         helper = mysql.MySQLHelper('foo', 'bar', host='hostA')
-        helper.connect(user='user', password='password', host='1.1.1.1')
-        mock_connect.assert_called_with(passwd='password', host='1.1.1.1',
-                                        user='user')
+        with mock.patch.object(mysql, 'log'):
+            helper.connect(user='user', password='password', host='1.1.1.1')
+        mysql.MySQLdb.connect.assert_called_with(
+            passwd='password', host='1.1.1.1', user='user')
 
-    @mock.patch.object(MySQLdb, 'connect')
-    def test_connect_host_not_defined(self, mock_connect):
+    def test_connect_host_not_defined(self):
         helper = mysql.MySQLHelper('foo', 'bar')
-        helper.connect(user='user', password='password')
-        mock_connect.assert_called_with(passwd='password', host='localhost',
-                                        user='user')
+        with mock.patch.object(mysql, 'log'):
+            helper.connect(user='user', password='password')
+        mysql.MySQLdb.connect.assert_called_with(
+            passwd='password', host='localhost', user='user')
 
     @mock.patch.object(mysql.MySQLHelper, 'normalize_address')
     @mock.patch.object(mysql.MySQLHelper, 'get_mysql_password')
```

tytus-kurek:
@ajkavanagh Thank you very much for the Python lesson! It was very useful, seriously.

ajkavanagh:
@tytus-kurek no problem; I was a bit concerned that I didn't understand the issue properly, so decided to have a crack at it, and then thought "I may as well just post it!"  Glad it was useful!

PR #245, opened 2018-11-22
Extract common code to pause/resume services

gnuoy:
Extract common code to pause/resume services into a new function
'manage_payload_services'. This can then be used for any action
that needs to be applied to all services (like stop and start).

### DIFF
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -73,6 +73,8 @@ from charmhelpers.core.host import (
     service_running,
     service_pause,
     service_resume,
+    service_stop,
+    service_start,
     restart_on_change_helper,
 )
 from charmhelpers.fetch import (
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/utils.py
```diff
@@ -1303,6 +1305,55 @@ def is_unit_paused_set():
         return False
 
 
+def manage_payload_services(action, services=None, charm_func=None):
+    """Run an action against all services.
+
+    An optional charm_func() can be called that can either raise an
+    Exception or return non None, None to indicate that the function
```
ajkavanagh: Typo? `non None, None`?

charmhelpers/contrib/openstack/utils.py
```diff
@@ -1303,6 +1305,55 @@ def is_unit_paused_set():
...
+    didn't execute cleanly.
+
+    The signature for charm_func is:
+    charm_func() -> message: string
```
ajkavanagh: Isn't it `charm_func() -> message: str`?

charmhelpers/contrib/openstack/utils.py
```diff
@@ -1303,6 +1305,55 @@ def is_unit_paused_set():
...
+      - None : no services were passed (an empty dict is returned)
+      - a list of strings
+      - A dictionary (optionally OrderedDict) {service_name: {'service': ..}}
+      - An array of [{'service': service_name, ...}, ...]
+
+    @param action: Action to run: pause, resume, start or stop
+    @param services: OPTIONAL see above
+    @param charm_func: function to run for custom charm pausing.
+    @returns Bool, [] Whether succesfull, Status messages
+    @raises Exception(message) on an error for action_fail().
```
ajkavanagh: I think we should be switching to `:param:` and `:type:` and `:rtype:` for our docstrings;  this is what we seem to have adopted elsewhere.  It also will help lead the way to typing at some unspecified point in the future.

charmhelpers/contrib/openstack/utils.py
```diff
@@ -1303,6 +1305,55 @@ def is_unit_paused_set():
...
+    """
+    actions = {
+        'pause': service_pause,
+        'resume': service_resume,
+        'start': service_start,
+        'stop': service_stop}
```
ajkavanagh: For robustness, it's probably worth adding a check like below here, as otherwise the `rc = actions[action](service)` will throw a `KeyError` which is a bit more obtuse.

```python
    action = action.lower()
    if action not in actions.keys():
        raise RuntimeError("action: {} must be on of: {}"
                           .format(action, ', '.join(actions.keys())
```

PR #248, opened 2018-11-24
Handle new juju charm proxy settings and https keyserver URLs

dshcherb:
This change reworks proxy server handling for add-apt-repository and
key retrieval cases and introduces support for new juju-prefixed proxy
setting environment variables that do not modify the http(s) connections
made from hook execution environments by charm code or forked and
execve-ed applications that were given environment variables of a parent
process. The new proxy settings are available as of Juju 2.4.0 but are
properly applied as of Juju 2.4.2 (see lp:1782236).

add-apt-repository comes from the software-properties package and only
reacts to HTTP_PROXY and HTTPS_PROXY environment variables as of bionic
when it was switched to using curl (lp:1433761) and HTTPS-based Ubuntu
keyserver URLs. For Xenial and other releases older than Bionic it is
necessary to use charm options specifying GPG keys in the radix format
AND source URLs in the following formats to avoid triggering the
add-apt-repository behavior related to GPG keys for ppa shortcuts:

deb [arch=<arches-csv] uri distribution [component1] [component2] [...]

See lp:1433761 and https://git.launchpad.net/ubuntu/+source/software-properties/commit/?id=f57935235ca0f52b32da7efe2a24cb26c7fc4573

A manpage for apt-key mentions the following in a section about the
"add" command:
"Note: Instead of using this command a keyring should be placed directly
in the /etc/apt/trusted.gpg.d/ directory with a descriptive name and
either "gpg" or "asc" as file extension."

The support for /etc/apt/trusted.gpg.d/ goes back to 2010:
https://salsa.debian.org/nathanruiz-guest/apt/commit/c24f6ce22cd6720004addad2e3382b3caa6b1b7c

Using "asc" in this directory is only supported as of apt 1.4.

https://salsa.debian.org/nathanruiz-guest/apt/commit/f77ea8235cafb258d1cb0b2b90e95aa36e5c4650
https://salsa.debian.org/nathanruiz-guest/apt/commit/2906182db398419a9c59a928b7ae73cf7c7aa307

Binary GPG format is used in this change given that trusty uses 1.0.1,
xenial uses 1.2.x and only bionic has 1.6.x. This requires de-armoring
of ASCII armor-formatted GPG keys downloaded from the Ubuntu keyserver.

apt-key usage is completely removed by this change.

HTTPS is used for key retrieval with this change which is a functional
change to a more secure way of retrieving GPG keys. A subset of charms
using PPAs will be affected by that.

Since HTTPS is used, if SSLBump-like HTTPS proxies are in place,
they will impersonate keyserver.ubuntu.com and generate a certificate
with keyserver.ubuntu.com in the CN field or in SubjAltName fields of a
certificate. If such proxy behavior is expected it is necessary to add
the CA certificate chain containing the intermediate CA of the SSLBump
proxy to every machine that this code runs on via ca-certs cloud-init
directive (via cloudinit-userdata model-config) or via other means
(such as through a custom charm option). curl relies on openssl provided
by the distribution which means it is affected by the system trusted
certificate store. Also note that DNS resolution for the hostname in a
URL is done at a proxy server - not at the client side.


dshcherb:
Pull-request updated, HEAD is now 936d83af46b1cb740cc88f35c40f81f80885c21d

dshcherb:
Manual functional test environment setup:
https://gist.github.com/dshcherb/adecba7f2f9b7a0aea9f2e6c8c9df3bd

Cases tested manually (landscape server charm, new-style proxy settings, default gateway disabled before charm execution):
- trusty, xenial with the deb-style source (16.06 and 17.03) and key id specified;
- bionic with a ppa shortcut-style 18.03 source;
- xenial and bionic with the deb-style source and ASCII armor key specified via a charm option;

dshcherb:
Pull-request updated, HEAD is now f7cf302e34942c69e73b1c53ea728c14a1e02c39

charmhelpers/fetch/ubuntu.py
```diff
@@ -457,7 +545,55 @@ def _add_apt_repository(spec):
     if '{series}' in spec:
```
ajkavanagh: As this function is being modded, it's a good opportunity to add the `:type spec:` to the docstring, please.

charmhelpers/fetch/ubuntu.py
```diff
@@ -311,35 +311,123 @@ def import_key(key):
...
+    keyserver_url = ('https://keyserver.ubuntu.com'
+                     '/pks/lookup?op=get&options=mr&exact=on&search=0x{}')
+    curl_cmd = ['curl', keyserver_url.format(keyid)]
+    # use proxy server settings in order to retrieve the key
+    return subprocess.check_output(curl_cmd,
+                                   env=_env_proxy_settings(['https']))
+
+
+def _dearmor_gpg_key(key_asc):
+    """
```
ajkavanagh: Start title on same line as `"""`

charmhelpers/fetch/ubuntu.py
```diff
@@ -311,35 +311,123 @@ def import_key(key):
...
+    if 'gpg: no valid OpenPGP data found.' in err.decode('utf-8'):
+        raise GPGKeyError('Invalid GPG key material. Check your network setup'
+                          ' (MTU, routing, DNS) and/or proxy server settings'
+                          ' as well as destination keyserver status.')
+    else:
+        return out
+
+
+def _write_apt_gpg_keyfile(key_name, key_material):
+    """
```
ajkavanagh: Start title on *same* line as `"""` please.

charmhelpers/fetch/ubuntu.py
```diff
@@ -457,7 +545,55 @@ def _add_apt_repository(spec):
...
-    _run_with_retries(['add-apt-repository', '--yes', spec])
+    # software-properties package for bionic properly reacts to proxy settings
+    # passed as environment variables (See lp:1433761). This is not the case
+    # LTS and non-LTS releases below bionic.
+    _run_with_retries(['add-apt-repository', '--yes', spec],
+                      cmd_env=_env_proxy_settings(['https']))
+
+
+def _env_proxy_settings(selected_settings=None):
+    """
```
ajkavanagh: Please change docstring comment to title + description.  Also title on same line as `"""`.

charmhelpers/fetch/ubuntu.py
```diff
@@ -457,7 +545,55 @@ def _add_apt_repository(spec):
...
+    support lowercase or uppercase notation (e.g. curl), some support only
+    lowercase (e.g. wget), there are also subjectively rare cases of only
+    uppercase notation support. no_proxy CIDR and wildcard support also varies
+    between runtimes and applications as there is no enforced standard.
+
+    Some applications may connect to multiple destinations and expose config
+    options that would affect only proxy settings for a specific destination
+    these should be handled in charms in an application-specific manner.
+
+    :param selected_settings: format only a subset of possible settings
```
ajkavanagh: type information for selected_settings please.  It's worth putting an example in the docstring above.

Also, this function returns something, so ':returns:' and ':rtype:' please, and it it's complex, then an example is also useful.

Note as the function returns `None` as an Option, then the :rtype: will be something like:

```
    :rtype: Option(None, dict[str, str])
```

charmhelpers/fetch/ubuntu.py
```diff
@@ -457,7 +545,55 @@ def _add_apt_repository(spec):
...
+    """
+    SUPPORTED_SETTINGS = {
+        'http': 'HTTP_PROXY',
+        'https': 'HTTPS_PROXY',
+        'no_proxy': 'NO_PROXY',
+        'ftp': 'FTP_PROXY'
+    }
+    if not selected_settings:
```
ajkavanagh: Please explicitly check for `None` if using it as a sentinel for "parameter not set by caller". e.g.

```python
if selected_settings is None:
    selected_settings = SUPPORTED_SETTINGS
```

This makes it clear that this piece of code is actually doing the default, rather than if the user passed `{}` for example.

charmhelpers/fetch/ubuntu.py
```diff
@@ -457,7 +545,55 @@ def _add_apt_repository(spec):
...
+        var_val = os.getenv(var)
+        if var_val:
+            proxy_settings[var] = var_val
+            proxy_settings[var.lower()] = var_val
+        # Now handle juju-prefixed environment variables. The legacy vs new
+        # environment variable usage is mutually exclusive
+        charm_var_val = os.getenv('JUJU_CHARM_{}'.format(var))
+        if charm_var_val:
+            proxy_settings[var] = charm_var_val
+            # add a lowercase version as well
```
ajkavanagh: I don't think this comment is needed.  i.e. there's no comment in the above equivalent.

dshcherb:
@ajkavanagh @ryan-beisner 

Thanks for the review and comments, I'll address them and update the review.

dshcherb:
Pull-request updated, HEAD is now 1e464f0b6aff17f0383c938a66303d8e3780f5de

dshcherb:
@ajkavanagh Thanks again for the thorough review, I think the latest revision is a bit cleaner because of that.

dshcherb:
Pull-request updated, HEAD is now 56b3cc19bcc753461750f0ffce550df3d540933c

dshcherb:
Pull-request updated, HEAD is now 08e4fc58416f3476ac56f58d22ca42fa000187ff

dshcherb:
Pull-request updated, HEAD is now a1345c18172be7e5d5263e068f059df92434bd29

dshcherb:
Pull-request updated, HEAD is now e43d24a62cbaf076c471ba028cdd8fb39f682ae4

charmhelpers/fetch/ubuntu.py
```diff
@@ -461,11 +566,62 @@ def _add_apt_repository(spec):
...
+    these should be handled in charms in an application-specific manner.
+
+    :param selected_settings: format only a subset of possible settings
+    :type selected_settings: dict
+    :rtype: Option(None, dict[str, str])
+    """
+    SUPPORTED_SETTINGS = {
+        'http': 'HTTP_PROXY',
+        'https': 'HTTPS_PROXY',
+        'no_proxy': 'NO_PROXY',
```
petevg: juju-no-proxy can contain subnets, which the NO_PROXY environment variable does not support. Are we introducing a bug by passing the value of juju-no-proxy to NO_PROXY without attempting to expand those subnets?

dshcherb: @petevg 

I would rather leave it as-is because the underlying software may or may not support this notation.

add-apt-repository specifically doesn't need no_proxy because it (at worst) connects to two endpoints:

1. launchpad.net to resolve a ppa shortcut to a full deb line;
2. keyserver.ubuntu.com to download a GPG key for a repo to add.

Both of these will be accessed through a proxy in real-life scenarios. If they really need to be accessed without a proxy when proxy settings are configured they can be added to no_proxy as hostnames.

So there is no harm in the fact that curl invoked by add-apt-repository will ignore no_proxy.

The strong argument against expanding no_proxy in every situation is that expanded no_proxy can reach the limits imposed on argument size in terms of a child process stack memory - the decision on whether to expand no_proxy or not should be done by the caller of `_env_proxy_settings`:

https://elixir.bootlin.com/linux/v4.14.96/source/fs/exec.c#L244
```
static struct page *get_arg_page(struct linux_binprm *bprm, unsigned long pos,
		int write)
// ...
		/*
		 * We've historically supported up to 32 pages (ARG_MAX)
		 * of argument strings even with small stacks
		 */
		if (size <= ARG_MAX)
			return page;
// ...
		/*
		 * Limit to 1/4 of the max stack size or 3/4 of _STK_LIM
		 * (whichever is smaller) for the argv+env strings.
		 * This ensures that:
		 *  - the remaining binfmt code will not run out of stack space,
		 *  - the program will have a reasonable amount of stack left
		 *    to work from.
		 */
		limit = _STK_LIM / 4 * 3;
		limit = min(limit, rlimit(RLIMIT_STACK) / 4);
		if (size > limit)
			goto fail;

```
https://elixir.bootlin.com/linux/v4.14.96/source/include/uapi/linux/resource.h#L66
`#define _STK_LIM	(8*1024*1024)`

dshcherb:
Pull-request updated, HEAD is now b9f658ea6fc4774125afc343760bbb900b0d0054

charmhelpers/fetch/ubuntu.py
petevg: @dshcherb Hmmm ... I'm tempted to say that we shouldn't include `NO_PROXY` in the default list of supported settings.

I guess the question is this: if I naively pass the result of this routine to _env_proxy_settings, without specifying any arguments, what is the best way for this to fail?

1) It ignores NO_PROXY by default, which means that I struggle to figure out why it isn't respecting my juju-no-proxy setting.
2) I have something in juju-no-proxy that will cause mischief, and my charm falls over as a result.

petevg: @dshcherb After talking this over with the team, I think that it would be better to handle both the cases where no_proxy has a cidr, and where it is too big inline here.

If juju-no-proxy contains a cidr, attempt to expand it.
If the cidr is now too big, either log a warning, or throw an Exception.

The goal is to make it easy to troubleshoot if something does go wrong. And also to attempt to do the "right" thing in the first place.

dshcherb:
Pull-request updated, HEAD is now ffb32ed6a60e03aede830883a42829ad61aaff05

dshcherb:
Pull-request updated, HEAD is now 805fd90f544961c5176ba41497e38befdbb5b10d

charmhelpers/fetch/ubuntu.py
petevg: @ajkavanagh better now?

dshcherb:
Pull-request updated, HEAD is now f48e3183b563ba077a684bcb230e6913814be1f5

dshcherb:
Rebased to master again. The previous test failure was some kind of timeout during test execution.

```
ReadTimeoutError: HTTPSConnectionPool(host='files.pythonhosted.org', port=443): Read timed out.
The command "pip install -r test_requirements.txt" failed and exited with 2 during .
```

ajkavanagh:
This is a really big piece of work, and nicely done.  However, it would be useful if a functional test of these changes in a charm could be presented as evidence that it does work -- I see the unit tests work, but it's too large a patch for me to determine if it actually works in a charm.  Thanks!

dshcherb:
Pull-request updated, HEAD is now d67fe9ce469660d2eca2ddb521188134bf146e2c

dshcherb:
Pull-request updated, HEAD is now 77b36ddcce190862492dbd83d00458430f6bfc28

dshcherb:
@ajkavanagh 
@petevg 
@ryan-beisner 

I moved env_proxy_settings to "hookenv" instead of "fetch" as this function is generally useful for charms to use and works with a hook context. I actually have a use for it already in a vendor charm that does a `curl` [not minding new proxy settings](https://github.com/Juniper/contrail-charms/blob/R5/contrail-analyticsdb/hooks/docker_utils.py#L19-L26).

Test results with the latest code when only proxy server connectivity is available:

xenial (full source line and a full-text ASCII armor key)
https://gist.github.com/dshcherb/6e1a4fff161e2091e92ed19aeca6edb0#file-proxy-test-landscape-xenial-full-txt

xenial (full source line and a long key id)
https://gist.github.com/dshcherb/9f59874dca12abbf24597647013c1b42

bionic (ppa shortcut and medium key id specified => add-apt-repository to launchpad.net connectivity via a proxy + keyserver via a proxy):
https://gist.github.com/dshcherb/67a9d7ea671c093124aa6e8a37a1e89d#file-proxy-landscape-bionic-ppa-keyid-txt

trusty (full source line and short key id)
https://gist.github.com/dshcherb/0b6fc364b34785b428f0ec4d92b2e2c9

charmhelpers/contrib/openstack/amulet/utils.py
```diff
@@ -96,8 +96,8 @@ def validate_endpoint_data(self, endpoints, admin_port, internal_port,
                 'url': self.valid_url,
                 'interface': self.not_null,
                 'service_id': expected['service_id']}
-        return validation_function(endpoints, admin_port, internal_port,
-                                   public_port, expected)
+            return validation_function(endpoints, admin_port, internal_port,
+                                       public_port, expected)
```
ajkavanagh: This looks like a bug?  Are you sure about the additional indent.  If the `if` fails the function will return `None` rather the the v2 endpoint data?

charmhelpers/core/hookenv.py
```diff
@@ -1414,3 +1419,77 @@ def unit_doomed(unit=None):
...
+    # Test for wildcard domains (*.foo.com or .foo.com)
+    if "*" in addresses:
+        return True
+    if addresses.startswith("."):
+        return True
+    if ",." in addresses:
+        return True
+    if " ." in addresses:
+        return True
+    return False
```
ajkavanagh: This would be much nicer (and easier to comprehend - I think?) if it was:

```python
return (
    "/" in addresses or
    "*" in addresses or
    addresses.startswith(".") or
    ",." in addresses or
    " ." in addresses)
```

Or some variant on that.

dshcherb:
Pull-request updated, HEAD is now 842e5e1f92dd8aa9afeb5b70b31661b0bc372d4e

dshcherb:
Just did a quick retest because I used ~dmitriis/landscape-server-0 which had an older revision of code (with _env_proxy_settings still in the fetch module).

With ~dmitriis/landscape-server-1:

https://gist.github.com/dshcherb/0b6fc364b34785b428f0ec4d92b2e2c9/82b09a52004ee6c74e08b16b82d1f30270790991#file-proxy-test-landscape-trusty-full-keyid-txt

https://gist.github.com/dshcherb/67a9d7ea671c093124aa6e8a37a1e89d#file-proxy-landscape-bionic-ppa-keyid-txt

https://gist.github.com/dshcherb/6e1a4fff161e2091e92ed19aeca6edb0#file-proxy-test-landscape-xenial-full-txt

charmhelpers/contrib/openstack/amulet/utils.py
dshcherb: Thanks! It is a bug indeed.

charmhelpers/core/hookenv.py
dshcherb: Yep, much nicer 👍

PR #253, opened 2018-12-03
Charms can sepcify additional delete & group info

gnuoy:
When a charm calls generate_ha_relation_data allow additional
resources to be deleted and groups to be specified.

### DIFF
--- a/charmhelpers/contrib/openstack/ha/utils.py
+++ b/charmhelpers/contrib/openstack/ha/utils.py
@@ -325,5 +325,5 @@ def update_hacluster_vip(service, relation_data):
             relation_data['groups'][key] = ' '.join(vip_group)
         except KeyError:
             relation_data['groups'] = {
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/ha/utils.py
```diff
@@ -314,9 +314,16 @@ def update_hacluster_vip(service, relation_data):
...
     if len(vip_group) >= 1:
-        relation_data['groups'] = {
-            'grp_{}_vips'.format(service): ' '.join(vip_group)
-        }
+        key = 'grp_{}_vips'.format(service)
+        try:
+            relation_data['groups'][key] = ' '.join(vip_group)
+        except KeyError:
+            relation_data['groups'] = {
+                'grp_{}_vips'.format(service): ' '.join(vip_group)
```
javacruft: you can just use 'key' here rather than re-forming the key again.

PR #265, opened 2019-01-15
check if key is in keyring, if not, update

andrewdmcleod:
First we check if there is already a keyring file, and report if it exists. Then we check if it contains the same key as in config - if not, we update the keyring file.


petevg:
I'm taking over as Andrew is EOD. Going to close out this PR, and create a new one from my own fork.

PR #266, opened 2019-01-15
Fix ceph update keyring

petevg:
Updates to @admcleod's earlier PR, addressing comments, and fixing tests.

### DIFF
--- a/charmhelpers/contrib/storage/linux/ceph.py
+++ b/charmhelpers/contrib/storage/linux/ceph.py
@@ -856,8 +856,14 @@ def _keyring_path(service):
     return KEYRING.format(service)
 
 
-def create_keyring(service, key):
-    """Create or update a Ceph keyring containing key."""
+def insert_key(service, key):
+    """
+    Add a key to a keyring.
+
+    Creates the keyring if it doesn't already exist.
+
+    Logs and returns if the key is already in the keyring.
+    """
     keyring = _keyring_path(service)
     if os.path.exists(keyring):
         with open(keyring) as ring:
@@ -872,6 +878,11 @@ def create_keyring(service, key):
            '--name=client.{}'.format(service), '--add-key={}'.format(key)]
     check_call(cmd)
     log('Created new ceph keyring at %s.' % keyring, level=DEBUG)
+
+
+def create_keyring(service, key):
+    """Deprecated. Please use the more accurately named 'insert_key'"""
+    return insert_key(service, key)
 
 
 def delete_keyring(service):
@@ -910,7 +921,7 @@ def get_ceph_nodes(relation='ceph'):
 
 def configure(service, key, auth, use_syslog):
     """Perform basic configuration of Ceph."""
-    create_keyring(service, key)
+    insert_key(service, key)
     create_key_file(service, key)
     hosts = get_ceph_nodes()
     with open('/etc/ceph/ceph.conf', 'w') as ceph_conf:
@@ -1073,7 +1084,7 @@ def ensure_ceph_keyring(service, user=None, group=None,
     if not key:
         return False
 
-    create_keyring(service=service, key=key)
+    insert_key(service=service, key=key)
     keyring = _keyring_path(service)
     if user and group:
         check_call(['chown', '%s.%s' % (user, group), keyring])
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -856,19 +856,35 @@ def _keyring_path(service):
...
+
+    Logs and returns if the key is already in the keyring.
+    """
     keyring = _keyring_path(service)
     if os.path.exists(keyring):
-        log('Ceph keyring exists at %s.' % keyring, level=WARNING)
-        return
+        with open(keyring) as ring:
+            if key not in ring.read():
+                log('Updating existing keyring %s.' % keyring, level=WARNING)
```
coreycb: Just a nit but I don't know if either of the first 2 log messages should be warnings.

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -856,19 +856,35 @@ def _keyring_path(service):
...
+            else:
+                log('Ceph keyring exists at %s and has not changed.' % keyring,
+                    level=WARNING)
+                return
 
     cmd = ['ceph-authtool', keyring, '--create-keyring',
            '--name=client.{}'.format(service), '--add-key={}'.format(key)]
     check_call(cmd)
     log('Created new ceph keyring at %s.' % keyring, level=DEBUG)
```
coreycb: This log message should probably be updated. It could be misleading if adding a key to an existing keyring.

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -856,19 +856,35 @@ def _keyring_path(service):
...
+
+    Creates the keyring if it doesn't already exist.
+
+    Logs and returns if the key is already in the keyring.
+    """
     keyring = _keyring_path(service)
     if os.path.exists(keyring):
-        log('Ceph keyring exists at %s.' % keyring, level=WARNING)
-        return
+        with open(keyring) as ring:
```
thedac: My py3 concern still stands. Will this be bytes and require decoding?

We should be explicit if this is 'r' or 'rb' and decode if necessary.

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -856,19 +856,35 @@ def _keyring_path(service):
...
+            if key not in ring.read():
```
thedac: Also since ware nit picking: It is clearer to do

if key in ring.read():
  log(" ...
  return

log("Adding key ....
cmd = ...

petevg: Done.

charmhelpers/contrib/storage/linux/ceph.py
petevg: I spun up an openstack base bundle in serverstack. I was able to read all the .keyring files that I found, in python3, with an open("filename", "r"), and check to see if a given string existed in the file. I don't think that we have to worry about unicode encoding. :-)

charmhelpers/contrib/storage/linux/ceph.py
petevg: Changed to DEBUG.

charmhelpers/contrib/storage/linux/ceph.py
petevg: That log message only shows up if the keyring exists *and* the key is in the keyring. So the message in that case is accurate -- we'll perform a noop.

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -856,19 +856,34 @@ def _keyring_path(service):
...
+            if key in ring.read():
+                log('Ceph keyring exists at %s and has not changed.' % keyring,
+                    level=DEBUG)
+                return
+            log('Updating existing keyring %s.' % keyring, level=DEBUG)
 
     cmd = ['ceph-authtool', keyring, '--create-keyring',
            '--name=client.{}'.format(service), '--add-key={}'.format(key)]
     check_call(cmd)
     log('Created new ceph keyring at %s.' % keyring, level=DEBUG)
```
coreycb: I still think this final log is confusing on an update but obviously a minor issue.

Issue #254, opened 2018-12-05
Ought to introduce a helper for "release = lsb_release()['DISTRIB_CODENAME'].lower()"


ajkavanagh:
charm-helpers itself, and lots of charms, use this exact line:

```python
release = lsb_release()['DISTRIB_CODENAME'].lower()
```

There ought to be a helper that does this, and do it through the code base.

Note, that this will also have to be done for the centos host_factory too.

dshcherb:
@ajkavanagh 

The Ubuntu portion:
https://github.com/juju/charm-helpers/pull/268

ajkavanagh:
Thanks @dshcherb for starting this off.  I've left a review on the PR. Thanks!

PR #268, opened 2019-01-27
Introduce get_distrib_codename helper

dshcherb:
Adds a simple helper to return a distribution codename.


charmhelpers/fetch/ubuntu.py
```diff
@@ -621,3 +621,7 @@ def get_upstream_version(package):
         return None
 
     return apt_pkg.upstream_version(pkg.current_ver.ver_str)
+
+
+def get_distrib_codename():
```
ajkavanagh: This is logically in the wrong file.  I think it should be imported from charmhelpers.core.host like `lsb_release()` is, which, yes, does mean implementing it in the host_factory/ubuntu.py and then bubbling it through the imports.

Also, a docstring is needed.  It needs something like:

"""Return the codename of the distribution

:returns: The codename
:rtype: str
"""

ajkavanagh:
@dshcherb (and reminder to merger or self): it ought to have a 'Closes-Bug: #254' in the commit message when its merged to auto-close the bug.

dshcherb:
Pull-request updated, HEAD is now fba1b133cea6b0e29f8e438058a0f8da25b98d1d

ajkavanagh:
Yeah, except it didn't deal with the issue that the rest of the code in the module hasn't been refactored to use the function; a bit of a negative and why I was still -ve on my review ...

dshcherb:
@ajkavanagh

I addressed that in the review about the ubuntu.py rework for gpg and proxy handling. Although this seems to be the right place to do it

ajkavanagh:
@dshcherb could you link to that review from here please.

dshcherb:
https://github.com/juju/charm-helpers/pull/248

I can rebase it on top of this change now (I merged this branch before it was merged in that PR).

PR #273, opened 2019-02-07
Add getrange command to unitdata CLI

johnsca:
Can now use:

```
chlp unitdata getrange foo.
```


johnsca:
Lint failures are fixed in #272 and not relevant to this PR

johnsca:
Hrm.  None of the CLI commands are documented other than in `chlp --help`.  It would be worth adding them to the docs in this PR.  👍

johnsca:
On second thought, adding the CLI commands to the docs is going to be non-trivial, since they rely on argparse to document themselves and it would need a new plugin to convert that to Sphinx docs.  Going to merge this without that, since it can be discovered via the CLI already.

PR #274, opened 2019-02-11
Add OpenStackSecurityGuide auditing

ChrisMacNaughton:
Charms should be able to write assertions against the running
system, in a way that's configurable at runtime, based on the
determined state of the system.


ChrisMacNaughton:
An example of code to use this in a charm's action:

```python
import sys

import charmhelpers.contrib.openstack.audits as audits
import charmhelpers.contrib.openstack.audits.openstack_security_guide

sys.path.append('hooks')

from charmhelpers.core.hookenv import action_fail, action_set


if __name__ == "__main__":
    config = audits.Config(
        config_path='/etc/nova',
        config_file='nova.conf',
        owner='nova',
        group='nova',
        permissions='640',
        audit_type=audits.AuditType.OpenStackSecurityGuide)

    sys.exit(audits.run(config))
```

charmhelpers/contrib/openstack/audits/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def since_package(pkg, pkg_version):
+    """This audit should be run after the specified package version (incl)."""
+    def should_run(config):
+        version = get_upstream_version(pkg)
+        if version:
+            if apt_pkg.version_compare(version, pkg_version) >= 0:
+                return True
+        return False
```
ajkavanagh: This bit of code can be written as a shortcut:

```python
return version and (apt_pkg.version_compare(version, pkg_version) >=0)
```

However, that might be less clear, particularly if someone is not clear that `and` shortcuts and doesn't evaluate the RHS if version is 'False-y'.  Still will return a boolean.

charmhelpers/contrib/openstack/audits/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+import collections
+from enum import Enum
+import traceback
+
+from charmhelpers.fetch.ubuntu import get_upstream_version
+
+from charmhelpers.core.hookenv import (
+    action_fail,
+    action_set,
+)
```
ajkavanagh: I tend to prefer the style of

```python
import charmhelpers.core.hookenv as hookenv

# and then use it as

hookenv.action_fail(...)
```

This is just to signal in code where something came from.  However, I'm not overly hung up on this as the rest of the code in charmhelpers tends to be of this style.

charmhelpers/contrib/openstack/audits/__init__.py
```diff
@@ -0,0 +1,140 @@
...
+    """Run the configured audits with the specified config.
+
+    :param config: Configuration for the audit
+    :type config: Config
+    """
+    errors = {}
+    results = {}
+    for name, audit in sorted(_audits.items()):
+        result_name = name.replace('_', '-')
+        if all(p(config) for p in audit.filters):
```
ajkavanagh: It would probably be sensible to have a test that `p` is `callable` before calling it.  That way you can pick up any non callables before they get to here; it's probably best to put a check in the decorator that the arguments are all callable that let it fail here?

charmhelpers/contrib/openstack/audits/openstack_security_guide.py
```diff
@@ -0,0 +1,216 @@
...
+    Get the Ownership information from a file.
+
+    :param file: The path to a file to stat
+    :type file: str
+    :returns: owner, group, and mode of the specified file
+    :rtype: Ownership
+    :raises subprocess.CalledProcessError: If the underlying stat fails
+    """
+    out = subprocess.check_output(['stat', '-c', '%U %G %a', file]).decode('utf-8')
+    return Ownership(*out.strip().split(' '))
```
ajkavanagh: Nice :)

charmhelpers/contrib/openstack/audits/openstack_security_guide.py
```diff
@@ -0,0 +1,216 @@
...
+    :type owner: str
+    :param group: Name of the group
+    :type group: str
+    :param file_name: Path to the file to verify
+    :type file_name: str
+    """
+    try:
+        ownership = _stat(file_name)
+    except subprocess.CalledProcessError as e:
+        print("Error reading file: {}".format(e))
```
ajkavanagh: `print` or `log`?

charmhelpers/contrib/openstack/audits/openstack_security_guide.py
```diff
@@ -0,0 +1,216 @@
...
+
+    :param mode: file mode that is desires
+    :type owner: str
+    :param file_name: Path to the file to verify
+    :type file_name: str
+    """
+    try:
+        ownership = _stat(file_name)
+    except subprocess.CalledProcessError as e:
+        print("Error reading file: {}".format(e))
```
ajkavanagh: print or log; not sure what the intention is here (i.e. if this is running on a unit, log is preferable so it turns up in the juju debug log?)

ChrisMacNaughton: Fun tidbit - STDOUT from an action goes into the unit's debug log :)

ajkavanagh: nice.

ChrisMacNaughton: Another thing to discuss regarding print or log, would be for testing purposes - if we use log, then we can easily mock out the charmhelpers.log function as well, and validate pass/fail via that method if we wish.

Given that we can also just mock dependencies correctly and have the test assertions (audits) work normally, it doesn't seem like that strong of a reason.

ChrisMacNaughton:
@javacruft @ajkavanagh I'd appreciate another review on this, and probably a merge. The list of files included is now reasonably complete (I believe) and so should be ready to migrate towards validation in the charms.

PR #279, opened 2019-02-18
Update ceph helpers for device class support

javacruft:
Newer ceph versions automatically classify underlying block devices
by class (hdd, sdd, nvme). Provide support in ceph helpers for using
device classes when creating pools and calculating pg sizing.

### DIFF
--- a/charmhelpers/contrib/storage/linux/ceph.py
+++ b/charmhelpers/contrib/storage/linux/ceph.py
@@ -766,7 +766,8 @@ def get_osds(service, device_class=None):
     :param device_class: Class of storage device for OSD's
     :type device_class: str
     """
-    if device_class:
+    luminous_or_later = cmp_pkgrevno('ceph', '12.0.0') >= 0
+    if luminous_or_later and device_class:
         out = check_output(['ceph', '--id', service,
                             'osd', 'crush', 'class',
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -660,6 +669,9 @@ def create_erasure_profile(service, profile_name, erasure_plugin_name='jerasure'
     else:
         cmd.append('ruleset-failure-domain=' + failure_domain)
 
+    if version and version >= '12.0.0' and device_class:
```
ChrisMacNaughton: Can we log on `else` please?

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -744,20 +756,25 @@ def pool_exists(service, name):
...
     """Return a list of all Ceph Object Storage Daemons currently in the
-    cluster.
+    cluster (optionally filtered by storage device class).
+
+    :param device_class: Class of storage device for OSD's
+    :type device_class: str
     """
-    version = ceph_version()
-    if version and version >= '0.56':
+    if device_class:
```
ChrisMacNaughton: What does this command do on older Ceph versions?

PR #284, opened 2019-02-26
Add support for creating erasure coded pool and setting ``max_objects`` quota

fnordahl:


tests/contrib/storage/test_linux_ceph.py
```diff
@@ -1602,3 +1618,13 @@ def test_add_op_create_pool(self):
         op = base_op.copy()
         op['app-name'] = 'earth'
         self.assertEqual(rq.ops, [op])
+        rq = ceph_utils.CephBrokerRq()
+        rq.add_op_create_pool('apool', max_bytes=42)
+        op = base_op.copy()
+        op['max-bytes'] = 42
+        self.assertEqual(rq.ops, [op])
+        rq = ceph_utils.CephBrokerRq()
+        rq.add_op_create_pool('apool', max_objects=42)
```
sahid: It would have been better to use a different value to ensure no mismatch

fnordahl: Thanks for the input @sahid. I certainly agree for when using mock.assert_called_with statements multiple times so you know which iteration of the call you are validating.  But since I am recreating the datastructure each time and using assertEqual I do believe this should be safe.

fnordahl:
Thanks for the quick review @sahid! I added a second change and did a rebase, would you want to review the second commit too?

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -1156,7 +1156,8 @@ def add_op_request_access_to_group(self, name, namespace=None,
 
     def add_op_create_pool(self, name, replica_count=3, pg_num=None,
```
sahid: The param `replica_count` only makes sense for a 'replicated' pool type. So I think we should refactor a bit this part of code to avoid confusion. Basically have two methods `add_op_create_replicate_pool()` and `add_op_create_erasure_pool()`. Or perhaps we could add a param like `pool_type` ?

fnordahl: Yes you are right.  I tend to overload when things are similar, but I really should not. Will push a second iteration.

PR #285, opened 2019-02-27
Common support for smart restart of containerized services

dmzoneill:
In containerized co-lacated services such as apache2 for WSGI openstack
services, a race condition is present for restarting of apache2.

This race condition does not check which is the correct apache2 process
for given container PID namespace.

The orginal fix for this was added in
https://review.openstack.org/#/c/608346/ for keystone.

Standard usage is as follows

from charmhelpers.contrib.openstack.utils import run_in_apache
from charmhelpers.core.host import restart_pid_check

my_list = { 'apache2': restart_pid_check, ... }

@restart_on_change(restart_map(), restart_functions=my_list)


ChrisMacNaughton:
Looks good, but I'd like to see some new unit tests to go along with it

charmhelpers/contrib/openstack/utils.py
```diff
@@ -1841,3 +1841,34 @@ def series_upgrade_complete(resume_unit_helper=None, configs=None):
         configs.write_all()
         if resume_unit_helper:
             resume_unit_helper(configs)
+
+
+def run_in_apache(package, release=None):
```
coreycb: Hey David, I put some comments about this code in https://review.openstack.org/#/c/641003/1 if you can address those. Thanks.

charmhelpers/contrib/openstack/utils.py
```diff
@@ -1841,3 +1841,34 @@ def series_upgrade_complete(resume_unit_helper=None, configs=None):
...
+
+    if (package == 'keystone' and
+            CompareOpenStackReleases(release) >= 'liberty'):
+        return True
+
+    if (package == 'cinder-api' and
+            CompareOpenStackReleases(release) >= 'ocata'):
+        return True
+
+    if (package == 'glance-api' and
```
coreycb: I don't think glance API is run under apache2. Can we drop this and then it's ready to merge from my pov.

charmhelpers/contrib/openstack/utils.py
dmzoneill: hey, these have been addressed

charmhelpers/contrib/openstack/utils.py
dmzoneill: I removed this

dmzoneill:
> I still don't see any new tests to cover this functionality

hey sorry, lost in translation, i copied and modified the original tests in keystone here.

charmhelpers/contrib/openstack/utils.py
```diff
@@ -1841,3 +1841,30 @@ def series_upgrade_complete(resume_unit_helper=None, configs=None):
         configs.write_all()
         if resume_unit_helper:
             resume_unit_helper(configs)
+
+
+def run_in_apache(package, release=None):
```
ajkavanagh: I don't think this function is particularly useful in charm-helpers.  Basically, it's being specialised for each of the charms by package.  These will be simpler, smaller, functions, if they are just put in each charms' utils.py or equivalent file.

I'd say it will impede understanding in a charm if this function is in charm-helpers, and for each new charm that needs this functionality, charm-helpers will need to be modified.  It's not a common function.  Please remove it.

charmhelpers/core/host.py
```diff
@@ -756,6 +757,25 @@ def restart_on_change_helper(lambda_f, restart_map, stopstart=False,
     return r
 
 
+def restart_pid_check(service_name, ptable_string=None):
+    """Stop a service, check the processes are gone, start service
+    @param service_name: service name as init system knows it
+    @param ptable_string: string to look for in process table to match service
```
ajkavanagh: This is 'old' style parameter naming; we tend to use this style please:

```python
    :param service_name: the service name as the init system knows it.
    :type service_name: str
    :param ptable_match: the string to lok for in the process table to match the service
    :param ptable_match: str
    :raises: AssertionError if the processes do not go in 'n' seconds
```

Note that I've changed the name of the variable to `ptable_match` as generally it's not normal to encode the type in the name of the variable.

Also, I've added the `Exception` that occurs when the retries are exhausted.  It would also be good to calculate 'n' (which is 3 + 6 + 9 + 12 + 15 seconds ... and I'm not sure about the 15 without looking at the code!)

charmhelpers/core/host.py
```diff
@@ -756,6 +757,25 @@ def restart_on_change_helper(lambda_f, restart_map, stopstart=False,
...
+    @retry_on_exception(5, base_delay=3, exc_type=AssertionError)
+    def check_pids_gone(svc_string):
+        log("Checking no pids for {} exist".format(svc_string), level=INFO)
+        assert(subprocess.call(["pgrep", svc_string, "--nslist", "pid",
+                               "--ns", str(os.getpid())]) == 1)
+
+    if not ptable_string:
+        ptable_string = service_name
+    service_stop(service_name)
+    check_pids_gone(ptable_string)
```
ajkavanagh: So if this fails, then the function will throw an `AssertionException` which  I doubt will be caught, and the hook will crash.  Is this really what is wanted at this stage?

This function is supposed to restart the associated `service_name`, but stopping it, making sure the associated processes are gone, and then restarting it.  But it gives up!

Would it not be better to catch the AssertionError and then just `pkill` those processes and wait for them to die?  i.e. let's be nice, but if they won't play ball, kill them.  And then restart the service.  This way the intent of the function is preserved *and* it's what the user would have to do anyway.

PR #295, opened 2019-03-08
Adding arch method in host

joedborg:


charmhelpers/core/host.py
```diff
@@ -1074,3 +1074,14 @@ def install_ca_cert(ca_cert, name=None):
     log("Installing new CA cert at: {}".format(cert_file), level=INFO)
     write_file(cert_file, ca_cert)
     subprocess.check_call(['update-ca-certificates', '--fresh'])
+
+
+def arch():
+    """
+    Return the package architecture as a string.
```
ajkavanagh: The header for the docstring should start on the same line as `"""` e.g.

```python
"""Return the package ...
```

charmhelpers/core/host.py
```diff
@@ -1074,3 +1074,14 @@ def install_ca_cert(ca_cert, name=None):
...
+
+    :return: String
```
ajkavanagh: The more correct docstring comment here is:

```python
    """...

    :returns: the architecture
    :rtype: str
    :raises: subprocess.CalledProcessError if dpkg command fails
    """
```

charmhelpers/core/host.py
```diff
@@ -1074,3 +1074,14 @@ def install_ca_cert(ca_cert, name=None):
     log("Installing new CA cert at: {}".format(cert_file), level=INFO)
     write_file(cert_file, ca_cert)
     subprocess.check_call(['update-ca-certificates', '--fresh'])
+
+
+def arch():
```
ajkavanagh: As this is an "expensive" function that's never going to change it's return value once called, it's probably useful to add `@cached` to the function:

```python
@cached
def arch():
    ...
```

`@cached` is available in charmhelpers/core/hookenv.py ; hookenv is imported at the top of the file.

joedborg:
Thanks for all the feedback @ajkavanagh, I think I've addressed the points in the commit.

PR #299, opened 2019-03-20
Update Keystone expectations to meet security guide

ChrisMacNaughton:
According to the [Security Checklist][1], Keystone's config
files should be owned by keystone:keystone

[1]: https://docs.openstack.org/security-guide/identity/checklist.html#check-identity-01-is-user-group-ownership-of-config-files-set-to-keystone

### DIFF
--- a/charmhelpers/contrib/openstack/audits/openstack_security_guide.py
+++ b/charmhelpers/contrib/openstack/audits/openstack_security_guide.py
@@ -84,19 +84,21 @@ FILE_ASSERTIONS = {
     'keystone': {
         # From security guide
         '/etc/keystone/keystone.conf':
-            {'user': 'keystone', 'group': 'keystone', 'mode': '640'},
+            {'owner': 'keystone', 'group': 'keystone', 'mode': '640'},
         '/etc/keystone/keystone-paste.ini':
-            {'user': 'keystone', 'group': 'keystone', 'mode': '640'},
+            {'owner': 'keystone', 'group': 'keystone', 'mode': '640'},
         '/etc/keystone/policy.json':
-            {'user': 'keystone', 'group': 'keystone', 'mode': '640'},
+            {'owner': 'keystone', 'group': 'keystone', 'mode': '640'},
         '/etc/keystone/logging.conf':
-            {'user': 'keystone', 'group': 'keystone', 'mode': '640'},
+            {'owner': 'keystone', 'group': 'keystone', 'mode': '640'},
         '/etc/keystone/ssl/certs/signing_cert.pem':
-            {'user': 'keystone', 'group': 'keystone', 'mode': '640'},
+            {'owner': 'keystone', 'group': 'keystone',
+             'mode': '640', 'optional': True},
         '/etc/keystone/ssl/private/signing_key.pem':
-            {'user': 'keystone', 'group': 'keystone', 'mode': '640'},
+            {'owner': 'keystone', 'group': 'keystone', 'mode': '640', 'optional': True},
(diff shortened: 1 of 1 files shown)


charmhelpers/contrib/openstack/audits/openstack_security_guide.py
```diff
@@ -83,16 +83,20 @@
...
         '/etc/keystone/keystone-paste.ini':
-            {'group': 'keystone', 'mode': '640'},
-        '/etc/keystone/policy.json': {'group': 'keystone', 'mode': '640'},
-        '/etc/keystone/logging.conf': {'group': 'keystone', 'mode': '640'},
+            {'user': 'keystone', 'group': 'keystone', 'mode': '640'},
+        '/etc/keystone/policy.json':
+            {'user': 'keystone', 'group': 'keystone', 'mode': '640'},
+        '/etc/keystone/logging.conf':
+            {'user': 'keystone', 'group': 'keystone', 'mode': '640'},
         '/etc/keystone/ssl/certs/signing_cert.pem':
```
fnordahl: In our implementation these certs will probably not be there, I assume non-existence of the file would make it pass?

ChrisMacNaughton: At the moment, actually, no. Missing file is going to fail - I'll modify this PR to allow for optional files to be included as well

ChrisMacNaughton:
> LGTM, thanks!
> 
> Does the guide mention the Fernet keys used for credentials and fernet tokens? Should we work to proactively add them?

The only mention of Fernet tokens is that you should be using them. The Keystone charm actually already has an audit to validate that as well.

fnordahl:
> The only mention of Fernet tokens is that you should be using them. The Keystone charm actually already has an audit to validate that as well.

Cool. Having the correct file ownership and permissons on the key files behind Fernet is quite important though, so maybe we should contribute that as a check to the upstream guide?

ChrisMacNaughton:
> maybe we should contribute that as a check to the upstream guide?

Indeed! I'd like to get the details ironed out and will happily submit a patch to the guide

charmhelpers/contrib/openstack/audits/openstack_security_guide.py
```diff
@@ -182,7 +184,7 @@ def _config_ini(path):
     return dict(conf)
 
 
-def _validate_file_ownership(owner, group, file_name):
+def _validate_file_ownership(owner, group, file_name, optional=False):
```
ajkavanagh: You'll need to add optional to the docstring below.

charmhelpers/contrib/openstack/audits/openstack_security_guide.py
```diff
@@ -207,7 +210,7 @@ def _validate_file_ownership(owner, group, file_name):
     print("Validate ownership of {}: PASS".format(file_name))
 
 
-def _validate_file_mode(mode, file_name):
+def _validate_file_mode(mode, file_name, optional=False):
```
ajkavanagh: And optional again to docstring

ChrisMacNaughton:
> Cool. Having the correct file ownership and permissons on the key files behind Fernet is quite important though

This latest update is a lot broader in its restrictions, and should be a good addition to the security guide, while reducing the complexity of requiring that every file that should be restricted is individually checked.

The reasoning behind moving to a top level 750 is that the owner (generally root) can write in the directory, and the group (the service) can read within it. This also gives new files inside the directory 644, which lets others read as well, except the directory permission restricts that before they get to the file permission check.

PR #300, opened 2019-03-20
Added an "ignore" option to sysctl_create

petevg:
Exposes sysctl's -e flag, which allows us to ignore "unknown variable"
errors. Gives us a lightweight way of setting what sysctl we can in a
charm, without probling /proc/sys to see what modules are actually
loaded.


petevg:
This addresses https://bugs.launchpad.net/charm-nova-compute/+bug/1820635

PR #301, opened 2019-03-21
Switch test runner to tox and update travis-ci definition

ryan-beisner:
This provides consistent testability for developers to exercise
the code locally in the same way that it will be tested in the
merge gate (ex. `tox -e py3`).


ryan-beisner:
@javacruft @thedac  Ready for review.  Confirmed in a Xenial container as well as travis.

ryan-beisner:
Keep in mind, what I am NOT solving here, are the unit tests which do unsavory things like sudo foo, call out to juju binaries, etc. -- that is a separate issue.

thedac:
Right. That will be phase two.

I have one request. Starting with a clean Xenial lxc container there are quite a number of dependencies required that are not represented in the .travis.yaml. Can we update the README to document them?

They are something like:

sudo apt install gcc python3-dev python3-pip libxml2-dev libxslt1-dev zlib1g-dev g++ tox python-minimal python-dev python3-dev bzr juju

PR #303, opened 2019-03-27
Added context generator for logrotate

dparv:
New simple context generator for logrotate + template.

The logrotate context generator has 3 arguments:
 * interval
 * rotate count
 * location of log files in /var/log/

Added coresponding template.

Related-Bug: #1692118

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -1935,3 +1935,30 @@ class VersionsContext(OSContextGenerator):
         return {
             'openstack_release': ostack,
             'operating_system_release': osystem}
+
+
+class LogrotateContext(OSContextGenerator):
+    """Common context generator for logrotate."""
+
+    def __init__(self, location, interval, count):
+        """
+        :param location: Absolute path for the logrotate config file
+        :type location: str
+        :param interval: The interval for the rotations. Valid values are
+                         'daily', 'weekly', 'monthly', 'yearly'
+        :type interval: str
+        :param count: The logrotate count option configures the 'count' times
+                      the log files are being rotated before being
+        :type count: int
+        """
+        self.location = location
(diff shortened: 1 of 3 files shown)


dparv:
Pull-request updated, HEAD is now 608096ebd63d87ce9983ce985359f779fde23fab

charmhelpers/contrib/openstack/context.py
```diff
@@ -1935,3 +1935,20 @@ def __call__(self):
         return {
             'openstack_release': ostack,
             'operating_system_release': osystem}
+
+
+class LogrotateContext(OSContextGenerator):
+    """Common context generator for logrotate."""
+
+    def __init__(self, location=None, interval=None, count=None):
```
ajkavanagh: Should these values be `None`.  I suspect that, according to the example, the signature should be something like:

```python
    def __init__(self, location, interval=4, count=4):
```

This function also needs a docstring to describe the parameters and their meaning. e.g.

```python
    :param location: Absolute path for the log rotate config file
    :type location: str
    :param interval: The interval between rotations in x units (default 4)
    :type interval: int
    :param count: The number of log files to keep (default 4)
    :type count: int
```

charmhelpers/contrib/openstack/context.py
```diff
@@ -1935,3 +1935,20 @@ def __call__(self):
...
+        self.location = location
+        self.interval = interval
+        self.count = ('rotate %s' % count)
```
ajkavanagh: This should use `.format()` please.

dparv:
Pull-request updated, HEAD is now 45fd83ec20390cb0a462d0b08e32a303b35f4da4

dparv:
Pull-request updated, HEAD is now a2bf593f473c95922e2b1d277bece3f9415a7da2

PR #311, opened 2019-04-25
Preserve old keymap entries on NRPE.write

auria:
* NRPE.write retrieves the [monitors] (monitors/remote/nrpe) relation
data before overriding it with relation_set
 * If NRPE.remove_check was called, old entries (removed) will be
removed, but others will be preserved

### DIFF
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -268,8 +268,8 @@ class NRPE(object):
     def add_check(self, *args, **kwargs):
         self.checks.append(Check(*args, **kwargs))
         shortname = kwargs['shortname']
-        if shortname in self.remove_check_queue.copy():
-            del self.remove_check_queue[shortname]
+        if shortname in self.remove_check_queue:
+            self.remove_check_queue.remove(shortname)
 
     def remove_check(self, *args, **kwargs):
         if kwargs.get('shortname') is None:
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -330,11 +330,12 @@ class NRPE(object):
                 old_nrpe_monitors = {k: v for k, v in old_nrpe_monitors.items() if k not in self.remove_check_queue}
                 # update/add nrpe_monitors
                 old_nrpe_monitors.update(nrpe_monitors)
+                old_monitors['monitors']['remote']['nrpe'] = old_nrpe_monitors
                 # write back to the relation
                 relation_set(relation_id=rid, monitors=yaml.dump(old_monitors))
             else:
                 # write a brand new set of monitors, as no existing ones.
-                relation_set(relation_id=rid, monitor=yaml.dump(monitors))
+                relation_set(relation_id=rid, monitors=yaml.dump(monitors))
 
         self.remove_check_queue.clear()
 
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -268,7 +268,7 @@ class NRPE(object):
     def add_check(self, *args, **kwargs):
         self.checks.append(Check(*args, **kwargs))
         shortname = kwargs['shortname']
-        if shortname in self.remove_check_queue:
+        if shortname in self.remove_check_queue.copy():
             del self.remove_check_queue[shortname]
 
     def remove_check(self, *args, **kwargs):
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -266,6 +266,9 @@ class NRPE(object):
         self.remove_check_queue = set()
 
     def add_check(self, *args, **kwargs):
+        if kwargs.get('shortname') is None:
+            raise ValueError('shortname of check must be specified')
+
         self.checks.append(Check(*args, **kwargs))
         shortname = kwargs['shortname']
         try:
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -323,19 +323,18 @@ class NRPE(object):
         for rid in monitor_ids:
             reldata = relation_get(unit=local_unit(), rid=rid)
             if 'monitors' in reldata:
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -260,11 +261,15 @@ def __init__(self, hostname=None, primary=True):
...
-            for rid in relation_ids('nrpe-external-master'):
+            for rid in relation:
                 relation_set(relation_id=rid, relation_settings={'primary': self.primary})
+        self.remove_check_queue = set()
 
     def add_check(self, *args, **kwargs):
         self.checks.append(Check(*args, **kwargs))
+        shortname = kwargs['shortname']
+        if shortname in self.remove_check_queue.copy():
+            del self.remove_check_queue[shortname]
```
ajkavanagh: Won't this `TypeError` if it is a set?

A better way to do this is:

```python
try:
    self.remove_check_queue.remove(shortname)
except KeyError:
    pass
```

ajkavanagh:
Hi @auria -- it looks like you need to rebase the branch.

auria:
Thank you for the reviews, @ajkavanagh . I have now rebased the branch and pushed it again.

charmhelpers/contrib/charmsupport/nrpe.py
auria: rewrote that part following your suggestion.

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -313,8 +321,24 @@ def write(self):
...
+                old_nrpe_monitors = old_monitors['monitors']['remote']['nrpe']
+                for shortname in old_nrpe_monitors.copy():
+                    if shortname in self.remove_check_queue:
+                        del old_nrpe_monitors[shortname]
+                    elif shortname in nrpe_monitors:
+                        old_nrpe_monitors[shortname] = nrpe_monitors[shortname].copy()
+                        del nrpe_monitors[shortname]
+                for shortname in nrpe_monitors:
+                    old_nrpe_monitors[shortname] = nrpe_monitors[shortname]
+                monitors = old_monitors.copy()
```
stub42: There is no need to make a second copy of the dict, because yaml.dump isn't going to mutate it.

ajkavanagh: I'm a bit confused by this code.  Reviewing alongside the existing code, here's what I *think* it is doing.

1. The existing code overwrote the relation data with newly calculated sets of checks.
2. The idea is to preserve existing entries unless they have been explicitly removed.
3. Any existing (in the relation data) monitor entries should be updated with new information (based on the newly computed info).

If that is the case, then the above code (for me) obfuscates it; thus it might become a maintenance issue.

The bit that seems to be changing is `nrpe_monitors` dictionary and the keynames appear to be the bit that is checked.  Essentially, remove elements that are in `self.remove_check_queue` and update any existing elements from the new `monitors` dict.

Curiously, only update the first occurrence of shortname found in `nrpe_monitors` as it is then deleted so it can't be used again.  Is that what's really wanted?  I'm not sure if that is a bug or not?

Anyway, I'm wondering if this is the code that you want:

```python
for rid in monitor_ids:
    reldata = relation_get(unit=local_unit(), rid=rid)
    if 'monitors' in reldata:
        # update the existing set of monitors with the new data
        old_monitors = yaml.safe_load(reldata['monitors'])
        old_nrpe_monitors = old_monitors['monitors']['remote']['nrpe']
        # remove keys that are in the remove_check_queue
        old_nrpe_monitors = {k: v for k, v in old_nrpe_monitors.items() if k not in self.remove_check_queue}
        # update/add nrpe_monitors
        old_nrpe_monitors.update(nrpe_monitors)
        # OPTIONAL (if needed): remove remove nrpe_monitor shortnames from nrpe_monitors (i.e. they only get applied once???)
        nrpe_monitors = {k: v for k, v in nrpe_monitors.items() if k in old_nrpe_monitors}
        # write back to the relation
        relation_set(relation_id=rid, monitors=yaml.dump(old_monitors))
    else:
        # write a brand new set of monitors, as no existing ones.
        relation_set(relation_id=rid, monitor=yaml.dump(monitors))
```

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -260,11 +261,17 @@ def __init__(self, hostname=None, primary=True):
...
         if relation:
             log("Setting charm primary status {}".format(primary))
-            for rid in relation_ids('nrpe-external-master'):
+            for rid in relation:
                 relation_set(relation_id=rid, relation_settings={'primary': self.primary})
+        self.remove_check_queue = set()
 
     def add_check(self, *args, **kwargs):
         self.checks.append(Check(*args, **kwargs))
+        shortname = kwargs['shortname']
```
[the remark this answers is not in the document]
auria: done, thank you.

auria:
@ajkavanagh I've spent some more time on this fix and troubleshooted your suggestion, and I think we now have a tested version that works (my tests suggest so). Could you please have a quick look to see if it can be merged? FTR, I've tested this nrpe.py version on https://bugs.launchpad.net/charm-openstack-service-checks/+bug/1826018 and worked as expected. Thank you.

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -313,7 +324,23 @@ def write(self):
...
             relation_ids("nrpe-external-master")
         for rid in monitor_ids:
-            relation_set(relation_id=rid, monitors=yaml.dump(monitors))
+            reldata = relation_get(unit=local_unit(), rid=rid)
+            if 'monitors' in reldata:
+                # update the existing set of monitors with the new data
+                old_monitors = yaml.safe_load(reldata['monitors'])
+                old_nrpe_monitors = old_monitors['monitors']['remote']['nrpe']
+                # remove keys that are in the remove_check_queue
+                old_nrpe_monitors = {k: v for k, v in old_nrpe_monitors.items() if k not in self.remove_check_queue}
```
ajkavanagh: This line should fail PEP E501 line too long.  if you reformat it to:

```python
old_nrpe_monitors = {k: v for k, v in old_nrpe_monitors.items()
                     if k not in self.remove_check_queue}
```

it'll look nicer too.

PR #313, opened 2019-04-29
Make XFS inode size configurable

tytus-kurek:
Make XFS inode size configurable, so that charm could set it to the
desired value, if different than 1024.

Partial-Bug: 1826552

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,23 @@
-dist: trusty
+dist: xenial
 sudo: required
 language: python
 python:
   - "2.7"
-  - "3.4"
+  - "3.5"
+addons:
+  snaps:
+  - name: juju
+    classic: true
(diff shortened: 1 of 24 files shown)


tytus-kurek:
@ajkavanagh 

The required changes have been already introduced, so why is this PR in the "Changes requested" state? Or am I missing something?

Thanks

ajkavanagh:
> The required changes have been already introduced, so why is this PR in the "Changes requested" state? Or am I missing something?

@tytus-kurek it's just an artefact of Github.  It's showing that I requested changes, but github doesn't know that they've been resolved until I indicate that they have.  Which I'm about to do

tytus-kurek:
@ajkavanagh I see, thank you very much for the explanation and the review.

PR #315, opened 2019-05-01
Bug/1786186

nikolayvinogradov:
Initial implementation of generation keystone-authsection in Python code with variable names included in the context. It allows to rename variables depending on OpenStack version, which in turn allows to support keystonemiddleware option deprecation for example.

Related-to: https://bugs.launchpad.net/charm-glance/+bug/1786186


charmhelpers/contrib/openstack/context.py
```diff
@@ -352,10 +352,101 @@ def _setup_pki_cache(self):
...
+
+        return pkg
+
+    def _get_keystone_authtoken_ctxt(self, ctxt, keystonemiddleware_os_rel):
+        """
+        Re-constructed from former template 'section-keystone-auth-mitaka'
+        """
+        config_items = []
+
+        if CompareOpenStackReleases(keystonemiddleware_os_rel) >= 'mitaka':
```
gnuoy: I am *really* sorry to mess you about. I know I suggested we should have support from the earliest release but I think I was wrong. I think we should support from Queens and above only.

charmhelpers/contrib/openstack/context.py
```diff
@@ -352,10 +352,101 @@ def _setup_pki_cache(self):
...
+                ('user_domain_name', 'default'),
+            )
+
+        config_items += (
+            ('project_name', ctxt.get('admin_tenant_name', '')),
+            ('username', ctxt.get('admin_user', '')),
+            ('password', ctxt.get('admin_password', '')),
+        )
+
+        if CompareOpenStackReleases(keystonemiddleware_os_rel) < 'mitaka':
```
gnuoy: Again *sorry* but lets stick to Queens and above. Please can we set project_domain_id and user_domain_id if the corresponding  ctxt['service_tenant_id'] and ctxt['service_domain_id'] are set.

charmhelpers/contrib/openstack/context.py
```diff
@@ -352,10 +352,101 @@ def _setup_pki_cache(self):
...
+
+        api_version = ctxt.get('api_version')
+
+        auth_uri = "{service_protocol}://{service_host}:{service_port}" \
+                   .format(**ctxt)
+
+        auth_url = "{auth_protocol}://{auth_host}:{auth_port}" \
+                   .format(**ctxt)
+
+        if float(api_version) >= 3.0:
```
gnuoy: Queens is v3 only so we can remove this conditional

charmhelpers/contrib/openstack/context.py
```diff
@@ -352,10 +352,101 @@ def _setup_pki_cache(self):
...
+        else:
+            config_items.append(
+                ('auth_uri', auth_uri),
+            )
+
+        config_items.append(
+            ('auth_url', auth_url),
+        )
+
+        if float(api_version) >= 3.0:
```
gnuoy: Change this to always set them to the values from the context.

charmhelpers/contrib/openstack/context.py
```diff
@@ -352,10 +352,102 @@ def _setup_pki_cache(self):
             return cachedir
         return None
 
+    def _get_ubuntu_pkg_name(self, python_name):
```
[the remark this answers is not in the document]
nikolayvinogradov: Thank you pointing that out. I remember we discussed the platforms part, but I forgot to clean this up... Sorry.

charmhelpers/contrib/openstack/context.py
nikolayvinogradov: No problem. This part transforms just to:
```
config_items.append(
    ('auth_type', 'password')
)
```
then. Also given that input, i think it is might be safe to drop charmhelpers/contrib/openstack/section-keystone-authtoken altogether, but not sure...

charmhelpers/contrib/openstack/context.py
nikolayvinogradov: NP, removing that as well.

charmhelpers/contrib/openstack/context.py
nikolayvinogradov: If we only support v3, then I think it would be good to remove extra conditions from keystone-authtoken-mitaka as well to clean up.

gnuoy:
LGTM, thanks for all your work on this.

PR #317, opened 2019-05-07
openstack: add global-physnet-mtu to NeutronAPIContext 

sahid:
This will be used to instruct MTU for DPDK interfaces.

There two commits, one to add global-physnet-mtu which will be used as the default or fallback value and physical-network-mtus which will be used to defined MTU based on networking cards.

Related-To:
  https://bugs.launchpad.net/charm-neutron-openvswitch/+bug/1827256

Signed-off-by: Sahid Orentino Ferdjaoui <sahid.ferdjaoui@canonical.com>

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -1611,6 +1611,10 @@ class NeutronAPIContext(OSContextGenerator):
                 'rel_key': 'global-physnet-mtu',
                 'default': 1500,
             },
+            'physical-network-mtus': {
+                'rel_key': 'physical-network-mtus',
+                'default': None,
+            },
         }
         ctxt = self.get_neutron_options({})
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/context.py
```diff
@@ -1598,6 +1598,10 @@ def __call__(self):
                 'rel_key': 'enable-nsg-logging',
                 'default': False,
             },
+            'global-physnet-mtu': {
```
dshcherb: I suggest that we also pass physical-network-mtus because MTU can be overridden on a per-physnet basis.

sahid: Makes sense, thanks.

ChrisMacNaughton:
Looks reasonable, but it needs to be rebased on master

PR #323, opened 2019-05-15
openstack: add send_notifications_to_logs option

sahid:
openstack: add send_notifications_to_logs option

This new option introduced will give to the charms ability to set
additional 'log' driver so the oslo messaging notifications will also
be sent to the system logs

Related-to:
  https://bugs.launchpad.net/charm-nova-compute/+bug/1825016


gnuoy:
LGTM, thanks

PR #324, opened 2019-05-16
Add Contexts for Nova Vendor Metadata

rodrigogansobarbieri:
Added two new generic contexts to avoid more duplication
of code on several projects when dealing with nova
vendor metadata.

NovaVendorMetadataContext is responsible for reading
"vendor-data" and "vendor-data-url" config options and
returning the proper context for setting the values to
nova.conf file. If a particular charm needs to complement
the logic, such as disregarding a certain value for a
given OpenStack release or propagating the values to a
relation, it is recommended that the context is inherited
and specialized for that particular charm.

NovaVendorMetadataJSONContext is responsible for reading
only the vendor-data config option, validating it and
returning the context to write the value to vendor_data.json
file. Since this context is simple enough, it is included a
min_release and max_release validation in order to not need
to be inherited and specialized for that purpose.

Related-Bug: #1777714

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -114,6 +114,11 @@ except ImportError:
         apt_install('python3-psutil', fatal=True)
     import psutil
 
+if six.PY3:
+    json_error = json.decoder.JSONDecodeError
+else:
+    json_error = ValueError
+
 CA_CERT_PATH = '/usr/local/share/ca-certificates/keystone_juju_ca_cert.crt'
 ADDRESS_TYPES = ['admin', 'internal', 'public']
 HAPROXY_RUN_DIR = '/var/run/haproxy/'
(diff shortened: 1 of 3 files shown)


charmhelpers/contrib/openstack/context.py
```diff
@@ -521,6 +526,64 @@ def __call__(self):
...
+            ctxt['vendordata_providers'] = ','.join(vdata_providers)
+
+        return ctxt
+
+
+class NovaVendorMetadataJSONContext(OSContextGenerator):
+    """Context used for writing nova vendor metadata json file."""
+
+    def __init__(self, os_release_pkg):
+        self.os_release_pkg = os_release_pkg
```
ajkavanagh: As above, re: docstring

charmhelpers/contrib/openstack/context.py
```diff
@@ -521,6 +526,64 @@ def __call__(self):
         return {}
 
 
+class NovaVendorMetadataContext(OSContextGenerator):
```
ajkavanagh: I'm a bit confused as to why this is called the `NovaVendor...` rather than just `VendorMeta...` context.  There doesn't seem to be anything specific in the class that is referring to nova?  If it is more generalised, then adding `Nova` may confuse things in the future?

rodrigogansobarbieri: Alex, I thought about naming NovaVendorMetadata so it is easier to know this would apply to Nova Metadata API service, and the context returned is aimed at nova.conf file.

charmhelpers/contrib/openstack/context.py
```diff
@@ -521,6 +526,64 @@ def __call__(self):
...
+    """Context used for configuring nova vendor metadata on nova.conf file."""
+
+    def __init__(self, os_release_pkg, interfaces=None):
```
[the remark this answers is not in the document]
rodrigogansobarbieri: Done. Thanks for the suggestion

charmhelpers/contrib/openstack/context.py
```diff
@@ -521,6 +526,64 @@ def __call__(self):
...
+        self.os_release_pkg = os_release_pkg
+        self.interfaces = interfaces if interfaces else []
```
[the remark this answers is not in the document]
rodrigogansobarbieri: Done, indeed it is much better now, thanks!

charmhelpers/contrib/openstack/context.py
rodrigogansobarbieri: Done. THanks!

charmhelpers/contrib/openstack/context.py
ajkavanagh: Rodrigo, ah, that is a good reason.  Let's keep it as is.

PR #325, opened 2019-05-17
NRPE: Don't filter quotes from Check.check_cmd

auria:
Closes-Bug: 1829496


auria:
AttributeError: 'module' object has no attribute 'quote'

shlex.quote is not supported in py2

PR #327, opened 2019-05-29
Fix wrong usage of relation_get in *_broker_action_done

xtrusia:
Because of wrong usage, it doesn't behaviour as expected.
e.g when config-changed hook, it runs multiple times 
when there are multiple ceph-mons
Making it use right signature and reduce 2 lines to 1 line.


xtrusia:
Thanks a lot @ajkavanagh 
I've done what you said.

PR #329, opened 2019-05-29
Ensure CephContext will correctly be incomplete

ChrisMacNaughton:
If the relation appears before a unit appears on the
relation, it is possible to end up with a context that
is considered complete even though it is missing all of
the required relation data. This change ensures that we
cannot complete until the relation is complete

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -781,6 +781,25 @@ class CephContext(OSContextGenerator):
         ensure_packages(['ceph-common'])
         return ctxt
 
+    def context_complete(self, ctxt):
+        """Overridden here to ensure the context is actually complete.
+
+        We set `key` and `auth` to None here, by default, to ensure
+        that the context will always evaluate to incomplete until the
+        Ceph relation has actually sent these details; otherwise,
+        there is a potential race condition between the relation
+        appearing and the first unit actually setting this data on the
+        relation.
+
+        :param ctxt: The current context members
+        :type ctxt: Dict[str, ANY]
+        :returns: True if the context is complete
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/context.py
```diff
@@ -746,7 +746,15 @@ def __call__(self):
 
         log('Generating template context for ceph', level=DEBUG)
         mon_hosts = []
+        # We set `key` and `auth` to None here, by default, to ensure
```
ajkavanagh: Rather than set it to `None` here, why not override `context_complete` to check for the missing keys and not allow them to be `None` (which could change the meaning elsewhere)?

e.g.

```python
def context_complete(self, ctxt):
    """Override to ...
    
    :param ctxt: The current context members ...
    :type ctxt: Dict[str, ANY]
    :returns: True if the context is complete
    :rtype: bool
    """
    if 'auth' not in ctxt or 'key' not in ctxt:
        return False
    return super(CephContext, self).context_complete(ctxt)
```

ChrisMacNaughton: Sounds elegant :)

charmhelpers/contrib/openstack/context.py
```diff
@@ -781,6 +781,18 @@ def __call__(self):
         ensure_packages(['ceph-common'])
         return ctxt
 
+    def context_complete(self, ctxt):
+        """Override to ...
```
ajkavanagh: I was hoping you'd expand the "..." :)  -- it was nice text you'd written in the last commit.

ChrisMacNaughton: How's that ;-)

PR #331, opened 2019-05-30
Allow disabling load balancing

tytus-kurek:
This patchset implements a logic to control whether HAProxy load
balancing should be enabled or not based on the 'ha-lb-enable' config
option.

Partial-Bug: 1827356

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -805,6 +805,13 @@ class HAProxyContext(OSContextGenerator):
         l_unit = local_unit().replace('/', '-')
         cluster_hosts = collections.OrderedDict()
 
+        # NOTE(tkurek): enable haproxy load balancing by default for legacy
+        # purposes
+        if config('ha-lb-enable') is None:
+            enable_lb = True
+        else:
+            enable_lb = config('ha-lb-enable')
+
         # NOTE(jamespage): build out map of configured network endpoints
         # and associated backends
         for addr_type in self.address_types:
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/context.py
```diff
@@ -805,6 +805,13 @@ def __call__(self):
         l_unit = local_unit().replace('/', '-')
         cluster_hosts = collections.OrderedDict()
 
+        # NOTE(tkurek): enable haproxy load balancing by default for legacy
+        # purposes
+        if config('ha-lb-enable') is None:
+            enable_lb = True
+        else:
+            enable_lb = config('ha-lb-enable')
```
ChrisMacNaughton: Why not do something like `enable_lb = config('ha-lb-enable') or True`

ajkavanagh:
@tytus-kurek is this just for swift?  Because if it's only for swift, then maybe it should only be done in swift :)  i.e. to subvert the currently meaning of the haproxy context which is currently all about load-balancing (and was the design ... basically to provide a single VIP for multiple units), then it ought to be specialised into a Swift haproxy context in the swift charm?  Or have I got it wrong? :/  (which, is entirely possible as I may not understand the full context of the change/requirement).

tytus-kurek:
@ajkavanagh:

I need it only for swift, but I think it would make sense to make it a general option in the same way as other HAProxy-related options. However, I am not the authority here, so I'll let the experts decide :).

tytus-kurek:
@ajkavanagh:

Please make a decision here and I'll adjust, but I'd really like to move forward here.

Thanks

ajkavanagh:
@tytus-kurek okay, my opinion, based on my current understanding, is that this should only be in swift until it's needed elsewhere.  My reasoning is that it's a modification to HAProxy that only swift actually requires (I may be wrong on this)  Thus, I won't merge it at this time.  I will approve a merge of the same code into the swift charm.  However, other members of the core reviewers may feel that it should be here, and may agree to merge it here.  Thanks.

tytus-kurek:
@ajkavanagh thank you for a swift response. I'll close this PR and create a patch for the `swift-proxy` charm.

tytus-kurek:
@ajkavanagh I am racking my brain around that and I am not sure how should the required changed to the `swift-proxy` charm look like.

So there's [`HAProxyContext`](https://github.com/openstack/charm-swift-proxy/blob/master/lib/swift_context.py#L39) class in the `swift-proxy` charm which extends the main `HAProxyContext` class from charm-helpers. However, what I need is not to extend it, but to overwrite parts of it (`frontends` value to be more precise). If I put the required logic, will I be able to overwrite it?

Any help would be much appreciated.

ajkavanagh:
@tytus-kurek There are **two** HAProxyContext classes; one from charmhelpers, and one in swift_context.py.  This is normal, apparently.  I'm going to have to do some digging to work out the best way of doing it.

tytus-kurek:
@ajkavanagh: OK, thanks in advance for looking into it.

PR #332, opened 2019-05-30
Fix vendor_data py3 issue of PR #324

rodrigogansobarbieri:
PR #324 imports json.decoder.JSONDecodeError which
only exists on Python 3.5+. Since that exception class
inherits from ValueError, it is better to use ValueError,
else deployments using Python < 3.5 will fail.

Also, improves the consistency of the 2 nova vendor metadata
contexts, by having NovaVendorMetadataContext not return
context values related to StaticJSON when an invalid JSON
is configured.

Related-Bug: #1777714


freyes:
LGTM

niedbalski:
LGTM

PR #334, opened 2019-06-12
Set unit_name when requesting certificates.

gnuoy:
A fix has landed in the tls-certificates interface to support
relating to vaults certificates relation over a cross-model
relation *1. The old style charms do not use this interface
so the fixes need to be ported to charm helpers. The fix works by
having the client explicitly publish a unit name in the
relation data. That unit name is then used by the client
when looking up its certificates in the returned dict.

*1 https://github.com/juju-solutions/interface-tls-certificates/commit/231c86ce4086fa4d3d78414594336089b519f704

### DIFF
--- a/charmhelpers/contrib/openstack/cert_utils.py
+++ b/charmhelpers/contrib/openstack/cert_utils.py
@@ -106,9 +106,11 @@ class CertRequest(object):
             sans = sorted(list(set(entry['addresses'])))
             request[entry['cn']] = {'sans': sans}
         if self.json_encode:
-            return {'cert_requests': json.dumps(request, sort_keys=True)}
+            req = {'cert_requests': json.dumps(request, sort_keys=True)}
         else:
-            return {'cert_requests': request}
+            req = {'cert_requests': request}
+        req['unit_name'] = local_unit().replace('/', '_')
+        return req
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/cert_utils.py
```diff
@@ -226,8 +239,10 @@ def process_certificates(service_name, relation_id, unit,
     data = relation_get(rid=relation_id, unit=unit)
     ssl_dir = os.path.join('/etc/apache2/ssl/', service_name)
     mkdir(path=ssl_dir)
-    name = local_unit().replace('/', '_')
-    certs = data.get('{}.processed_requests'.format(name))
+    certs = data.get('{}.processed_requests'.format(get_unit_name()))
+    if not certs:
+        name = local_unit().replace('/', '_')
+        certs = data.get('{}.processed_requests'.format(name))
```
javacruft: Not sure I understand how the if not certs codepath is different to L242?

gnuoy: Ah yes, sorry, I was originally including the model name in the unit name but I removed it to be consistent with the fix in the interface.

PR #337, opened 2019-06-18
Make ConfigParser not strict

ChrisMacNaughton:
When strict is enabled, duplicate options are not allowed in the
parsed INI; however, Oslo allows duplicate values. This change
causes us to ignore the duplicate values which is acceptable as
long as we don't validate any multi-value options

### DIFF
--- a/charmhelpers/contrib/openstack/audits/openstack_security_guide.py
+++ b/charmhelpers/contrib/openstack/audits/openstack_security_guide.py
@@ -126,7 +126,11 @@ def _config_ini(path):
     :returns: Configuration contained in path
     :rtype: Dict
     """
-    conf = configparser.ConfigParser()
+    # When strict is enabled, duplicate options are not allowed in the
+    # parsed INI; however, Oslo allows duplicate values. This change
+    # causes us to ignore the duplicate values which is acceptable as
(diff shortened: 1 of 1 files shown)


charmhelpers/contrib/openstack/audits/openstack_security_guide.py
```diff
@@ -126,7 +126,11 @@ def _config_ini(path):
     :returns: Configuration contained in path
     :rtype: Dict
     """
-    conf = configparser.ConfigParser()
+    # When strict is enabled, duplicate options are not allowed in the
+    # parsed INI; however, Oslo allows duplicate values. This change
+    # causes us to ignore the duplicate values which is acceptable as
+    # long as we don't validate any multi-value options
+    conf = configparser.ConfigParser(strict=False)
```
ajkavanagh: As this is a library function, I'd add a `strict=True` to the function signature and override it in the relevant charm? (or make it `strict=False`) if every charm needs it, just in case we *do* need to override.

Also, does this not have a test anywhere that fails now?

ChrisMacNaughton:
This is pending functional validation in a full recheck on https://review.opendev.org/#/c/663313/

PR #339, opened 2019-06-18
Update tests to actually be run

ChrisMacNaughton:


charmhelpers/contrib/openstack/audits/openstack_security_guide.py
```diff
@@ -204,7 +204,7 @@ def validate_file_ownership(config):
                     "Invalid ownership configuration: {}".format(key))
         owner = options.get('owner', config.get('owner', 'root'))
         group = options.get('group', config.get('group', 'root'))
-        optional = options.get('optional', config.get('optional', 'False'))
+        optional = options.get('optional', config.get('optional', False))
```
ajkavanagh: I didn't explain myself properly.  I think what you want, (if 'optional' is a str), is:

```python
    optional = options.get('optional', config.get('optional', 'false')).lower() == 'true'
```

I definitely messed up with my parenthesis earlier!

ChrisMacNaughton: Fortunately, options is a dict that is sourced from Python, so we can actually do types (`True`/`False`) rather than parsing str values!

ajkavanagh: Ah, cool!

PR #340, opened 2019-06-20
Change relation_set() to use a cached --file detection fn

ajkavanagh:
The hookenv.relation_set(...) function included a relation-set --help
subprocess check_output for every usage of relation-set.  This patch
adds an additional function which caches the result of relation-set
--help w.r.t. '--file' so that the cost of detection is only on the
first relation-set call.


ajkavanagh:
@stub42 I think that's a reasonable comment; I will update and submit a new commit.  Thanks for the review.

ChrisMacNaughton (approved):

PR #341, opened 2019-06-26
Replace ``python-apt`` functionality

fnordahl:
The ``python-apt`` package is a wrapper around the ``apt`` C++
library which is tightly connected to the version of the
distribution it is shipped on.
    
This in turn makes it incredibly hard to distribute as a wheel for
a charm that supports a large span of distro versions.
    
We do not want to rely on system installed Python packages but
distribute the direct charm dependencies as part of the charms
wheelhouse.
    
As the span of distributions we need to support with reactive
charms widens we will run into compability problems with the
current model.
    
For further reference see discussion in LP: #1824112 and
juju-solutions/layer-basic#135

### DIFF
--- a/charmhelpers/fetch/ubuntu.py
+++ b/charmhelpers/fetch/ubuntu.py
@@ -218,11 +218,11 @@ def filter_missing_packages(packages):
     )
 
 
-def apt_cache(in_memory=True, progress=None):
+def apt_cache(*_):
     """Shim returning an object simulating the apt_pkg Cache.
 
-    :param in_memory: Kept for compability reasons, has no effect.
-    :param progress: Kept for compability reasons, has no effect.
+    :param _: Accept arguments for compability, not used.
+    :type _: any
     :returns:Object used to interrogate the system apt and dpkg databases.
     :rtype:ubuntu_apt_pkg.Cache
     """
--- a/charmhelpers/fetch/ubuntu.py
+++ b/charmhelpers/fetch/ubuntu.py
@@ -218,11 +218,13 @@ def filter_missing_packages(packages):
     )
 
 
-def apt_cache(*_):
+def apt_cache(*_, **__):
     """Shim returning an object simulating the apt_pkg Cache.
 
     :param _: Accept arguments for compability, not used.
     :type _: any
+    :param __: Accept keyword arguments for compability, not used.
+    :type __: any
     :returns:Object used to interrogate the system apt and dpkg databases.
     :rtype:ubuntu_apt_pkg.Cache
     """
--- a/charmhelpers/fetch/ubuntu.py
+++ b/charmhelpers/fetch/ubuntu.py
@@ -17,8 +17,8 @@ import os
 import platform
 import re
 import six
-import time
 import subprocess
+import time
 
 from charmhelpers.core.host import get_distrib_codename
 
@@ -29,6 +29,7 @@ from charmhelpers.core.hookenv import (
     env_proxy_settings,
 )
 from charmhelpers.fetch import SourceConfigError, GPGKeyError
+from charmhelpers.fetch import ubuntu_apt_pkg
 
 PROPOSED_POCKET = (
     "# Proposed\n"
@@ -217,13 +218,14 @@ def filter_missing_packages(packages):
 
 
 def apt_cache(in_memory=True, progress=None):
-    """Build and return an apt cache."""
-    from apt import apt_pkg
-    apt_pkg.init()
-    if in_memory:
-        apt_pkg.config.set("Dir::Cache::pkgcache", "")
-        apt_pkg.config.set("Dir::Cache::srcpkgcache", "")
-    return apt_pkg.Cache(progress)
+    """Shim returning an object simulating the apt_pkg Cache.
+
+    :param in_memory: Kept for compability reasons, has no effect.
+    :param progress: Kept for compability reasons, has no effect.
+    :returns:Object used to interrogate the system apt and dpkg databases.
+    :rtype:ubuntu_apt_pkg.Cache
+    """
+    return ubuntu_apt_pkg.Cache()
 
 
 def apt_install(packages, options=None, fatal=False):
@@ -723,7 +725,6 @@ def get_upstream_version(package):
 
     @returns None (if not installed) or the upstream version
     """
-    import apt_pkg
     cache = apt_cache()
     try:
         pkg = cache[package]
@@ -735,4 +736,4 @@ def get_upstream_version(package):
         # package is known, but no version is currently installed.
         return None
 
-    return apt_pkg.upstream_version(pkg.current_ver.ver_str)
+    return ubuntu_apt_pkg.upstream_version(pkg.current_ver.ver_str)
--- a/charmhelpers/fetch/ubuntu.py
+++ b/charmhelpers/fetch/ubuntu.py
@@ -18,6 +18,7 @@ import platform
 import re
 import six
 import subprocess
+import sys
 import time
 
 from charmhelpers.core.host import get_distrib_codename
@@ -225,6 +226,17 @@ def apt_cache(in_memory=True, progress=None):
     :returns:Object used to interrogate the system apt and dpkg databases.
     :rtype:ubuntu_apt_pkg.Cache
     """
+    if 'apt_pkg' in sys.modules:
+        # NOTE(fnordahl): When our consumer use the upstream ``apt_pkg`` module
+        # in conjunction with the apt_cache helper function, they may expect us
+        # to call ``apt_pkg.init()`` for them.
+        #
+        # Detect this situation, log a warning and make the call to
+        # ``apt_pkg.init()`` to avoid the consumer Python interpreter from
+        # crashing with a segmentation fault.
+        log('Support for use of upstream ``apt_pkg`` module in conjunction'
+            'with charm-helpers is deprecated since 2019-06-25', level=WARNING)
+        sys.modules['apt_pkg'].init()
     return ubuntu_apt_pkg.Cache()
 
 
--- a/charmhelpers/fetch/ubuntu_apt_pkg.py
+++ b/charmhelpers/fetch/ubuntu_apt_pkg.py
@@ -40,17 +40,10 @@ import os
 import subprocess
 
 
-class _container(object):
+class _container(dict):
     """Simple container for attributes."""
-    def __init__(self, attr_map):
-        """Initialize package attribute container.
-
-        :param attr_map: Dictionary key value pairs to transform into
-                         attributes with a value on instance of the class.
-        :type attr_map: Dict[str, str]
-        """
-        for k, v in attr_map.items():
-            setattr(self, k, v)
+    __getattr__ = dict.__getitem__
+    __setattr__ = dict.__setitem__
 
 
 class Package(_container):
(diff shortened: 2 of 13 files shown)


fnordahl:
Functional test results in pre-existing charms with these bits available [0][1].

0: https://review.opendev.org/#/c/655370/
1: https://review.opendev.org/#/c/655394/

fnordahl:
Fresh set of functional tests run with these bits on two reactive charms [0][1] [2][3] and one classic charm [4][5] in sum spanning trusty through disco.

0: https://openstack-ci-reports.ubuntu.com/artifacts/test_charm_pipeline_func_full/openstack/charm-aodh/655370/2/3264/index.html
1: https://review.opendev.org/#/c/655370/

2: https://openstack-ci-reports.ubuntu.com/artifacts/test_charm_pipeline_func_full/openstack/charm-octavia/655394/4/3268/consoleText.test_charm_func_full_6449.txt
3: https://review.opendev.org/#/c/655394/

4: https://openstack-ci-reports.ubuntu.com/artifacts/test_charm_pipeline_func_full/openstack/charm-ceph-osd/667673/2/3267/consoleText.test_charm_func_full_6448.txt
5: https://review.opendev.org/#/c/667673/

charmhelpers/fetch/ubuntu.py
```diff
@@ -189,17 +190,185 @@
...
+    Note that one of the main purposes for this function is to avoid dependency
+    on the ``python-apt`` package.
+
+    The indicated package is a wrapper around the ``apt`` C++ library which is
+    tightly connected to the version of the distribution it is shipped on.
+    This in turn makes it incredibly hard to distribute as a wheel for a charm
+    that supports a large span of distro versions.
+
+    :param packages: Packages to get data from
+    :type packages: list[str]
```
ajkavanagh: This is picky, but it's `List[str]`

(and I believe it's because it's a class in the types module).

charmhelpers/fetch/ubuntu.py
```diff
@@ -189,17 +190,185 @@
...
+    :returns: Structured data about installed packages, keys like
+              ``dpkg-query --list``
+    :rtype: dict
```
ajkavanagh: This should be `Dict[str, str]` I believe

charmhelpers/fetch/ubuntu.py
```diff
@@ -189,17 +190,185 @@
...
+                                         universal_newlines=True)
+        headings = []
+        for line in output.splitlines():
+            if line.startswith('||/'):
+                headings = line.split()
+                headings.pop(0)
+                continue
+            elif (line.startswith('|') or line.startswith('+') or
+                  line.startswith('dpkg-query:')):
+                continue
```
ajkavanagh: Feels like there should be another `else` clause here, or convert the `elif` to an else (i.e. continue if the line didn't start with `||/`)?

charmhelpers/fetch/ubuntu.py
```diff
@@ -189,17 +190,185 @@
...
+            data = line.split(None, 4)
+            status = data.pop(0)
+            if status != 'ii':
+                continue
+            pkg = {}
+            for kv in zip(headings, data):
+                pkg.update({kv[0].lower(): kv[1]})
```
ajkavanagh: Nicer if you do something like:

```python
    pkg.update({k.lower(): v for k, v in zip(headings, data)})
```

charmhelpers/fetch/ubuntu.py
```diff
@@ -189,17 +190,185 @@
...
+            if 'name' in pkg:
+                pkgs.update({pkg['name']: pkg})
+    except subprocess.CalledProcessError as cp:
+        # ``dpkg-query`` may return error and at the same time have produced
```
ajkavanagh: Does this mean we *still* want to use the packages?  In which case, move the except up to just after the checkcall() and allow it to continue if it's the case;  here `pkgs` will be `{}`

charmhelpers/fetch/ubuntu.py
```diff
@@ -189,17 +190,185 @@
...
+    Note that one of the main purposes for this function is to avoid dependency
+    on the ``python-apt`` package.
+
+    The indicated package is a wrapper around the ``apt`` C++ library which is
+    tightly connected to the version of the distribution it is shipped on.
+    This in turn makes it incredibly hard to distribute as a wheel for a charm
+    that supports a large span of distro versions.
+
+    :param packages: Packages to get data from
+    :type packages: list[str]
```
ajkavanagh: As above

charmhelpers/fetch/ubuntu.py
```diff
@@ -189,17 +190,185 @@
...
+    :param version: Version string
+    :type version: str
+    :returns: Upstream version
+    :rtype: str
+
+    Upstream reference: https://salsa.debian.org/apt-team/apt/blob/master/
+                                apt-pkg/deb/debversion.cc#L259
+    """
+    if version:
+        version = version[version.find(':') + 1:]
```
ajkavanagh: Alternative:

```python
    version = version.split(':')[-1]
```

It will always pass, even if there is no ':' in the string (it returns the whole string).  if you are worried about more than one ':', use `split(':', 1)`

The `+ 1` makes it a bit more interesting.  Add a `[1:]` after the `[-1]` to strip off the first char.

charmhelpers/fetch/ubuntu.py
```diff
@@ -189,17 +190,185 @@
...
+        if '-' in version:
+            version = version[:version.find('-')]
```
ajkavanagh: as above, but `version.split('-')[-1]`

charmhelpers/fetch/ubuntu.py
```diff
@@ -189,17 +190,185 @@
...
+    try:
+        subprocess.check_call(['dpkg', '--compare-versions', candidate, 'gt',
+                               current],
+                              stderr=subprocess.STDOUT,
+                              universal_newlines=True)
+        return True
+    except subprocess.CalledProcessError as cp:
+        if cp.returncode == 1:
+            return False
+        raise
```
ajkavanagh: Do we really want to `raise` here?  i.e. wouldn't an error be `False` anyway?

charmhelpers/fetch/ubuntu.py
```diff
@@ -189,17 +190,185 @@
...
+
+
 def filter_installed_packages(packages):
     """Return a list of packages that require installation."""
```
ajkavanagh: docstring?

ajkavanagh:
A thought occurs to me; one not keep the API the same as apk_pkg so it's a plug in replacement?

charmhelpers/fetch/ubuntu.py
fnordahl: :+1:

charmhelpers/fetch/ubuntu.py
fnordahl: :+1:

charmhelpers/fetch/ubuntu.py
fnordahl: :+1:

charmhelpers/fetch/ubuntu.py
fnordahl: That's indeed much nicer, thx! :+1:

charmhelpers/fetch/ubuntu.py
fnordahl: Good catch

charmhelpers/fetch/ubuntu.py
fnordahl: :+1:

charmhelpers/fetch/ubuntu.py
fnordahl: Yes, if the system or the ``dpkg`` binary causes a unexpected error we would want to know about it. True/False is only valid when we can determine the relationship between the versions.  A updated version is on the way where this function restores it's ternary result (for >, = or <)

charmhelpers/fetch/ubuntu.py
fnordahl: Thx, next patch will provide a subset of the  apt_pkg API the redefinition of this function will no longer be necessary.

fnordahl:
Yes, you are absolutely right in that that would be a much safer path given we do not control all of our consumers. A set of commits that implement this is on the way.

charmhelpers/fetch/ubuntu.py
```diff
@@ -217,13 +219,25 @@ def filter_missing_packages(packages):
...
-    """Build and return an apt cache."""
-    from apt import apt_pkg
-    apt_pkg.init()
-    if in_memory:
-        apt_pkg.config.set("Dir::Cache::pkgcache", "")
-        apt_pkg.config.set("Dir::Cache::srcpkgcache", "")
-    return apt_pkg.Cache(progress)
+    """Shim returning an object simulating the apt_pkg Cache.
+
+    :param in_memory: Kept for compability reasons, has no effect.
```
ajkavanagh: You could just use `*args` in the function signature to show that there might be args, but that you are ignoring them.  (or even `*_` which would make it even more obvious)?

charmhelpers/fetch/ubuntu_apt_pkg.py
```diff
@@ -0,0 +1,245 @@
...
+    """Simple container for attributes."""
+    def __init__(self, attr_map):
+        """Initialize package attribute container.
+
+        :param attr_map: Dictionary key value pairs to transform into
+                         attributes with a value on instance of the class.
+        :type attr_map: Dict[str, str]
+        """
+        for k, v in attr_map.items():
+            setattr(self, k, v)
```
ajkavanagh: This is a cute way of almost getting the same thing:

```python
class AttributeDict(dict): 
    __getattr__ = dict.__getitem__
    __setattr__ = dict.__setitem__
```

i.e. make `x.attr` work like `x['attr']`.  However, you may be aiming to achieve another objective.

charmhelpers/fetch/ubuntu.py
fnordahl: :+1:

charmhelpers/fetch/ubuntu_apt_pkg.py
fnordahl: That's very neat, thank you for the tip!

charmhelpers/fetch/ubuntu.py
fnordahl: Thank you for the tip, have incorporated this in the function now.

charmhelpers/fetch/ubuntu.py
fnordahl: Thank you for the tip, have incorporated a slightly modified version of this in the function now. (I am interested in the first part of the split, not the last)

charmhelpers/fetch/ubuntu.py
```diff
@@ -216,14 +218,26 @@ def filter_missing_packages(packages):
...
 
-def apt_cache(in_memory=True, progress=None):
-    """Build and return an apt cache."""
-    from apt import apt_pkg
-    apt_pkg.init()
-    if in_memory:
-        apt_pkg.config.set("Dir::Cache::pkgcache", "")
-        apt_pkg.config.set("Dir::Cache::srcpkgcache", "")
-    return apt_pkg.Cache(progress)
+def apt_cache(*_):
```
johnsca: In case someone is calling this externally using keyword-style params, it might be worth handling those.  Something like:

```python
def apt_cache(*_, **__):
```

NB: Apparently, you can't reuse a single underscore as a placeholder name for multiple params in a function def.

fnordahl: Good catch, I'll post a commit to address that just in case.

PR #343, opened 2019-07-04
Allow enabling the pg autoscaler when the module is enabled

ChrisMacNaughton:


charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -324,12 +325,23 @@ def create(self):
...
                     set_app_name_for_pool(client=self.service,
                                           pool=self.name,
                                           name=self.app_name)
                 except CalledProcessError:
                     log('Could not set app name for pool {}'.format(self.name, level=WARNING))
+                if 'pg_autoscaler' in enabled_manager_modules():
+                    try:
+                        enable_pg_autoscale(self.service, self.name)
+                    except CalledProcessError:
+                        log('Could not configure auto scaling for pool {}'.format(self.name, level=WARNING))
```
fnordahl: For possible future debugging, would it be useful to log returncode and output from command too?

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -382,13 +394,50 @@ def create(self):
...
+                if nautilus_or_later:
+                    # Ensure we set the expected pool ratio
+                    update_pool(client=self.service,
+                                pool=self.name,
+                                settings={'target_size_ratio': str(self.percent_data / 100.0)})
+                if 'pg_autoscaler' in enabled_manager_modules():
+                    try:
+                        enable_pg_autoscale(self.service, self.name)
+                    except CalledProcessError:
+                        log('Could not configure auto scaling for pool {}'.format(self.name, level=WARNING))
```
fnordahl: For possible future debugging, would it be useful to log returncode and output from command too?

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -382,13 +394,50 @@ def create(self):
...
+def enabled_manager_modules():
+    """Return a list of enabled manager modules.
+
+    :returns: list[str]
+    """
+    cmd = ['ceph', 'mgr', 'module', 'ls']
+    try:
+        modules = check_output(cmd)
+    except CalledProcessError as e:
+        log("Failed to list ceph modules: {}".format(e), WARNING)
```
fnordahl: For possible future debugging, would it be useful to log returncode and output from command too?

tests/contrib/storage/test_linux_ceph.py
```diff
@@ -434,6 +480,34 @@ def test_erasure_pool_create(self, get_osds, erasure_profile):
...
+    @patch.object(ceph_utils, 'kv')
+    @patch.object(ceph_utils, 'get_erasure_profile')
+    @patch.object(ceph_utils, 'get_osds')
+    def test_erasure_pool_create_autoscaler(self,
+                                            get_osds,
+                                            erasure_profile,
+                                            kv):
+        db = MagicMock()
+        kv.return_value = db
+        db.get.return_value = ['pg_autoscaler']
```
fnordahl: This unit test appears to do uneccessary mocking now that you query the state directly from Ceph with subprocess in ``enabled_manager_modules``

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -382,13 +394,50 @@ def create(self):
...
                 raise
 
     """Get an existing erasure code profile if it already exists.
        Returns json formatted output"""
 
 
+def enabled_manager_modules():
+    """Return a list of enabled manager modules.
+
+    :returns: list[str]
```
fnordahl: nit: This should be ``:rtype:`` and list is capitalized.

https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html
https://mypy.readthedocs.io/en/latest/builtin_types.html

charmhelpers/contrib/storage/linux/ceph.py
ChrisMacNaughton: At this point, there is no output, and the return data should be logged via the CalledProcessError?

```
out = subprocess.check_output(['/bin/sh', '-c', 'echo test && false'])
  File "/usr/lib/python3.6/subprocess.py", line 356, in check_output
    **kwargs).stdout
  File "/usr/lib/python3.6/subprocess.py", line 438, in run
    output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '['/bin/sh', '-c', 'echo test && false']' returned non-zero exit status 1.
>>> out
NameError: name 'out' is not defined
>>> out = subprocess.check_output(['/bin/sh', '-c', 'echo test && true'])
>>> out
b'test\n'
```

Doing something like this snippet gives us:

```
>>> try:
...   out = subprocess.check_output(['/bin/sh', '-c', 'echo test && false'])
... except subprocess.CalledProcessError as e:
...   print("Error: {}".format(e))
... 
Error: Command '['/bin/sh', '-c', 'echo test && false']' returned non-zero exit status 1.
```

fnordahl: Any output from subprocess when it returns non-zero is available in ``CalledProcessError.output``, but your change at least adds the return code which may be sufficient.

PR #345, opened 2019-07-09
Make NRPE.add_check(shortname=...) optional again

auria:
Commit 859a149 enforced the use of "shortname" in add_check when *args
can carry the same value as first element. This change makes "shortname"
key optional again, as documented at the beginning of the module.

### DIFF
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -266,11 +266,14 @@ class NRPE(object):
         self.remove_check_queue = set()
 
     def add_check(self, *args, **kwargs):
+        shortname = None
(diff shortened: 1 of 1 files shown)


sparkiegeek:
What about `remove_check`, it seems to suffer from the same issue?

What about when there's no argument nor `short_check=...` as a keyword argument?

auria:
@sparkiegeek this fix (f2) reverted the behavior changed in `add_check` by a previous fix (f1). Since no changes to `remove_check` were done on fix "f1", I didn't touch it now on fix "f2".

For such reason, no code should be broken (or no more than it was before any of the fixes), but I would agree that both `add_check` and `remove_check` should allow similar arguments (args or kwargs)... that would make the code simpler.

PR #346, opened 2019-07-10
Add service_{project,domain}_id keys to Ident ctxt

gnuoy:
Add service_{project,domain}_id keys to IdentityServiceContext. The
data is already present in the admin_domain_name and admin_tenant_id
keys but those key names are confusing.

Adding service_{project,domain}_id keys to the context is gated on
the use of keystone api version greater than 2. It is assumed that
if api 3 is being used then the service project and domain ids
are being passed down the relation which is true as of the 18.04
keystone charms (https://review.opendev.org/#/c/548264).

### DIFF
--- a/tests/contrib/openstack/test_os_contexts.py
+++ b/tests/contrib/openstack/test_os_contexts.py
@@ -1216,8 +1216,8 @@ class ContextTests(unittest.TestCase):
     @patch.object(context, 'filter_installed_packages')
     @patch.object(context, 'os_release')
     def test_keystone_authtoken_www_authenticate_uri_stein_apiv3(self, mock_os_release, mock_filter_installed_packages):
-        relation = FakeRelation(
-            relation_data=IDENTITY_SERVICE_RELATION_VERSIONED)
+        relation_data = deepcopy(IDENTITY_SERVICE_RELATION_VERSIONED)
+        relation = FakeRelation(relation_data=relation_data)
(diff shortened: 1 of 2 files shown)


tests/contrib/openstack/test_os_contexts.py
```diff
@@ -1211,9 +1216,8 @@ def test_identity_service_context_with_missing_relation(self, *args):
     @patch.object(context, 'filter_installed_packages')
     @patch.object(context, 'os_release')
     def test_keystone_authtoken_www_authenticate_uri_stein_apiv3(self, mock_os_release, mock_filter_installed_packages):
-        relation_data = deepcopy(IDENTITY_SERVICE_RELATION_UNSET)
-        relation_data['api_version'] = '3'
-        relation = FakeRelation(relation_data=relation_data)
+        relation = FakeRelation(
+            relation_data=IDENTITY_SERVICE_RELATION_VERSIONED)
```
ajkavanagh: Why did you remove the deepcopy?

gnuoy: Because I'm not changing the data

ajkavanagh: But it was a good defensive programming strategy.  As it's a constant, it's a good idea to deepcopy it for a test.  that's why I was curious.

PR #356, opened 2019-08-07
Update the logic for getting Juju version

johnsca:
Since ~ Juju 2.0, the JUJU_VERSION env var has provided the charm info about what version of Juju the charm is running under.  The old hack may not work going forward (particularly on Kubernetes charms).

One caveat of this change is that the env var doesn't include the series and arch info, but I doubt anyone was actually depending on that.


mitechie:
There's code coming into Juju in 2.7 that gives actual git sha information as well which helps for auto updated channels (edge snaps and such) that might be worth considering. Ping'ing @wallyworld who's driven these changes. Not required for this, but worth being aware of.

johnsca:
@ryan-beisner This code is intended for use by charms.  It doesn't look like you're running in a hook context, and most charms don't have `juju` installed on the unit.  I confirmed that this does work for candidate:

```
[johnsca@raza-u:~] $ snap refresh juju --channel=candidate
juju (candidate) 2.6.6 from Canonical✓ refreshed

[johnsca@raza-u:~] $ juju bootstrap lxd lxd-candidate
Creating Juju controller "lxd-candidate" on lxd/localhost
Looking for packaged Juju agent version 2.6.6 for amd64
To configure your system to better support LXD containers, please see: https://github.com/lxc/lxd/blob/master/doc/production-setup.md
Launching controller instance(s) on localhost/localhost...
 - juju-186509-0 (arch=amd64)          
Installing Juju agent on bootstrap instance
Fetching Juju GUI 2.14.0
Waiting for address
Attempting to connect to 10.4.23.103:22
Connected to 10.4.23.103
Running machine configuration script...
Bootstrap agent now started
Contacting Juju controller at 10.4.23.103 to verify accessibility...

Bootstrap complete, controller "lxd-candidate" now is available
Controller machines are in the "controller" model
Initial model "default" added

[johnsca@raza-u:~] $ juju deploy cs:ubuntu-0
Located charm "cs:ubuntu-0".
Deploying charm "cs:ubuntu-0".

[johnsca@raza-u:~] 2m6s $ juju run --unit ubuntu/0 -- env | grep JUJU_VERSION
JUJU_VERSION=2.6.6
```

johnsca:
@mitechie This is primarily used for feature detection / selection (and really only in the cases where the "optimistically try it and handle errors" approach won't work for some reason), so I would think that major / minor is all people really care about.

johnsca:
NB: This came from discussion about the upcoming change in format for `pod-spec-set` YAML for K8s charms.

PR #360, opened 2019-08-19
fetch: Override apt execution environment

fnordahl:
Override the execution environment for the apt_* family of
functions with a sanitized environment.
    
The need for this rises from consumers of the library running the
code from within a virtualenv.  In such a situation a deb
post-install script interacting with Python may interact with the
virtualenv instead of the system Python environment.
    
Clean up a pre-existing left behind unit test change after update
to _run_with_retries().
    
LP: #1840257

### DIFF
--- a/charmhelpers/core/host.py
+++ b/charmhelpers/core/host.py
@@ -1075,3 +1075,32 @@ def install_ca_cert(ca_cert, name=None):
     log("Installing new CA cert at: {}".format(cert_file), level=INFO)
     write_file(cert_file, ca_cert)
     subprocess.check_call(['update-ca-certificates', '--fresh'])
+
+
+def get_system_env(key, default=None):
+    """Get data from system environment as represented in ``/etc/environment``.
+
+    :param key: Key to look up
+    :type key: str
+    :param default: Value to return if key is not found
+    :type default: any
+    :returns: Value for key if found or contents of default parameter
+    :rtype: any
+    :raises: subprocess.CalledProcessError, KeyError
+    """
+    env_file = '/etc/environment'
+    # use the shell and env(1) to parse the global environments file.  This is
+    # done to get the correct result even if the user has shell variable
+    # substitutions or other shell logic in that file.
+    output = subprocess.check_output(
+        ['env', '-i', '/bin/bash', '-c',
+         'set -a && source {} && env'.format(env_file)],
+        universal_newlines=True)
+    for k, v in (line.split('=', 1)
+                 for line in output.splitlines() if '=' in line):
+        if k.upper() == key.upper():
+            return v
+    else:
+        if default:
+            return default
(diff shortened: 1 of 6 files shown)


charmhelpers/fetch/ubuntu.py
```diff
@@ -242,8 +242,20 @@ def apt_cache(*_, **__):
     return ubuntu_apt_pkg.Cache()
 
 
-def apt_install(packages, options=None, fatal=False):
-    """Install one or more packages."""
+def apt_install(packages, options=None, fatal=False, override_env=None):
+    """Install one or more packages.
+
+    :param packages: Package(s) to install
+    :type packages: Option(str, List)
```
ajkavanagh: The type here is (I think): `Option[str, List[str]]`

I'm not really sure why mypy went for square brackets `[]` vs parenthesis `()`, but sadly, and unintuitively, they did.  See http://mypy-lang.org/ for more details.

charmhelpers/fetch/ubuntu.py
```diff
@@ -242,8 +242,20 @@ def apt_cache(*_, **__):
     return ubuntu_apt_pkg.Cache()
 
 
-def apt_install(packages, options=None, fatal=False):
-    """Install one or more packages."""
+def apt_install(packages, options=None, fatal=False, override_env=None):
```
ajkavanagh: I'm not sure about the `override_env=None` here.  My rationale is that I think that a function like `apt_install` should *by default* use a sanitized environment, and not expect the caller to work out what environment is safe for `apt_install` to work in.

I think this is true for all the functions in this module, that have been modified with the `override_env` module.  I would suspect that the cleaned environment of `/etc/environment`'s path should be enough to ensure `apt_*` functions work?

fnordahl: Ok, let's do it, we're well into changing the world this cycle anyway.

charmhelpers/fetch/ubuntu.py
fnordahl: :+1:

charmhelpers/core/host.py
```diff
@@ -1075,3 +1075,32 @@ def install_ca_cert(ca_cert, name=None):
     log("Installing new CA cert at: {}".format(cert_file), level=INFO)
     write_file(cert_file, ca_cert)
     subprocess.check_call(['update-ca-certificates', '--fresh'])
+
+
+def get_system_env(key, default=None):
```
timClicks: To maintain mental consistency with `dict.get()`, I would return `None` rather than raise a `KeyError` error on a method that looks like `get_`.

fnordahl: Good point @timClicks, the intention was to be able to require success, but I guess that can be done by the caller by checking the return rather than, perhaps surprisingly, raising a KeyError.

PR #362, opened 2019-08-21
Adding duplicate check for ops.

hyperbolic2346:
This came about due to changes to ceph ops to allow creating multiple pools.
The thought of multiple calls with the same pool names came up. While this
prevents multiple ops from the same relation, it isn't going to prevent two
different applications from attempting to create multiple pools. The point
of that comment is to point out that deplication is still possible.


hyperbolic2346:
test_ceph_broker_rq_class is actually making a duplicate add-permissions-to-key op named service_test. I'm not sure if this is valid.
```
            'ops': [{'op': 'create-pool', 'name': 'pool1', 'replicas': 1},
                    {'op': 'create-pool', 'name': 'pool2', 'replicas': 3},
                    {'op': 'create-pool', 'name': 'pool3', 'replicas': 3, 'group': 'test'},
                    {'op': 'add-permissions-to-key', 'group': 'test', 'name': 'service_test'},
                    {'op': 'add-permissions-to-key', 'group': 'objects', 'name': 'test'},
                    {
                        'op': 'add-permissions-to-key',
                        'group': 'others',
                        'name': 'service_test',
                        'object-prefix-permissions': {u'rwx': [u'prefix1']}}]
```

I'm unsure about the second failure where the two ops shouldn't be the same but are.

ChrisMacNaughton:
It should be valid to make duplicate (by key) requests to add permissions, the idea is that a given key can have permissions added. In this case, it looks like it's adding:

- access to pool3, via the "test" group
- rwx permissions onto objects with the prefix "prefix1" in the "objects" group

PR #363, opened 2019-08-21
Refactor ceph broker messaging methods

rodrigogansobarbieri:
Related-Bug: lp#1835045

Due to ambiguous behavior of is_broker_action_done function
(returning False for when there was not a broker response and
also for when it was not marked done) causing unnecessary
restarts in nova-compute service, the ceph broker messaging
is being refactored in charm-nova-compute.

In order to improve is_broker_action_done, it was also necessary
to improve the logic of how the broker response is obtained and
handled.

The broker response was always initially obtained by
get_request_states function while iterating over all
relation data. However, it was used to mark the request as
complete and then had to be obtained again for other functions.

Other functions were obtaining the broker response by reading the
relation data either specifically with rid and unit (when
looping through all charm units during a config-changed), or with
the remote context (when invoked by a relation-changed hook).

During the latter, the broker response could not be obtained by
is_broker_action_done unless on the appropriate remote context,
even though it was already being obtained by get_request_states.

Therefore under those circunstances the code execution was
pointless and causing undesired side-effects (nova-compute
restarts).

This refactor aims to allow the invoking charms to run
get_request_state only once per hook execution and save the
broker response regardless of remote context. Using the
saved broker response, is_broker_action_done can be invoked
passing the broker response directly and both the ambiguous
behavior and the pointless executions with side-effects are
eliminated.

Since this affects all charms that use helper functions for
ceph broker messaging, a summary of changes in each function
is provided:

- find_broker_rsp_for_completed_request replaces the function
  is_request_complete_for_rid, whereas now it will return the
  rid, unit and broker response when it finds a completed request.
  The rid parameter is now also optional, as it will iterate over
  all relation ids if the parameter is not provided.

- get_request_states now additionally returns the unit and
  broker response when the request is complete. This avoids the
  need to re-iterate across all relation data to get the broker
  response.

- is_request_sent now accepts a states parameter in order to not
  run get_request_states again if it has already been run.

- is_request_complete now acepts a states parameter in order to
  not run get_request states again if it has already been run.

- get_broker_rsp_from_completed_request is a new function that
  returns the broker response for a completed request through
  get_request_states. It also accepts a states parameter to
  avoid running get_request_states again if it has already been
  run.

- send_request_if_needed has part of its functionality
  separated into a new function send_broker_request so it
  can be invoked directly. It now also accepts a states
  parameter to pass over to is_request_sent, therefore
  avoid running get_request_states again if it has already been
  run.

- is_broker_action_done now requires a broker response as a
  parameter. Therefore it does not need to care about scenarios
  where a broker response is not present. There is only one
  scenario that can return True or False.

- mark_broker_action_done now requires a broker response as a
  parameter. Therefore it does not need to care about scenarios
  where a broker response is not present.

### DIFF
--- a/charmhelpers/contrib/storage/linux/ceph.py
+++ b/charmhelpers/contrib/storage/linux/ceph.py
@@ -1437,19 +1437,24 @@ def get_request_states(request, relation='ceph'):
 
     @param request: A CephBrokerRq object
     """
-    complete = []
     requests = {}
     for rid in relation_ids(relation):
         complete = False
+        broker_unit = None
+        broker_rsp = None
         previous_request = get_previous_request(rid)
         if request == previous_request:
             sent = True
-            complete = is_request_complete_for_rid(previous_request, rid)
+            _, broker_unit, broker_rsp = find_broker_rsp_for_completed_request(
+                previous_request, rid=rid)
+            if broker_rsp:
+                complete = True
         else:
             sent = False
-            complete = False
 
         requests[rid] = {
+            'unit': broker_unit,
+            'broker_rsp': broker_rsp,
             'sent': sent,
             'complete': complete,
         }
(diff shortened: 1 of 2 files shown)


rodrigogansobarbieri:
rebased

rodrigogansobarbieri:
rebased again.

Gerrit patches:
https://review.opendev.org/679861 - Adds ceph backend and functional tests to charm-nova-compute CI.
https://review.opendev.org/677818 - Updates nova-compute to use the new methods from this pull request (and tests with the new functional tests). Official patch that needs the PR to merged to pass tests.
https://review.opendev.org/681073 - [DEBUG] patch to test the nova-compute patch above checking out the PR update, in order to test it.

ajkavanagh:
The associated reviews have been abandoned on gerrit and there's been no progress on this for 3 years.  Please feel free to re-open as necessary, but do note the comments on the gerrit reviews as well.

PR #366, opened 2019-08-27
Decode base64 certificates before using them.

Pekkari:
lp: #1841429

Signed-off-by: José Pekkarinen <jose.pekkarinen@canonical.com>

### DIFF
--- a/charmhelpers/contrib/hahelpers/apache.py
+++ b/charmhelpers/contrib/hahelpers/apache.py
@@ -23,6 +23,7 @@
 #
 
 import os
+import base64
 
 from charmhelpers.core import host
 from charmhelpers.core.hookenv import (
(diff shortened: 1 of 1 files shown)


charmhelpers/contrib/hahelpers/apache.py
```diff
@@ -37,8 +38,8 @@
 
 def get_cert(cn=None):
     # TODO: deal with multiple https endpoints via charm config
-    cert = config_get('ssl_cert')
-    key = config_get('ssl_key')
+    cert = base64.b64decode(config_get('ssl_cert'))
```
fnordahl: This is a helper function to reliably get the base64 encoded data from config/relations.  It is consumed by the ApacheSSLContext [0] which does the base64 decoding.

0: https://github.com/juju/charm-helpers/blob/9248c1a06dfbdfba21efeddd02af13fd7b58386e/charmhelpers/contrib/openstack/context.py#L959

Pekkari: well, according to lp:#1841429 this it doesn't.

fnordahl:
Re discussion in LP: 1841429

PR #370, opened 2019-09-03
Conditionally add port_forwarding to l3_extension_plugins

dshcherb:
Adds port_forwarding which is an extension to an l3 agent to support the
port forwarding feature.

See LP: #1842353

Partial-Bug: #1842353


dshcherb:
https://review.opendev.org/#/c/679694/ - for context

PR #371, opened 2019-09-08
Allow overriding usage of cache for hookenv config

hloeung:
In certain cases, we may not want to use the cached results from
config-get such as when hook is run as a non-root user and does not
have access to the cache file (currently charm/.juju-persistent-config).

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -391,7 +391,7 @@ class Config(dict):
 _cache_config = None
 
 
-def config(scope=None):
+def config(scope=None, use_cache=True):
     """
     Get the juju charm configuration (scope==None) or individual key,
     (scope=str).  The returned value is a Python data structure loaded as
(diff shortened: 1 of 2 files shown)


hloeung:
https://bugs.launchpad.net/ubuntu-repository-cache/+bug/1828658

ajkavanagh:
I don't think this is correct, but I may be misunderstanding the use case.  `.juju-persistent-config` *ought* to only be read by root, as charms only ever run as root (I think)?  i.e. there might be sensitive data in the config that shouldn't be world readable?  I guess the main question is why a hook would be run as a non-root user?

hloeung:
Right, in this case, the u-r-c charm switches uids to a non-root user - it still has access to the JUJU_* environment variables and able to call config-get etc. This looks like a one off and likely just poorly designed in the u-r-c charm.

In any case, I'll close this PR off.

PR #380, opened 2019-10-03
Allow OpenStack deployments from PPA packages

thedac:
When using a PPA for an OpenStack package the current implementation
throws errors because it cannot determine the OpenStack version from
the openstack-origin. It also makes terrible assumptions about which
version of OpenStack is in use when unable to determine it from
openstack-origin.

This change does the following:
 * Gracefully handles if the openstack-origin is not a known source.
 * Uses a sane default for os_relase base, by determining the default
   OpenStack release for the currently deployed Ubuntu series.


javacruft:
Some unit tests are failing which look related to this PR

PR #381, opened 2019-10-04
Stop duplicate ops being added to CephBrokerRq

gnuoy:
If an op already exists in a CephBrokerRq then there is no
reason to add it again.

### DIFF
--- a/charmhelpers/contrib/storage/linux/ceph.py
+++ b/charmhelpers/contrib/storage/linux/ceph.py
@@ -1185,6 +1185,17 @@ class CephBrokerRq(object):
             self.request_id = str(uuid.uuid1())
         self.ops = []
 
+    def add_op(self, op):
+        """Add an op if it not already in the list.
+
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -1185,6 +1185,17 @@ def __init__(self, api_version=1, request_id=None):
...
+    def add_op(self, op):
+        """Add an op if it is not already in the list.
+
+        :param op: Operation to add.
+        :type op: dict
+        """
+        for existing_op in self.ops:
+            if op == existing_op:
+                return
+        self.ops.append(op)
```
javacruft: Won't this do the same:

if op not in self.ops:
   self.ops.append(op)

gnuoy: Err, yes, thats much nicer.

PR #390, opened 2019-11-11
ufw: Allow use of `prepend` keyword when inserting rules

fnordahl:


charmhelpers/contrib/network/ufw.py
```diff
@@ -236,29 +237,45 @@ def default_policy(policy='deny', direction='incoming'):
...
     :param action: `allow` or `delete`
+    :type action: str
     :param index: if different from None the rule is inserted at the given
                   `index`.
+    :type index: Optional[int]
+    :param prepend: Whether to insert the rule before all other rules matching
+                    the rule's IP type.
+    :type prepend: bool
+    :param comment: Create the rule with a comment
+    :type comment: Optional[str]
```
ajkavanagh: TIL about `Optional`.  I've been using `Union[None,...]` but this is better and more correct.

charmhelpers/contrib/network/ufw.py
```diff
@@ -35,8 +35,9 @@
   >>> ufw.enable()
   >>> ufw.service('4949', 'close')  # munin
 """
-import re
 import os
+import re
```
ajkavanagh: Is `re` being used?

PR #393, opened 2019-11-15
Policyd library changes to support openstack-dashboard

ajkavanagh:
OpenStack dashboard (horizon) has a much more complex policy
overrides concept and these changes are to support that.

Also a slight refactor on the
maybe_do_policyd_overrides_on_config_changed as there was much
duplicated code between that and maybe_do_policyd_overrides.

### DIFF
--- a/charmhelpers/contrib/openstack/policyd.py
+++ b/charmhelpers/contrib/openstack/policyd.py
@@ -685,9 +685,9 @@ def process_policy_resource_file(resource_file,
     black list.
 
     The yaml filename can be modified in two ways.  If the `preserve_topdir`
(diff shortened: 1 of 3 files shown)


charmhelpers/contrib/openstack/policyd.py
```diff
@@ -638,6 +684,16 @@ def process_policy_resource_file(resource_file,
     its file path reconstructed.  This, also, must not match any path in the
     black list.
 
+    The yaml filename can be modified in two ways.  If the `preserve_topdir`
+    param is True, then files will be flattened to the this top dir.  This
```
ChrisMacNaughton: "to the this" -> "to the"

PR #394, opened 2019-11-15
Add function-get/set/fail methods alongside the action ones

wallyworld:
We have methods for action_set(), action_get(), action_fail().
Deprecate these and add equivalent function_set(), function_get(), function_fail() versions.
Also, rename the new but not yet used action_log() method to function_log()


charmhelpers/core/hookenv.py
```diff
@@ -967,6 +967,8 @@ def charm_dir():
 @cached
 def action_get(key=None):
     """Gets the value of an action parameter, or all key/value param pairs"""
+    log("action_get is deprecated. Please consider using function_get "
+        "instead.", level=WARNING)
```
ajkavanagh: There is a handy `deprecate` decorator in the `charmhelpers/__init__.py` which can be used; I'm not sure how it plays with cached, so it should probably go after the `@cached` decorator to maximise caching.

wallyworld: Thanks! Fixed.

charmhelpers/core/hookenv.py
```diff
@@ -983,28 +1010,76 @@ def action_set(values):
...
     """Get the name of the currently executing action."""
     return os.environ.get('JUJU_ACTION_NAME')
 
 
+def function_name():
+    """Get the name of the currently executing function."""
+    name = os.environ.get('JUJU_FUNCTION_NAME')
+    if name:
+        return name
+    return action_name()
```
ycliuhw: we can do
```python
    return os.environ.get('JUJU_FUNCTION_NAME') or action_name()
```

charmhelpers/core/hookenv.py
```diff
@@ -119,19 +121,19 @@ def log(message, level=None):
             raise
 
 
-def action_log(message):
-    """Write an action progress message"""
-    command = ['action-log']
+def function_log(message):
```
ycliuhw: do we need to keep `action_log`?

wallyworld: no since it was only added during rc and is not used

charmhelpers/core/hookenv.py
wallyworld: derp, of course

ycliuhw:
ah the tests failed, mock `cmd_exist`?

wallyworld:
> ah the tests failed, mock `cmd_exists `?

Updated the tests

charmhelpers/core/hookenv.py
```diff
@@ -964,7 +966,16 @@ def charm_dir():
     return os.environ.get('CHARM_DIR')
 
 
+def cmd_exists(cmd):
+    """Return True if the specified cmd exists in the path"""
+    return any(
+        os.access(os.path.join(path, cmd), os.X_OK)
+        for path in os.environ["PATH"].split(os.pathsep)
+    )
```
johnsca: Could use [`shutil.which(cmd)`](https://docs.python.org/3/library/shutil.html?highlight=.which#shutil.which)

charmhelpers/core/hookenv.py
```diff
@@ -964,7 +966,16 @@ def charm_dir():
...
+
+
 @cached
+@deprecate("moved to function_get()", "2019-11", log=log)
```
johnsca: I didn't even realize we had this decorator.  I like it, but I think it's more important for the deprecation message to go into the docs.  I'd recommend using the syntax used by [`service_name`](https://github.com/juju/charm-helpers/blob/master/charmhelpers/core/hookenv.py#L237).  It would be nice if the decorator was automatically turned into the doc addition as well, but that's for future work.

Also, the date is supposed to be when the function will actually be removed from the code base, not when it is deprecated, though I don't know if anything will ever actually be removed, so it's probably fine as is.

charmhelpers/core/hookenv.py
```diff
@@ -975,6 +986,22 @@ def action_get(key=None):
     return action_data
 
 
+@cached
+def function_get(key=None):
+    """Gets the value of an action parameter, or all key/value param pairs"""
+    # Fallback for older charms.
+    if not cmd_exists('function-get'):
+        return action_get(key)
```
johnsca: Hrm, this will lead to deprecation warnings being emitted even if the charm is using the correct code.  Seems like it would be better to just change the value of `cmd`.

charmhelpers/core/hookenv.py
wallyworld: this doesn't exist in Python 2 so I used something that would work in all cases

charmhelpers/core/hookenv.py
wallyworld: I removed the date arg and added "deprecated" to the doc strings

charmhelpers/core/hookenv.py
wallyworld: fixed

charmhelpers/core/hookenv.py
```diff
@@ -964,9 +966,22 @@ def charm_dir():
...
+
 @cached
+@deprecate("moved to function_get()", log=log)
 def action_get(key=None):
-    """Gets the value of an action parameter, or all key/value param pairs"""
+    """Gets the value of an action parameter, or all key/value param pairs.
+
+    deprecated:: 0.20.5
+    Alias for :func:`function_get`.
+    """
```
johnsca: The syntax for this should be:

```python
def action_get(key=None):
    """
    .. deprecated:: 0.20.7
       Alias for :func:`application_name`.

    Gets the value of an action parameter, or all key/value param pairs.
    """
```

wallyworld: thanks, fixed

PR #395, opened 2019-11-20
MySQL configuration handling

thedac:
Generalize configuration helper to support both mysql-innodb-cluster and
percona-cluster.

### DIFF
--- a/charmhelpers/contrib/database/mysql.py
+++ b/charmhelpers/contrib/database/mysql.py
@@ -471,7 +471,8 @@ class MySQLHelper(object):
         return password
 
 
-class PerconaClusterHelper(object):
+class MySQLConfigHelper(object):
+    """Base configuration helper for MySQL."""
 
     # Going for the biggest page size to avoid wasted bytes.
     # InnoDB page size is 16MB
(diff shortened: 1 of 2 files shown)


thedac:
Tested against mysql-innodb-cluster and percona-cluster charms.

charmhelpers/contrib/database/mysql.py
```diff
@@ -547,36 +548,48 @@ def get_mem_total(self):
...
-            mysql_config['innodb_flush_log_at_trx_commit'] = \
-                self.INNODB_FLUSH_CONFIG_VALUES.get(config['tuning-level'], 1)
+        :returns: Numeric value for innodb_flush_log_at_trx_commit
+        :rtype: int
+        """
+        if config_get('innodb-flush-log-at-trx-commit'):
+            return config_get('innodb-flush-log-at-trx-commit')
+        elif config_get('tuning-level'):
+            return self.INNODB_FLUSH_CONFIG_VALUES.get(
+                config_get('tuning-level'), 1)
```
ajkavanagh: This code has a subtle potential bug (it's a partial function);  if neither `if` is truthy then the function returns `None` (silently).

Either this is expected (in which case, `:rtype: Optional[int]` or it should raise an `Exception` for the situation when neither of the conditions is true.

Also, repeated calls to config_get, which could be rationalised to 2 with intermediate vars.

charmhelpers/contrib/database/mysql.py
```diff
@@ -547,36 +548,48 @@ def get_mem_total(self):
...
+        INNODB_VALID_BUFFERING_VALUES to get the innodb_change_buffering value.
 
-        # Set a sane default key_buffer size
-        mysql_config['key_buffer'] = self.human_to_bytes('32M')
+        :returns: String value for innodb_change_buffering.
+        :rtype: str
+        """
+        _icb = config_get('innodb-change-buffering')
+        if _icb and _icb in self.INNODB_VALID_BUFFERING_VALUES:
+            return _icb
```
ajkavanagh: Again, this is a partial function; it returns `None` if the if statement isn't True.  Either `:rtype: Optional[str]` or raise an `Exception` or return an empty `""`.

charmhelpers/contrib/database/mysql.py
```diff
@@ -547,36 +548,48 @@ def get_mem_total(self):
...
+
+        :returns: Numeric value for innodb_buffer_pool_size.
+        :rtype: int
+        """
         total_memory = self.human_to_bytes(self.get_mem_total())
 
-        dataset_bytes = config.get('dataset-size', None)
-        innodb_buffer_pool_size = config.get('innodb-buffer-pool-size', None)
+        dataset_bytes = config_get('dataset-size')
+        innodb_buffer_pool_size = config_get('innodb-buffer-pool-size')
```
ajkavanagh: Does this definitely return an `int`; i.e. never a `None`?

charmhelpers/contrib/database/mysql.py
thedac: Updating the docstring

charmhelpers/contrib/database/mysql.py
thedac: Updating the docstring

charmhelpers/contrib/database/mysql.py
thedac: This always returns an int.

PR #397, opened 2019-11-22
MySQL Helper Singleton

thedac:
Create singleton to avoid multiple instantiations of the class.

### DIFF
--- a/charmhelpers/contrib/database/mysql.py
+++ b/charmhelpers/contrib/database/mysql.py
@@ -471,6 +471,18 @@ class MySQLHelper(object):
         return password
 
 
+# `_singleton_config_helper` stores the instance of the helper class that is
+# being used during a hook invocation.
+_singleton_config_helper = None
+
+
+def get_mysql_config_helper(cls):
+    global _singleton_config_helper
+    if _singleton_config_helper is None:
+        _singleton_config_helper = cls()
+    return _singleton_config_helper
+
+
 class MySQLConfigHelper(object):
     """Base configuration helper for MySQL."""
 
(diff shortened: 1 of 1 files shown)


charmhelpers/contrib/database/mysql.py
```diff
@@ -471,6 +471,18 @@ def configure_db(self, hostname, database, username, admin=False):
...
+# `_singleton_config_helper` stores the instance of the helper class that is
+# being used during a hook invocation.
+_singleton_config_helper = None
+
+
+def get_mysql_config_helper(cls):
+    global _singleton_config_helper
+    if _singleton_config_helper is None:
+        _singleton_config_helper = cls()
+    return _singleton_config_helper
```
ajkavanagh: I wouldn't pass the class in it means the caller could mess with the singleton.  Also, once set it would no longer have a value.  I'd just code it as:

```python
def get_mysql_config_helper():
    global _singleton_config_helper
    if _singleton_config_helper is None:
        _singleton_config_helper = MySQLConfigHelper()
    return _singleton_config_helper
```

charmhelpers/contrib/database/mysql.py
```diff
@@ -496,6 +508,15 @@ class MySQLConfigHelper(object):
...
 
+    @property
+    def singleton_config_helper(self):
+        """Either returns the already created helper, or create a new one.
+        """
+        global _singleton_config_helper
+        if _singleton_config_helper is None:
+            _singleton_config_helper = get_mysql_config_helper(self.__class__)
+        return _singleton_config_helper
+
```
ajkavanagh: You probably don't need this as well.  It's either/or really.  To do it this way it could be coded as:

```python
class MySQLConfigHelper(object):

    _singleton_config_helper = None

    def __init__(self):
        if self._singleton_config_helper is not None:
            raise RuntimeError("Mustn't instantiate more than once ...")
        # ... and other init stuff

    @property
    def singleton(self):
        """Either returns the already created helper, or create a new one.
        """
        if self._singleton_config_helper is None:
            self._singleton_config_helper = MySQLConfigHelper()
        return self._singleton_config_helper
```

You can even go as far as patching out the `__init__()` on the class during the class, but you don't really need to do that ... :)

---

So you can either go with the module global and a global function, or with the class global and a property singleton to achieve the same effect.

---

My personal preference is the module global/module function, rather than class global/class singleton.  But, that doesn't make the class version more/less bad; it's simply that I prefer functional over object-orientated programming, although strictly speaking, this is "programming with objects" rather than "object-orientated" :)

PR #398, opened 2019-11-26
Update now-deprecated usage of platform.dist

### LINKED ISSUE #293: PendingDeprecationWarning: dist() and linux_distribution() functions are deprecated in Python 3.5
`charmhelpers/osplatform.py:13: PendingDeprecationWarning: dist() and linux_distribution() functions are deprecated in Python 3.5`

Seeing this warning. Looks like it's being deprecated in favor of distro.linux_distribution().

techalchemy:
- Rely instead on `distro.name` which is also `python2.7` compatible
- `distro.name` does not return a tuple but instead only the name of the
  distro which seems to be what is needed
- Fixes #293

Signed-off-by: Dan Ryan <dan.ryan@canonical.com>

### DIFF
--- a/charmhelpers/osplatform.py
+++ b/charmhelpers/osplatform.py
@@ -1,4 +1,5 @@
-import platform
+# -*- coding=utf-8 -*-
+import distro
 
 
 def get_platform():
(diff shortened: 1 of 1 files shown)


techalchemy:
Can be closed in favor of #423 when merged

PR #400, opened 2019-11-29
Yum + LSB fix making package installation on CentOS possible

hallback:
When using Python 3, package installation with fetch in charmhelpers does not work since the Python module "yum" is only available for Python2. Simple tasks like checking if packages are installed can easily be performed by RPM instead.

Furthermore, package installations on CentOS were previously carried out with apt (which fails), this has been fixed.

LSB: The variable DISTRIB_CODENAME used for checking if we're using a systemd system is not available in /etc/os-release on CentOS, which throws a KeyError. This information is available in /etc/os-release from CentOS/RHEL 7 and above, which also use systemd.

### DIFF
--- a/charmhelpers/core/host.py
+++ b/charmhelpers/core/host.py
@@ -313,8 +313,17 @@ SYSTEMD_SYSTEM = '/run/systemd/system'
 
 def init_is_systemd():
     """Return True if the host system uses systemd, False otherwise."""
-    if lsb_release()['DISTRIB_CODENAME'] == 'trusty':
-        return False
+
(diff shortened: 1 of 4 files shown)


erik78se:
LGTM

hallback:
> LGTM

Thanks @ajkavanagh, I think Rick Harding will bring this up with you later today, just a heads up.

jamesbeedy:
Hey guys, how is this coming along?

mitechie:
> Hey guys, how is this coming along?

I think we were looking for the branch to get updated so it was up to date with latest trunk?

PR #401, opened 2019-12-06
Rework vault token management

dosaboy:
The code in contrib.openstack.vaultlocker that handles token
is not tolerant of eventual consistency across the remote
units of vault (i.e. the provider). This change allows the
consumer to try tokens, log them as used so they don't get
reused and keep trying available tokens until one succeeds.

Related-Bug: #1849323

### DIFF
--- a/charmhelpers/contrib/openstack/vaultlocker.py
+++ b/charmhelpers/contrib/openstack/vaultlocker.py
@@ -37,9 +37,13 @@ class VaultKVContext(context.OSContextGenerator):
         )
 
     def __call__(self):
+        import hvac
+        ctxt = {}
+        # NOTE(hopem): see https://bugs.launchpad.net/charm-helpers/+bug/1849323
         db = unitdata.kv()
-        last_token = db.get('last-token')
+        # currently known-good secret-id
         secret_id = db.get('secret-id')
+
         for relation_id in hookenv.relation_ids(self.interfaces[0]):
             for unit in hookenv.related_units(relation_id):
                 data = hookenv.relation_get(unit=unit,
@@ -54,27 +58,48 @@ class VaultKVContext(context.OSContextGenerator):
 
                     # Tokens may change when secret_id's are being
                     # reissued - if so use token to get new secret_id
-                    if token != last_token:
+                    token_success = False
+                    try:
                         secret_id = retrieve_secret_id(
                             url=vault_url,
                             token=token
                         )
+                        token_success = True
+                    except hvac.exceptions.InvalidRequest:
+                        # Try next
+                        pass
+
+                    if token_success:
                         db.set('secret-id', secret_id)
-                        db.set('last-token', token)
                         db.flush()
 
-                    ctxt = {
-                        'vault_url': vault_url,
-                        'role_id': json.loads(role_id),
-                        'secret_id': secret_id,
-                        'secret_backend': self.secret_backend,
-                    }
-                    vault_ca = data.get('vault_ca')
-                    if vault_ca:
-                        ctxt['vault_ca'] = json.loads(vault_ca)
-                    self.complete = True
-                    return ctxt
-        return {}
+                        ctxt['vault_url'] = vault_url
+                        ctxt['role_id'] = json.loads(role_id)
+                        ctxt['secret_id'] = secret_id
+                        ctxt['secret_backend'] = self.secret_backend
+                        vault_ca = data.get('vault_ca')
+                        if vault_ca:
+                            ctxt['vault_ca'] = json.loads(vault_ca)
+
+                        self.complete = True
+                        break
+                    else:
+                        if secret_id:
+                            ctxt['vault_url'] = vault_url
+                            ctxt['role_id'] = json.loads(role_id)
+                            ctxt['secret_id'] = secret_id
+                            ctxt['secret_backend'] = self.secret_backend
+                            vault_ca = data.get('vault_ca')
+                            if vault_ca:
+                                ctxt['vault_ca'] = json.loads(vault_ca)
+
+            if self.complete:
+                break
+
+        if ctxt:
+            self.complete = True
+
+        return ctxt
 
 
 def write_vaultlocker_conf(context, priority=100):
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/vaultlocker.py
```diff
@@ -52,29 +71,70 @@ def __call__(self):
...
+
+                        self.complete = True
+                    else:
+                        if token not in used_tokens:
+                            used_tokens[token] = None
+                        elif used_tokens[token] == secret_id:
+                            ctxt['vault_url'] = vault_url
+                            ctxt['role_id'] = json.loads(role_id)
+                            ctxt['secret_id'] = secret_id
+                            ctxt['secret_backend'] = self.secret_backend
```
johnsca: Should probably also handle changes to the `vault_ca` in this case.

charmhelpers/contrib/openstack/vaultlocker.py
```diff
@@ -52,29 +71,70 @@ def __call__(self):
...
+
+        # if we had a successful run, purge old tokens
+        if self.complete and "sentinel" not in current_tokens:
+            for token in used_tokens.keys():
+                if token not in current_tokens:
+                    del used_tokens[token]
```
tpsilva: Changing a dictionary while iterating over its keys() throws a RuntimeError with Python 3. Can be reproduced with a "refresh-secrets" action on a deployment with ceph-osd and vault in HA: https://pastebin.ubuntu.com/p/QN3Yr945VD/

charmhelpers/contrib/openstack/vaultlocker.py
```diff
@@ -52,29 +71,70 @@ def __call__(self):
...
+                    if token not in current_tokens:
+                        if "sentinel" in current_tokens:
+                            current_tokens = [token]
+                        else:
+                            current_tokens.append(token)
+
+                    # If we have already had a successful request then all we
+                    # are doing is collecting current tokens so that we can
+                    # purge old ones.
+                    if self.complete:
```
tpsilva: Superclass (OSContextGenerator) already has an attribute called "complete", that's why this is not currently failing for not being initialized. Please change this attribute name and initialize it.

charmhelpers/contrib/openstack/vaultlocker.py
```diff
@@ -37,9 +37,28 @@ def __init__(self, secret_backend=None):
...
         db = unitdata.kv()
+        # get last-token (back-compat)
         last_token = db.get('last-token')
+        # currently known-good secret-id
         secret_id = db.get('secret-id')
+        # log of tokens we see in this iteration
+        current_tokens = ["sentinel"]
+        # log of tokens we have already tried and, if known, the secret-id
+        # retreived.
+        used_tokens = db.get('used-token')
```
tpsilva: used_tokens = db.get('used-tokens') # Missing s at the end

charmhelpers/contrib/openstack/vaultlocker.py
```diff
@@ -37,9 +37,28 @@ def __init__(self, secret_backend=None):
...
+        import hvac
+        ctxt = {}
+        # NOTE(hopem): see https://bugs.launchpad.net/charm-helpers/+bug/1849323
         db = unitdata.kv()
+        # get last-token (back-compat)
         last_token = db.get('last-token')
+        # currently known-good secret-id
         secret_id = db.get('secret-id')
+        # log of tokens we see in this iteration
+        current_tokens = ["sentinel"]
```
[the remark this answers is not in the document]
tpsilva: +1. I think the whole logic to control current_tokens can be removed since this variable is supposed to store the values of tokens from each vault unit related, so the condition in line 74 will never be false (because tokens are unique for each vault unit). And at line 126 this list will only be empty in case vault was not initialized and did not provide tokens. After initializing, the list will always have at least one element on each run.

If I'm not missing something, I'm +1 on removing the whole logic and just appending the current tokens to the list.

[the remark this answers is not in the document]
dosaboy: My original rationale for using a sentinel is to differentiate between an empty list that is empty because there are no tokens available and one that is empty because for whatever reason there are missing settings on the  relation that mean that we could not get to read the token i.e. this code:

if all([vault_url, role_id, token]):

Upon review I think it is indeed extra complexity without gain so I will remove it.

charmhelpers/contrib/openstack/vaultlocker.py
dosaboy: ack ill fix that

charmhelpers/contrib/openstack/vaultlocker.py
dosaboy: +1 will fix that

charmhelpers/contrib/openstack/vaultlocker.py
dosaboy: ack

charmhelpers/contrib/openstack/vaultlocker.py
dosaboy: Why do you expect this to fail? That variable is initialised to false by the base class which is what I expect so  don't think there is an issue here.

charmhelpers/contrib/openstack/vaultlocker.py
```diff
@@ -54,27 +58,48 @@ def __call__(self):
...
-                    }
-                    vault_ca = data.get('vault_ca')
-                    if vault_ca:
-                        ctxt['vault_ca'] = json.loads(vault_ca)
-                    self.complete = True
-                    return ctxt
-        return {}
+                        ctxt['vault_url'] = vault_url
+                        ctxt['role_id'] = json.loads(role_id)
+                        ctxt['secret_id'] = secret_id
```
javacruft: This secret-id needs to be persisted as its now the most recent one that's been retrieved.

javacruft: Oh wait I see that above now!

PR #405, opened 2019-12-14
openstack: add notification_topics option

skatsaounis:
This new option introduced will give to the charms ability to set
oslo notification topics.

Signed-off-by: Stamatis Katsaounis <katsaouniss@gmail.com>


skatsaounis:
There is an old patch (https://github.com/juju/charm-helpers/commit/05bbaffb0bd2ea4d0a8d782e1a18f683dbe3b3e2) which introduced the section-oslo-notifications with notification-topics included. At the same time, there is no python code which can consume the related charm option and fill the template correctly. This patch does just that.

@javacruft I would love to have a review if you can spare a couple of minutes

PR #409, opened 2019-12-24
Add function to remove deprecated checks

soumplis:
This patch adds a function to allow removing deprecated or obsolete
checks.

### DIFF
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -490,7 +490,10 @@ def remove_deprecated_check(nrpe, deprecated_services):
     """
     Remove checks fro deprecated services in list
 
-    :param NRPE nrpe: NRPE object to remove check from
+    :param nrpe: NRPE object to remove check from
+    :type nrpe: NRPE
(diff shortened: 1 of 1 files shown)


charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -484,3 +484,14 @@ def add_haproxy_checks(nrpe, unit_name):
...
         description='Check HAProxy queue depth {%s}' % unit_name,
         check_cmd='check_haproxy_queue_depth.sh')
+
+
+def remove_deprecated_check(nrpe, deprecated_services):
+    """
+    Remove checks fro deprecated services in list
+
+    :param NRPE nrpe: NRPE object to remove check from
+    """
```
ajkavanagh: Please could we add `:type nrpe: ...` and param and type for `deprecated_services`?  Although it's patchy in the library, it's useful if we can fully spec the docstrings for new functions and when old ones are modified.

soumplis: Thank you for the comment, I think is OK now.

PR #415, opened 2020-01-09
HostInfoContext: Fix retrieval of a hosts primary FQDN

fnordahl:
The implementation of ``socket.getfqdn()`` in the standard Python
library does not exhaust all methods of getting the official name
of a host ref Python issue https://bugs.python.org/issue5004

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -2181,21 +2181,22 @@ class LogrotateContext(OSContextGenerator):
 class HostInfoContext(OSContextGenerator):
     """Context to provide host information."""
 
-    def _get_use_fqdn_hint(self):
-        """Get hint from local unit KV store for whether FQDN should be used
+    def __init__(self, use_fqdn_hint_cb=None):
+        """Initialize HostInfoContext
 
-        Depending on the workload a charm manages, the use of FQDN vs.
-        shortname may be a deploy-time decision, i.e. behavour can not change
-        on post deploy charm upgrade or configuration change.
-
-        The hint is passed on as a flag in the context to allow the decision to
-        be made in the Jinja2 configuration template.
-
-        :returns: True or False
-        :rtype: bool
+        :param use_fqdn_hint_cb: Callback whose return value used to populate
+                                 `use_fqdn_hint`
+        :type use_fqdn_hint_cb: Callable[[], bool]
         """
-        db = kv()
-        return db.get('openstack-hostinfocontext-use-fqdn-hint', False)
+        # Store callback used to get hint for whether FQDN should be used
+
+        # Depending on the workload a charm manages, the use of FQDN vs.
+        # shortname may be a deploy-time decision, i.e. behaviour can not
+        # change on charm upgrade or post-deployment configuration change.
+
+        # The hint is passed on as a flag in the context to allow the decision
+        # to be made in the Jinja2 configuration template.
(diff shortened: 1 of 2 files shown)


fnordahl:
This PR is a prerequisite for https://review.opendev.org/701928 and https://review.opendev.org/701929

charmhelpers/contrib/openstack/context.py
```diff
@@ -2181,9 +2181,64 @@ def __call__(self):
...
+        on post deploy charm upgrade or configuration change.
+
+        The hint is passed on as a flag in the context to allow the decision to
+        be made in the Jinja2 configuration template.
+
+        :returns: True or False
+        :rtype: bool
+        """
+        db = kv()
+        return db.get('openstack-hostinfocontext-use-fqdn-hint', False)
```
thedac: Does this get set in the charm code? I don't see anywhere in charm-helpers where it gets set to True.

charmhelpers/contrib/openstack/context.py
```diff
@@ -2181,9 +2181,64 @@ def __call__(self):
...
+            exc = OSError
+
+        try:
+            addrs = socket.getaddrinfo(
+                name, None, 0, socket.SOCK_DGRAM, 0, socket.AI_CANONNAME)
+        except exc:
+            pass
+        else:
+            for addr in addrs:
+                if addr[3]:
```
rodrigogansobarbieri: This is not inside a try/except block, isn't it safer to check the length before trying to access it?

charmhelpers/contrib/openstack/context.py
rodrigogansobarbieri: It is in the charm itself: https://review.opendev.org/#/c/701928/3/hooks/nova_compute_hooks.py

charmhelpers/contrib/openstack/context.py
fnordahl: Normally I would agree with you, but the library call is well [documented](https://docs.python.org/3/library/socket.html#socket.getaddrinfo) and defined, so I do not see an opportunity for it to deviate from that in this case.

charmhelpers/contrib/openstack/context.py
fnordahl: @thedac as @rodrigogansobarbieri pointed out it is set by the charm, but your question spurred a thought that the context should really not force the charm authors hand as to how it passes on information for the `use_fqdn_hint`.

So based on your feedback I have changed the context to get this from a callback to the charm code instead.

charmhelpers/contrib/openstack/context.py
```diff
@@ -2181,9 +2181,64 @@ def __call__(self):
 class HostInfoContext(OSContextGenerator):
     """Context to provide host information."""
 
+    def _get_use_fqdn_hint(self):
+        """Get hint from local unit KV store for whether FQDN should be used
+
+        Depending on the workload a charm manages, the use of FQDN vs.
+        shortname may be a deploy-time decision, i.e. behavour can not change
```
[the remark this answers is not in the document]
fnordahl: :+1:

charmhelpers/contrib/openstack/context.py
```diff
@@ -2181,9 +2181,64 @@ def __call__(self):
...
+        on post deploy charm upgrade or configuration change.
```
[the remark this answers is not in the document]
fnordahl: :+1:

PR #416, opened 2020-01-10
Fix apt hardening module

tpsilva:
Commit d2ea1b8 replaced the original apt_pkg module with a custom
implementation. This new implementation, however, has a few differences
with how the original module handled some of its calls. It is also
missing the 'config' submodule which is needed by the apt hardening
modules.

This commit implements the 'config' functionality and fixes the
compatibility issues with this new implementation.

Related-Bug: #1857258


tpsilva:
Tested with ceph-mon charm.

PR #422, opened 2020-01-25
DHCPAgentContext for Neutron Plugin API charms

thedac:
Generalize the DHCPAgentContext for Neutron Plugin API charms,
neutron-gateway and neutron-openvswitch.

Primarily to handle the ovs-use-veth setting problem. Get the right
value from config or existing dhcp_agent.ini file.  Existing has
precedence. Attempt to default to "False" without disrupting existing
deployments. Handle existing deployments and upgrades safely. See LP

Partial-Bug: #1831935

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -2267,8 +2267,8 @@ def validate_ovs_use_veth(*args, **kwargs):
     :rtype: Union[(None, None), (string, string)]
     """
     existing_ovs_use_veth = (
-        DHCPAgentContext().get_existing_ovs_use_veth())
-    config_ovs_use_veth = config("ovs-use-veth")
+        DHCPAgentContext.get_existing_ovs_use_veth())
+    config_ovs_use_veth = DHCPAgentContext.parse_ovs_use_veth()
 
     # Check settings are set and not None
     if existing_ovs_use_veth is not None and config_ovs_use_veth is not None:
@@ -2293,12 +2293,14 @@ def validate_ovs_use_veth(*args, **kwargs):
 class DHCPAgentContext(OSContextGenerator):
 
     def __call__(self):
-        """Return the 'default_availability_zone' from the principal that this
+        """Return the DHCPAGentContext.
+
+        Return all DHCP Agent INI related configuration.
         ovs unit is attached to (as a subordinate) and the 'dns_domain' from
         the neutron-plugin-api relations (if one is set).
 
-        :returns: {} if no relation set, or
-            {'availability_zone': availability_zone from principal relation}
+        :returns: Dictionary context
+        :rtype: Dict
         """
 
         ctxt = {}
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/context.py
```diff
@@ -2244,3 +2248,119 @@ def __call__(self):
...
+class DHCPAgentContext(OSContextGenerator):
+
+    def __call__(self):
+        """Return the 'default_availability_zone' from the principal that this
+        ovs unit is attached to (as a subordinate) and the 'dns_domain' from
+        the neutron-plugin-api relations (if one is set).
+
+        :returns: {} if no relation set, or
+            {'availability_zone': availability_zone from principal relation}
+        """
```
ajkavanagh: :rtype: Dict[str,?]  ?

charmhelpers/contrib/openstack/context.py
```diff
@@ -2244,3 +2248,119 @@ def __call__(self):
...
+            ctxt['enable_metadata_network'] = True
+            ctxt['enable_isolated_metadata'] = True
+
+        return ctxt
+
+    def get_existing_ovs_use_veth(self):
+        """Return existing ovs_use_veth setting from dhcp_agent.ini.
+
+        :returns: Value of ovs_use_veth setting: None or Bool
+        :rtype: Union[None, Bool]
```
ajkavanagh: As Frode has discovered the excellent `Optional[Bool]` which is the same as `Union[None, Bool]` (which I also have previously been using), we should probably try to switch to the former?

charmhelpers/contrib/openstack/context.py
```diff
@@ -2244,3 +2248,119 @@ def __call__(self):
...
+    def get_ovs_use_veth(self):
+        """Return correct ovs_use_veth setting.
+
+        Get the right value from config or existing dhcp_agent.ini file.
+        Existing has precedence. Attempt to default to "False" without
+        disrupting existing deployments. Handle existing deployments and
+        upgrades safely. See LP Bug#1831935
+
+        :returns: Value to use for ovs_use_veth setting
+        :rtype: string
```
ajkavanagh: :rtype: str

charmhelpers/contrib/openstack/context.py
```diff
@@ -2244,3 +2248,119 @@ def __call__(self):
...
+        existing_ovs_use_veth = None
+        # If there is a dhcp_agent.ini file read the current setting
+        if os.path.isfile(dhcp_agent_ini):
+            # config_ini does the right thing and returns None if the setting is
+            # commented.
+            existing_ovs_use_veth = (
+                config_ini(dhcp_agent_ini)["DEFAULT"].get("ovs_use_veth"))
+        return existing_ovs_use_veth
+
+    def get_ovs_use_veth(self):
```
ajkavanagh: Shouldn't this also be a `@staticmethod`?  You might need to adjust the call to `get_existing_ovs_use_veth()` below.

charmhelpers/contrib/openstack/context.py
```diff
@@ -2244,3 +2248,119 @@ def __call__(self):
...
+        # Override user supplied config for these plugins as these settings are
+        # mandatory
+        if config('plugin') in ['nvp', 'nsx', 'n1kv']:
+            ctxt['enable_metadata_network'] = True
+            ctxt['enable_isolated_metadata'] = True
+
+        return ctxt
+
+    def get_existing_ovs_use_veth(self):
+        """Return existing ovs_use_veth setting from dhcp_agent.ini.
```
ajkavanagh: Shouldn't this be a `@staticmethod` ?  i.e. no need for self in this.

charmhelpers/contrib/openstack/context.py
```diff
@@ -2244,3 +2248,119 @@ def __call__(self):
...
+        _existing = self.get_existing_ovs_use_veth()
+        if _existing is not None:
+            return str(_existing)
+
+        _config = config('ovs-use-veth')
+        if _config is None:
+            # New better default
+            return "False"
+        else:
+            return str(_config)
```
ajkavanagh: I think this needs more validation of the input.  It's a String, but really it's a trinary boolean value of True, False, None.

I think (up for discussion) that this function should return Boolean (True/False) and do the work in the function to extract the boolean from the str.  i.e. parse it to a type here and then that type is the version to use.  My rationale is that if a str is allowed to escape this function without validation, ovs-use-veth could be set to "None" as a string, not realising that it should really be a boolean.

charmhelpers/contrib/openstack/context.py
```diff
@@ -2244,3 +2248,119 @@ def __call__(self):
...
+            log(msg, ERROR)
+            return (
+                "blocked",
+                "Mismatched existing and configured ovs-use-veth. See log.")
+
+    # Everything is OK
+    return None, None
+
+
+class DHCPAgentContext(OSContextGenerator):
```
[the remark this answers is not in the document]
thedac: I have taken pains to make sure we do not lose anything. ZoneContext was specific to neutron.conf. The DHCPAgentContext only affects items in dhcp_agent.ini

charmhelpers/contrib/openstack/context.py
thedac: This was copy pasta. Fixing.

charmhelpers/contrib/openstack/context.py
thedac: Love it. Updating.

charmhelpers/contrib/openstack/context.py
thedac: Agreed. Adding a parse_ovs_use_veth method to convert string to bool and then this method will return bool.

charmhelpers/contrib/openstack/context.py
thedac: Makes sense. Changing.

charmhelpers/contrib/openstack/context.py
```diff
@@ -2244,3 +2248,151 @@ def __call__(self):
...
+        # An unset parameter returns None. Just in case we will also check for
+        # an empty string: "". Ironically, (the problem we are trying to avoid)
+        # "False" returns True and "" returns False.
+        if _config is None or not _config:
+            # Return None
+            return
+        # bool_from_string handles many variations of true and false strings
+        # as well as upper and lowercases including:
+        # ['y', 'yes', 'true', 't', 'on', 'n', 'no', 'false', 'f', 'off']
+        return bool_from_string(_config)
```
ajkavanagh: This is lovely!

PR #423, opened 2020-01-27
Support for python3.8 and python3.4

addyess:
* Begin testing with python 3.8 and focal
* resolves issue where `os` modules no longer has `errno` attribute
* adds feature allowing py27 and py34 environments to use `ufw.status()`
* resolves issue where `platform.linux_dsitribution()` was removed in py38
* resolves issue with policyd in py34 
* updates project to use pyyaml 5.3 rather than 3.10
  * python3.8 wouldn't compile with an older version of PyYAML==3.10
  * python3.4 was dropped from 5.3, so using 5.2 for py34
  * this change modifies the style of text written out by `yaml.dump()`
* updates project to use Jinja2==2.11
  * python3.4 was dropped from 2.11, so using 2.10 for py34
* uses `yaml.safe_load()` rather than `yaml.load()`
  * this removed warnings during unit testing


addyess:
Could help with issue: https://github.com/juju/charm-helpers/issues/293

charmhelpers/osplatform.py
```diff
@@ -9,9 +10,13 @@ def get_platform():
...
     # linux_distribution is deprecated and will be removed in Python 3.7
-    # Warings *not* disabled, as we certainly need to fix this.
-    tuple_platform = platform.linux_distribution()
-    current_platform = tuple_platform[0]
+    # Warnings *not* disabled, as we certainly need to fix this.
+    if hasattr(platform, 'linux_distribution'):
+        tuple_platform = platform.linux_distribution()
+        current_platform = tuple_platform[0]
+    else:
+        current_platform = _get_platform_from_fs()
```
coreycb: I think you can get these details from:
import distro
distro.linux_distribution()

tests/core/test_hookenv.py
```diff
@@ -1288,7 +1288,7 @@ def test_relation_set_file(self, check_call, check_output, remove):
         self.assertEqual("--file", command[1])
         temp_file = command[2]
         with open(temp_file, "r") as f:
-            self.assertEqual("{foo: bar}", f.read().strip())
+            self.assertEqual("foo: bar", f.read().strip())
```
coreycb: Can you add a pointer to why this has changed and is it backward compatibile with earlier py3's and py2.7?

coreycb:
Nevermind my commit message comment. You have several commits here with what looks like enough detail.

tests/core/test_hookenv.py
coreycb: This is due to pyyaml 5.3 rather than 3.10  and since test-requirements.txt is pinned to 5.3 these changes in tests should be good.

.travis.yml
```diff
@@ -4,6 +4,7 @@ language: python
 python:
   - "2.7"
   - "3.5"
+  - "3.8"
```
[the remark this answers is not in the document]
addyess: i believe the `.travis.yml` instructs travis to run the unit tests in a 2.7 and 3.5 environment.  I wanted to make sure we also ran under 3.8.  I don't know why 3.6 and 3.7 were ignored.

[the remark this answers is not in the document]
coreycb: Yes looks like this controls the integrated github CI. Seems fine to me to at least run 3.8.

charmhelpers/osplatform.py
addyess: question for the larger community... i didn't see a "requirements.txt" but i did set a setup.py.  I'd be willing to add the 'distro' package as a dependent library but didn't know the communities reaction to addition of another library and if the downstream charms would somehow know to add this dep.

tests/core/test_hookenv.py
addyess: yes, this is because testing now takes place using pyyaml 5.3  Python 3.8 was thoroughly unhappy with pyyaml 3.10.  You might see on travis that there were errors building that version of pyyaml.

charmhelpers/osplatform.py
coreycb: That is a good question.

tests/contrib/openstack/test_audits.py
```diff
@@ -61,7 +61,7 @@ def test(options):
         try:
             # Again!
             @audits.audit(should_run)  # noqa
-            def test(options):
+            def test(options):         # noqa
```
addyess: comment why this is here...

.travis.yml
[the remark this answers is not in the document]
ryan-beisner: I would suggest that we align this to the Python versions that ship with each Ubuntu LTS.

.travis.yml
```diff
@@ -3,7 +3,9 @@ sudo: required
 language: python
 python:
   - "2.7"
-  - "3.5"
+  - "3.4"  # trusty
+  - "3.5"  # xenial
```
coreycb: I think we want 3.6

tests/contrib/network/test_ufw.py
```diff
@@ -531,12 +531,8 @@ def test_reload_fail(self, modprobe, check_output, log):
         log.assert_any_call("ufw couldn't be reloaded", level='WARN')
 
     def test_status(self):
-        if six.PY2:
-            return
-        with mock.patch('subprocess.run') as run:
-            cp = mock.MagicMock()
-            cp.stdout = UFW_STATUS_NUMBERED
-            run.return_value = cp
+        with mock.patch('subprocess.check_output') as check_output:
```
addyess: @fnordahl @ajkavanagh -- can you confirm this a better fix?

charmhelpers/contrib/network/ufw.py
```diff
@@ -366,12 +366,10 @@ def status():
         (1, {'to': '', 'action':, 'from':, '', ipv6: True, 'comment': ''})
     :rtype: Iterator[Tuple[int, Dict[str, Union[bool, str]]]]
     """
-    if six.PY2:
-        raise RuntimeError('Call to function not supported on Python2')
-    cp = subprocess.run(('ufw', 'status', 'numbered',),
-                        stdout=subprocess.PIPE, stderr=subprocess.STDOUT,
-                        check=True, universal_newlines=True)
-    for line in cp.stdout.splitlines():
+    cp = subprocess.check_output(('ufw', 'status', 'numbered',),
```
ajkavanagh: Why are we losing PY2 guards?

charmhelpers/osplatform.py
ajkavanagh: Is `distro` on PyPi and pure python?  If not, then it wouldn't work in a venv, which would break reactive charms (and maybe the new operator framework).

tests/contrib/network/test_ufw.py
ajkavanagh: More concerned with losing PY2 support from the library.

charmhelpers/osplatform.py
coreycb: After more discussion I think what you have here is good, as opposed to requiring a new dependency to be pulled into classic charms. I tried your code out successfully and it looks good. Do you know if this works with CentOS?

charmhelpers/contrib/network/ufw.py
addyess: this was trying to protect the use of `subprocess.run`.  But honestly, subprocess.run was introduced in python35.  i'm trying make charm-helpers continue to work in py34.  Because six.PY2 is false in py34, this began to fail.  

I felt like using check_output is more broadly available across py2 and py34 as well as py38

tests/contrib/network/test_ufw.py
addyess: this actually includes py2 support rather than losing it.

charmhelpers/osplatform.py
addyess: `distro` is an extra package.  It would force classic charms to include it.  That's why this implementation was chosen.

```
(venv)root@first:~/juju/charm-helpers# python
Python 3.4.3 (default, Nov 12 2018, 22:25:49) 
[GCC 4.8.4] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import distro
ImportError: No module named 'distro'
```

addyess: i'll try to run a lxd container in centos w/ py38 and ensure this works

addyess:
Does anyone know if the VERSION file needs to be updated?

charmhelpers/osplatform.py
addyess: OK... i took a simpler approach.  the answer is yes, it works on [centos](https://paste.ubuntu.com/p/2KkVb85T8v/)

charmhelpers/contrib/openstack/policyd.py
```diff
@@ -531,7 +530,7 @@ def clean_policyd_dir_for(service, keep_paths=None, user=None, group=None):
     hookenv.log("Cleaning path: {}".format(path), level=hookenv.DEBUG)
     if not os.path.exists(path):
         ch_host.mkdir(path, owner=_user, group=_group, perms=0o775)
-    _scanner = os.scandir if sys.version_info > (3, 4) else _py2_scandir
+    _scanner = os.scandir if hasattr(os, 'scandir') else _fallback_scandir
```
timClicks: I like the renaming, FWII

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -1042,7 +1042,7 @@ def filesystem_mounted(fs):
 def make_filesystem(blk_device, fstype='ext4', timeout=10):
     """Make a new filesystem on the specified block device."""
     count = 0
-    e_noent = os.errno.ENOENT
+    e_noent = errno.ENOENT
```
timClicks: Curious, why is `errno` in global scope?

charmhelpers/core/host_factory/ubuntu.py
```diff
@@ -25,6 +25,7 @@
     'cosmic',
     'disco',
     'eoan',
+    'focal'
 )
```
timClicks: We should consider changing `UBUNTU_RELEASES` to a set

charmhelpers/contrib/storage/linux/ceph.py
addyess: i guess because there's a couple of places in this file where the module is used?

charmhelpers/core/host_factory/ubuntu.py
addyess: i think the intent is that the order matters... but I see your point.  Maybe an 'OrderedSet' if there was such a thing...

tests/contrib/openstack/test_policyd.py
```diff
@@ -236,11 +237,11 @@ def test_clean_policyd_dir_for(self,
                                    mock_shutil_rmtree,
                                    mock_os_remove,
                                    mock_log):
-        if six.PY3:
+        if hasattr(os, 'scandir'):
             mock_scan_dir_parts = (mock.patch, ["os.scandir"])
         else:
             mock_scan_dir_parts = (mock.patch.object,
-                                   [policyd, "_py2_scandir"])
+                                   [policyd, "_fallback_scandir"])
```
timClicks: This could get brittle if `_fallback_scandir` is ever changed in the future. Perhaps add a comment in that function that its relied on by the mocks?

tests/core/test_hugepage.py
```diff
@@ -37,8 +37,9 @@ def test_hugepage_support(self):
         self.fstab.Fstab().get_entry_by_attr.return_value = 'old fstab entry'
         self.fstab.Fstab().Entry.return_value = 'new fstab entry'
         hugepage.hugepage_support('nova')
-        sysctl_expect = ("{vm.hugetlb_shm_group: '1010', "
-                         "vm.max_map_count: 65536, vm.nr_hugepages: 256}\n")
+        sysctl_expect = ("vm.hugetlb_shm_group: '1010'\n"
+                         "vm.max_map_count: 65536\n"
+                         "vm.nr_hugepages: 256\n")
```
timClicks: Maybe use a multi-line string literal? It looks like a tuple for someone skim-reading.

tests/contrib/openstack/test_policyd.py
addyess: are you saying comment in `./contrib/openstack/policyd.py`  around `_fallback_scandir`?  To let a future dev know that this test is mocking it out?

techalchemy:
thanks for your hard work on this, I am not a committer but would very much like to see this merged.

charmhelpers/osplatform.py
coreycb: thanks for checking

addyess:
functional smoke for [nova-compute](https://github.com/openstack/charm-nova-compute) still works with these changes

tests/contrib/ansible/test_ansible.py
```diff
@@ -165,7 +165,7 @@ def test_writes_vars_file(self):
         self.assertEqual(stats.st_mode & stat.S_IRWXG, 0)
         self.assertEqual(stats.st_mode & stat.S_IRWXO, 0)
         with open(self.vars_path, 'r') as vars_file:
-            result = yaml.load(vars_file.read())
+            result = yaml.safe_load(vars_file.read())
```
ChrisMacNaughton: Thanks for this :)

tests/contrib/openstack/test_audits.py
ChrisMacNaughton: :-D

ChrisMacNaughton:
After a review from @ryan-beisner, this should be OK to land (assuming his changes are resolved)!

charmhelpers/core/host_factory/ubuntu.py
ajkavanagh: The order is very important; this particular #def is used to determine the order of releases and determine that "focal" comes after zesty, for example.

tests/contrib/network/test_ufw.py
ajkavanagh: That's good then!

ajkavanagh:
@addyess thanks for the work on this!

PR #425, opened 2020-01-31
replace ansible extra-vars parsing with json

matuskosut:
Ansible-playbook supports json string in --extra-vars, which makes more sense to use than converting extra variables into key/value strings and gives possibility to pass dictionaries as values to ansible.

- does not break compatibility with current use cases (variable=value dictionaries used now will be passed as same variables)
- additional is the possibility to use nested dictionary, so value can be a dictionary
- json.dumps also handles wrapping string path such as '@myvars.yml' or '@randomvars.json' which is also supported by ansible

https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#passing-variables-on-the-command-line


szeestraten-ntnu:
Looks good. Recommend adding docstring for `apply_playbook` to clarify what input it takes.

charmhelpers/contrib/ansible/__init__.py
```diff
@@ -139,6 +139,42 @@ def install_ansible_support(from_ppa=True, ppa_location='ppa:rquillo/ansible'):
...
+    """Run a playbook.
+
+    This helper runs a playbook with juju state variables as context,
+    therefore variables set in application config can be used directly.
+    List of tags (--tags) and dictionary with extra_vars (--extra-vars)
+    can be passed as additional parameters.
+
+    Read more about playbook `_variables`_ online.
+
+    .. _variables: https://docs.ansible.com/ansible/2.5/user_guide/playbooks_variables.html
```
szeestraten-ntnu: Should be latest upstream docs: https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html

PR #434, opened 2020-02-20
Allow Jinja2 lib <=2.11

auria:


requirements.txt
```diff
@@ -9,9 +9,8 @@ PyYAML;     python_version == '2.7' or python_version >= '3.5'  # all else
 # https://jinja.palletsprojects.com/en/2.11.x/changelog/
 # Jinja2==2.10 is last supported for py34
 # Jinja2==2.11 is last supported for py27 & py35
-Jinja2==2.10;python_version >= '3.0' and python_version <= '3.4' # py3 trusty
-Jinja2==2.11;python_version == '2.7' or python_version == '3.5'  # py27, py35
-Jinja2;      python_version >= '3.6'                             # py36 and on
```
johnsca: The comment in layer-basic's `wheelhouse.txt` indicates that it's attempting to support trusty, but the current syntax in charm-helpers is more sophisticated and seems to support newer versions as well.  I think it would probably be better to copy the more sophisticated syntax over to layer-basic, rather than dumbing down charm-helpers?

fnordahl: @johnsca it would be awesome if that worked, but I think you will find that with this approach a set of wheels will be built by charm-tools and layer-basic will expect a different set of wheels on the host it deploys on.  If you have a solution for charm-tools that would allow including all of the permutations of the requested wheels I'm all ears :)

fnordahl:
Just wanted to add a note about a proposal from the original author to revert to the previous requirements here: #435

johnsca:
Closing in favor of #435 which was merged.

Issue #433, opened 2020-02-20
Jinja2==2.11 is not installed with layer-basic


xavpaice:
Since commit juju/charm-helpers@2e65842 we need 2.11 in order to install/upgrade charms, but the wheelhouse.txt in layer-basic specifies Jinja2<=2.10.1.

Not sure if this is simply that wheelhouse.txt needs an update, or if charmhelpers needs a fix, will report in both places.

fnordahl:
This is most likely due to a mismatch between the build environment and the deploy environment.

That mismatch will always be there so I think the changes to the setup and requires files in the referenced commit, while with good intention, does not actually work in practice.

Perhaps @addyess would want to take a look and perhaps revert that part of the change?

auria:
I have tested with the cs:ubuntu charm, which uses layer-basic to install default python libs.

Indeed, this issue happens because cs:ubuntu had "charm build" run from a machine with a py3 version >=3.4 (not a trusty machine).

charmhelpers is forcing (ie. in bionic) Jinja2==2.11 but lower versions should be allowed. I've pushed a MR and testing passed.

See https://github.com/juju/charm-helpers/pull/434

auria:
BTW, the issue can be reproduces by running:

```
for SERIES in trusty xenial bionic;do
juju deploy --series $SERIES cs:ubuntu ubuntu-${SERIES}
done
```
By running the above, the `install` hook will fail in trusty and xenial.

auria:
The issue has also been filed as affecting charms such as https://bugs.launchpad.net/telegraf-charm/+bug/1863962

addyess:
i might be able to upload a *new* change, but revert is no good.  

Previously the `setup.py` listed just "Jinja2" with no pinning at all.  That just means, it would follow the tip of that library wherever it might go.  

Perhaps the easiest solution at this point is to pin to `Jinja2==2.10` which currently holds the broadest support from py27 -> py38.

```
> python3 -m virtualenv -p python3 venv
Already using interpreter /usr/bin/python3
Using base prefix '/usr'
/usr/lib/python3/dist-packages/virtualenv.py:1090: DeprecationWarning: the imp module is New python executable in /home/ubuntu/venv/bin/python3
Also creating executable in /home/ubuntu/venv/bin/python
Installing setuptools, pkg_resources, pip, wheel...done.
> source venv/bin/activate
(venv) > pip install Jinja2==2.10
Collecting Jinja2==2.10
[...]

PR #435, opened 2020-02-20
Unpinning Jinja2 and PyYAML

addyess:
In an attempt to address https://github.com/juju-solutions/layer-basic/issues/155 and https://github.com/juju/charm-helpers/issues/433, remove specific version pinning of Jinja2 and PyYAML


addyess:
@fnordahl, do you see why pinning is a requirement?  Maybe I should use specific pinning tricks in the test_requirements.txt?

addyess:
@xavpaice Maybe this is what you're looking for?

addyess:
wrt `libraries shouldn't pin versions...`

someone should tell that to PyYAML and Jinja2.

PR #437, opened 2020-03-02
Ensure that apache can handle certificates from Vault

ChrisMacNaughton:
Closes-Bug: #1846189


ajkavanagh:
@ChrisMacNaughton is this still current?  If so, it's got a conflict in `cert_utils.py`.  Thanks!

Issue #408, opened 2019-12-24
NRPE: add check: rebuilding command after path validation doesn't requote properly


exsdev0:
Given an NRPE check command that includes quotes around its arguments (e.g. `check_timed_logs.pl -timepattern '%b %e %H:%M:%S'`), the corresponding generated NRPE check config file (e.g. `/etc/nagios/nrpe.d/check_timed_logs.cfg`) does not preserve the original check command argument quoting (i.e. it will have `command[check_timed_logs]=/usr/local/lib/nagios/plugins/check_timed_logs.pl -timepattern %b %e %H:%M:%S`).

In order to determine if the check command exists on disk, the original command is first run through `shlex` (which creates parts and removes quoting). After the existence of the check script is validated, there is an attempt to rebuild the original command from the parts -- but this rebuild doesn't respect the original quoting.

I believe the offending line is as follows: https://github.com/juju/charm-helpers/blob/master/charmhelpers/contrib/charmsupport/nrpe.py#L176

Here's a snippet of code that exposes the issue:

```python
#!/usr/bin/python3

import os
import shlex
import yaml

yaml_content = """
--- 
nagios-replication-check-params: 
  default: "-timepattern '%b %e %H:%M:%S'"
"""

parsed = yaml.load(yaml_content, Loader=yaml.FullLoader)


check_cmd = 'check_timed_logs.pl {}'.format(parsed["nagios-replication-check-params"]["default"])
print("check_cmd string: ", check_cmd)

parts = shlex.split(check_cmd)
print("shlex parts: ", parts)

command = os.path.join("/usr/local/lib/nagios/plugins", parts[0])
if len(parts) > 1:
    command += " " + " ".join(parts[1:])
print("rebuilt command: ", command)

```

The output is as follows:
```
check_cmd string:  check_timed_logs.pl -timepattern '%b %e %H:%M:%S'
shlex parts:  ['check_timed_logs.pl', '-timepattern', '%b %e %H:%M:%S']
rebuilt command:  /usr/local/lib/nagios/plugins/check_timed_logs.pl -timepattern %b %e %H:%M:%S
```

Note that in the rebuilt command, the quoting around the time pattern (%b %e %H:%M:%S) is gone. This needs to be quoted properly or it will be treated as three separate arguments (space separated).

PR #438, opened 2020-03-04
_locate_cmd: use shlex.join to retain word boundaries in nrpe check commands

jetpackdanger:
Addresses https://github.com/juju/charm-helpers/issues/408 and https://bugs.launchpad.net/charm-helpers/+bug/1706265

### DIFF
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -122,6 +122,34 @@ from charmhelpers.core import host
 #    ln -s hooks.py local-monitors-relation-changed
 
 
+# If shlex.join is not available (new as of Python 3.8) then let's emulate it.
+if 'join' in dir(shlex):
+    shlex_join = shlex.join
+else:
+    if 'ASCII' in dir(re):
+        # Python 3 defaults to Unicode
+        _find_unsafe = re.compile(r'[^\w@%+=:,./-]', re.ASCII).search
+    else:
+        # Python 2 defaults to ASCII
+        _find_unsafe = re.compile(r'[^\w@%+=:,./-]').search
+
+    # This is just shlex.join and friends from Python 3.8 with no substantive changes.
+    def shlex_quote(s):
+        """Return a shell-escaped version of the string *s*."""
+        if not s:
+            return "''"
+        if _find_unsafe(s) is None:
+            return s
+
+        # use single quotes, and put single quotes into double quotes
+        # the string $'b is then quoted as '$'"'"'b'
+        return "'" + s.replace("'", "'\"'\"'") + "'"
+
+    def shlex_join(split_command):
+        """Return a shell-escaped string from *split_command*."""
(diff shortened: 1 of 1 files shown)


jetpackdanger:
Ugh, looks like shlex.join is too new.

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -122,6 +122,34 @@
...
+            return s
+
+        # use single quotes, and put single quotes into double quotes
+        # the string $'b is then quoted as '$'"'"'b'
+        return "'" + s.replace("'", "'\"'\"'") + "'"
+
+    def shlex_join(split_command):
+        """Return a shell-escaped string from *split_command*."""
+        return ' '.join(shlex_quote(arg) for arg in split_command)
+
```
ajkavanagh: It feels like there should be a unit test for this to verify that it is doing what the Py3.8 version does?

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -122,6 +122,34 @@
 #    ln -s hooks.py local-monitors-relation-changed
 
 
+# If shlex.join is not available (new as of Python 3.8) then let's emulate it.
+if 'join' in dir(shlex):
```
ajkavanagh: Maybe use:

```python
try:
    shlex_join = shlex.join
except AttributeError:  # it might be something else?
   # rest of code, etc.
```

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -122,6 +122,34 @@
...
+    shlex_join = shlex.join
+else:
+    if 'ASCII' in dir(re):
+        # Python 3 defaults to Unicode
+        _find_unsafe = re.compile(r'[^\w@%+=:,./-]', re.ASCII).search
+    else:
+        # Python 2 defaults to ASCII
```
ajkavanagh: We've got `six` available so that a test of `if six.PY3: ...` can be done.

jetpackdanger:
Thank you for all the excellent suggestions!

However, I've just realized there's a fatal flaw with this approach, which is that all of the existing `add_check` calls will  get an extra level of escaping that will probably break them. For example:

    >>> command = '/usr/lib/nagios/plugins/check_http -k "Accept:\\ application/json" -I 127.0.0.1 -p 9000 -u /api -s cluster_id'
    >>> print(' '.join(shlex.split(command)))  # currently
    /usr/lib/nagios/plugins/check_http -k Accept:\ application/json -I 127.0.0.1 -p 9000 -u /api -s cluster_id
    >>> print(shlex.join(shlex.split(command)))  # with my change
    /usr/lib/nagios/plugins/check_http -k 'Accept:\ application/json' -I 127.0.0.1 -p 9000 -u /api -s cluster_id
    >>> _

I haven't been able to think of a clever idea to detect such cases yet, so I'll  close this PR and think about it some more.

PR #442, opened 2020-03-27
add helpers for log with levels

guoqiao:
currently to write an error log with charmhelpers.hookenv:

    hookenv.log('this is an error!', level=hookenv.ERROR)

The `level` arg is a bit overwhelmed for user.
Thus, it ends up ignored normally, and defaults to INFO even it's an ERROR in a lot of charms.

Add simple helpers, so we can write a log with correct level without thinking.

Signed-off-by: Joe Guo <joe.guo@canonical.com>

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -121,6 +121,30 @@ def log(message, level=None):
             raise
 
 
+def log_trace(message):
+    return log(message, level=TRACE)
+
+
+def log_debug(message):
+    return log(message, level=DEBUG)
+
+
+def log_info(message):
+    return log(message, level=INFO)
+
+
+def log_warning(message):
+    return log(message, level=WARNING)
+
+
+def log_error(message):
+    return log(message, level=ERROR)
+
+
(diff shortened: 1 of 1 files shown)


charmhelpers/core/hookenv.py
```diff
@@ -121,6 +121,30 @@ def log(message, level=None):
...
+def log_warning(message):
+    return log(message, level=WARNING)
+
+
+def log_error(message):
+    return log(message, level=ERROR)
+
+
+def log_critical(message):
+    return log(message, level=CRITICAL)
```
ajkavanagh: These need docstrings to indicate what they do and the types of the params.  Also, a related test is needed to verify that each one does actually do what it says.

AurelienLourot:
Thanks for the contribution, but I also don't think this is adding much value. This is purely a matter of personal taste, which syntax one prefers. Personally I'm neutral on which one is best, so I'd just stick to the solution with the least amount of code to maintain. Closing this for now, but feel free to re-open, thanks!

PR #443, opened 2020-03-30
Move OVS/OVN helpers from charm-layer-ovn

fnordahl:


charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -303,3 +306,154 @@ def port_to_br(port):
...
+        3: https://github.com/openvswitch/ovs/blob/
+               20dac08fdcce4b7fda1d07add3b346aa9751cfbc/
+                   lib/db-ctl-base.c#L189-L215
+    :type data: Optional[Dict[str,Union[str,Dict[str,str]]]]
+    :param table: Name of table to operate on
+    :type table: str
+    :param entity: Name of entity to operate on
+    :type entity: str
+    :returns: '--' separated ``ovs-vsctl set`` commands
+    :rtype: Iterator[str]
```
ajkavanagh: It doesn't appear to return a `str` .. should it be a `Tuple[str, str, str, str, str]` ?

charmhelpers/contrib/network/ovs/ovn.py
```diff
@@ -0,0 +1,148 @@
...
+    Open vSwitch, from 20.03 and onwards they are kept in a separate directory.
+
+    :returns: path to OVN rundir
+    :rtype: str
+    """
+    OVN_RUNDIR = '/var/run/ovn'
+    OVS_RUNDIR = '/var/run/openvswitch'
+    for rundir in OVN_RUNDIR, OVS_RUNDIR:
+        if os.path.exists(rundir):
+            return (rundir)
```
ajkavanagh: Not sure why `(rundir)` is in parenthesis?

charmhelpers/contrib/network/ovs/ovn.py
```diff
@@ -0,0 +1,148 @@
...
+    # Fall back to new default
+    return OVN_RUNDIR
```
ajkavanagh: I missing something here?  `OVN_RUNDIR` doesn't exist at this point so why return anything?  Why not raise an exception?

charmhelpers/contrib/network/ovs/ovn.py
```diff
@@ -0,0 +1,148 @@
...
+    :returns: path to OVN sysconfdir
+    :rtype: str
+    """
+    OVN_SYSCONFDIR = '/etc/ovn'
+    OVS_SYSCONFDIR = '/etc/openvswitch'
+    for sysconfdir in OVN_SYSCONFDIR, OVS_SYSCONFDIR:
+        if os.path.exists(sysconfdir):
+            return (sysconfdir)
+    # Fall back to new default
+    return OVN_SYSCONFDIR
```
ajkavanagh: Same comments as above.  Why the parenthesis?  Also, the default directory doesn't exist at the end of the function?

charmhelpers/contrib/network/ovs/ovn.py
```diff
@@ -0,0 +1,148 @@
...
+            k = k.replace(' ', '_')
+            if v:
+                if k in ('cluster_id', 'server_id',):
+                    v = v.replace('(', '')
+                    v = v.replace(')', '')
+                    status[k] = tuple(v.split())
+                else:
+                    status[k] = v.lstrip()
+            else:
+                status[k] = []
```
ajkavanagh: The middle `status[k]` assigns a`str` whereas the other two assign a `List[str]`.  Is that a bug, or by design?

charmhelpers/contrib/network/ovs/ovsdb.py
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(self, tool, table):
+        """SimpleOVSDB constructor
+
+        :param tool: Which tool with database commands to operate on.
+                     Usually one of `ovs-vsctl`, `ovn-nbctl`, `ovn-sbctl`
+        :type tool: str
+        :param table: Which table to operate on
+        :type table: str
+        """
+        self.tool = tool
```
ajkavanagh: It might be useful to validate the `tool` really is one of the tools that can be used.  i.e.

```python
if tool not in ['ovs-vsctl', 'ovn-nbctl', 'ovn-sbctl']:
    raise RuntimeError("param tool passed to SimpleOVSCB() not one of ...")
```

charmhelpers/contrib/network/ovs/ovsdb.py
```diff
@@ -0,0 +1,98 @@
...
+        self.tbl = table
+
+    def _find_tbl(self, condition=None):
+        """Run and parse output of OVSDB `find` command.
+
+        :param condition: An optional RFC 7047 5.1 match condition
+        :type condition: Optional[str]
+        :returns: Dictionary with data
+        :rtype: Dict[str, any]
```
ajkavanagh: Function appears to yield an `Iterator[Dict[str, ANY]}` ?

charmhelpers/contrib/network/ovs/utils.py
```diff
@@ -0,0 +1,27 @@
...
+    """Run a process, check result, capture decoded output from STDERR/STDOUT.
+
+    :param args: Command and arguments to run
+    :type args: Tuple[str, ...]
+    :returns: Information about the completed process
+    :rtype: str
+    :raises subprocess.CalledProcessError
+    """
+    return subprocess.check_output(args, stderr=subprocess.STDOUT,
+                                   universal_newlines=True)
```
ajkavanagh: Bearing in mine the recent discussion regarding piping STDERR to STDOUT and the brittleness that this provides, should it be captured to STDOUT?  i.e. is it useful to capture the STDERR as part of the output of the function?

tests/contrib/network/ovs/test_ovsdb.py
```diff
@@ -0,0 +1,96 @@
...
+{"data":[[["uuid","1e21ba48-61ff-4b32-b35e-cb80411da351"],["set",[]],["set",[]],"0000a0369fdd3890","","<unknown>",["map",[["charm-ovn-chassis","managed"],["other","value"]]],["set",[]],["set",[]],["map",[]],["set",[]],false,["set",[]],"br-test",["set",[]],["map",[]],["set",[["uuid","617f9359-77e2-41be-8af6-4c44e7a6bcc3"],["uuid","da840476-8809-4107-8733-591f4696f056"]]],["set",[]],false,["map",[]],["set",[]],["map",[]],false],[["uuid","bb685b0f-a383-40a1-b7a5-b5c2066bfa42"],["set",[]],["set",[]],"00000e5b68bba140","","<unknown>",["map",[]],"secure",["set",[]],["map",[]],["set",[]],false,["set",[]],"br-int",["set",[]],["map",[["disable-in-band","true"]]],["set",[["uuid","07f4c231-9fd2-49b0-a558-5b69d657fdb0"],["uuid","8bbd2441-866f-4317-a284-09491702776c"],["uuid","d9e9c081-6482-4006-b7d6-239182b56c2e"]]],["set",[]],false,["map",[]],["set",[]],["map",[]],false]],"headings":["_uuid","auto_attach","controller","datapath_id","datapath_type","datapath_version","external_ids","fail_mode","flood_vlans","flow_tables","ipfix","mcast_snooping_enable","mirrors","name","netflow","other_config","ports","protocols","rstp_enable","rstp_status","sflow","status","stp_enable"]}
```
ajkavanagh: This is one huge line?

charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -303,3 +306,158 @@ def port_to_br(port):
...
+        if isinstance(v, dict):
+            entries = {
+                '{}:{}'.format(k, dk): dv for (dk, dv) in v.items()}
+        else:
+            entries = {k: v}
+        for (colk, colv) in entries.items():
+            yield ('--', 'set', table, entity, '{}={}'.format(colk, colv))
+
+
+def add_br(bridge, brdata=None, exclusive=False):
```
AurelienLourot: FYI: I'm preparing a pull-request on top of yours in which I add here a few more args that I need, and especially in which I mark add_bridge() as deprecated in favor of add_br() and add_bridge_port() in favor of add_port().

Do you think this should be done in your pull-request already? If not, then this file is going to be confusing right after your merge (it'll be hard for the consumer to chose between these similar functions)

charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -303,3 +306,158 @@ def port_to_br(port):
...
+        have a more pleasant data structure.
+
+        0: http://www.openvswitch.org/ovs-vswitchd.conf.db.5.pdf
+        1: https://tools.ietf.org/html/rfc7047
+        2: http://docs.openvswitch.org/en/latest/topics/integration/
+        3: https://github.com/openvswitch/ovs/blob/
+               20dac08fdcce4b7fda1d07add3b346aa9751cfbc/
+                   lib/db-ctl-base.c#L189-L215
+    :type brdata: Optional[Dict[str,Union[str,Dict[str,str]]]]
+    :param exclusive: If True, raise exception if port exists
```
AurelienLourot: Copy-paste mistake: should be "raise exception if **bridge** exists"

charmhelpers/contrib/network/ovs/ovn.py
fnordahl: My pre-Python habits present themselves sometimes apparently, fixed.

charmhelpers/contrib/network/ovs/__init__.py
fnordahl: :+1:

charmhelpers/contrib/network/ovs/ovn.py
fnordahl: Prior to Focal and Ussuri UCA OVN was distributed as a part of Open vSwitch and by default placement of config and runtime files was together.  From Focal and Ussuri onwards they are distributed as two separate source packages and default to have configuration and runtime files separate.

A side effect of how the packaging is done pre-Ussuri and post-Ussuri this is a good way to determine which world we live in.

The consuming charm uses this information to figure out in which world we currently live in and may call this function before packages have been installed.  Given we first enabled OVN as a tech preview at Train and the default onward will be the new world order I thought it to be useful to default to the correct value.

Raising an exception here would make the current use and purpose of the function moot and it would basically have to be re-implemented in the consuming charms.

charmhelpers/contrib/network/ovs/ovn.py
fnordahl: Parenthesis fixed, see explanation above.

charmhelpers/contrib/network/ovs/ovn.py
fnordahl: This is by design.  The output contain some fields that are multi-line/multi-value which are encoded in an array.  Single-line multi-value fields are encoded in a tuple and single-line single-value fields are plain strings

I'll add a couple of comments to make it more clear what's going on.

charmhelpers/contrib/network/ovs/utils.py
fnordahl: Reviewing the usage of this utility function I see no need for capturing ``STDERR`` at this time, I'll drop it and we can add handling of that as needed.

The original version of the code used the modern ``subprocess.run`` interface which give you a separated view of ``STDOUT`` and ``STDERR``in the ``CompletedProcess`` response, but sadly we cannot use such modern interfaces for code in charm-helpers, so it got lost in translation.

charmhelpers/contrib/network/ovs/__init__.py
fnordahl: I deliberately kept both as the existing functions do extra work not strictly confined to adding bridges and ports and we need to work out how to handle the transition, so I suggest we handle that in a follow-up.

charmhelpers/contrib/network/ovs/__init__.py
fnordahl: :+1:

charmhelpers/contrib/network/ovs/ovsdb.py
fnordahl: Since most all OVS/OVN command line tools that interface to a underlying OVSDB Server support the command line interfaces used the ``SimpleOVSDB`` class may be useful for tools we do not know about or have yet envisioned.  So I think validation of the specific tool used here may be limiting the user.  We could of course log/print a warning if used with something we currently don't know about?

charmhelpers/contrib/network/ovs/ovn.py
ajkavanagh: Okay, I tend to accept the reasoning.  But would it be better to just check for the non-default path, return that if it exists, and otherwise return the default path.  i.e. only check for the non-default path location?

charmhelpers/contrib/network/ovs/ovsdb.py
ajkavanagh: It's a library function; we should always verify that it's the things that we allow rather than being permissive -- otherwise bugs (like getting the wrong tool name) propagate to weird places in the code.

fnordahl: Yes, you are right.

charmhelpers/contrib/network/ovs/__init__.py
ajkavanagh: Software engineering issues: cache invalidation, naming things and off-by-one errors, etc.  Can I just point out that `add_bridge` is a MUCH, MUCH better name than `add_br` and you should strive to keep the former rather than the latter.

charmhelpers/contrib/network/ovs/ovn.py
fnordahl: Sure, I'll also revisit the actual need for the helpers, we might also make the library functions default to the current world order and have the consumer optionally override it.

I think this came about before I had proper release selection coded into the charms.

charmhelpers/contrib/network/ovs/ovn.py
```diff
@@ -99,19 +99,27 @@ def cluster_status(target, schema=None):
...
                     v = v.replace(')', '')
                     status[k] = tuple(v.split())
                 else:
                     status[k] = v.lstrip()
             else:
+                # this is a line with only key which means a multi-line/
+                # multi-value item.  Store key as List which will be
+                # populated on subsequent iterations.
                 status[k] = []
     return status
```
ajkavanagh: I can't comment at the top of the function, as there are no changes there, so this comment is about the whole `cluster_status` function.

The return value is complex.  It's a `Dict[str, ANY]`, but that means that every calling function has to dis-ambiguate and test each value from the dictionary to process it.  Yet, that information is *known* here.  Options:

1. Is this 3 separate functions?  If there are 3 different types of return values (in the dictionary), is the function trying to do to much.
2. If it's not, then can the different return types be separated into different dictionaries.  i.e. return a tuple:  `Tuple[Dict[str, str], Dict[str, List[str]], Dict[str, Tuple[str,str,...]]`  -- that could also be a NamedTuple to make accessing it more usable.
* Can the values be tagged in some way to indicate what they are.  e.g. NameTuples rather than Tuples.  classes?

I'm really concerned that this library function is essentially returning quite an opaque object that requires reading the code to work out what is being returned.

charmhelpers/contrib/network/ovs/__init__.py
AurelienLourot: @ajkavanagh The "problem" is that we need to give the new one with new signature a new name and keep the old one for a while, while deprecating it, so we remain backward compatible for a while. Or should we just not be backward compatible?

ajkavanagh: What does the old `add_bridge` do differently to the new `add_br` that requires deprecation and the introduction of a new function?

AurelienLourot: `add_brige` takes a `datapath_type` optional argument, that neither OVN nor neutron-gateway need, but another charm somewhere out there might be using it. But true, we could take this arg as well in the signature of `add_br` and name it `add_bridge`

fnordahl: OK, I guess merging the functions now it is then.

charmhelpers/contrib/network/ovs/ovn.py
fnordahl: The system provides its status using quite opaque types (both short and long identifiers, lists etc.).  I think having one function understand the tools output is valuable rather than having to query the system status three times to get the required information from the same output, even though it has to handle three types of data in one output.

I guess putting this into a named and typed data structure rather than a ``Dict`` can have value.  You would still need to read the code defining that data structure to understand it and use it, but it will undeniably be more clear.

The flip side is that it would require maintenance to the library any/every time the upstream project adds information to the output, but I guess it does make sense to handle that at the core, where there also is expert knowledge about what every item really means as opposed to letting downstream code cope with it as best it can.

charmhelpers/contrib/network/ovs/__init__.py
ajkavanagh: I guess what is at issue here, is that we have several similar functions that do functionally similar things being merged into library and we need to be happy that we're not duplicating things or making the user experience worse.  Now it may be *correct* to actually deprecate and go with the 'worse' name as a) the world is not perfect, and b) thus where the library started from doesn't allow for a better name at this point.

My questions may be covering old ground, so if that's the case, I'll go away :)

charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -303,3 +306,158 @@ def port_to_br(port):
...
+    :type exclusive: bool
+    :raises: subprocess.CalledProcessError
+    """
+    cmd = ['ovs-vsctl', 'add-br', bridge]
+    if not exclusive:
+        cmd.append('--may-exist')
+    if brdata:
+        for setcmd in _dict_to_vsctl(brdata, 'bridge', bridge):
+            cmd.extend(setcmd)
+    utils._run(*cmd)
```
AurelienLourot: This seems to be broken. In my case I call it like this:

```
            add_br(
                'br-int', brdata={
                    'external-ids': {
                        OVS_EXTERNAL_ID_KEY: OVS_EXTERNAL_ID_BRIDGE_VALUE
                    }
                }
            )

```

and it produces the following command:

```
ovs-vsctl add-br br-int --may-exist -- set bridge br-int external-ids:charm-neutron-gateway=managed
```

which exits with 1 and prints:

```
ovs-vsctl: 'add-br' command takes exactly 1 or 3 arguments
```

It looks like the former `add_bridge()` was producing a slightly different command that would look like `ovs-vsctl -- --may-exist add-br ...`

AurelienLourot: Suggested fix: https://github.com/AurelienLourot/charm-neutron-gateway/commit/57e27117
Works for charm-neutron-gateway

charmhelpers/contrib/network/ovs/__init__.py
fnordahl: No, it's fine and your input is very valuable and needed. You're holding the bar high while I'm suffering from tunnel vision having fixed my eyes on one point in the distance for too long and trying to rushing through to get there.

Having multiple functions which serves more or less the same purpose when not needed is just adding to the pile of technical debt.

As both you and @AurelienLourot has pointed out the wanted functionality can be added to the existing functions without breaking the existing API, so doing that is the right thing to do.

charmhelpers/contrib/network/ovs/__init__.py
fnordahl: Yes, that got lost in translation apparently, ta.

charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -83,36 +84,171 @@ def get_bridges_and_ports_map():
...
+    """
     log('Creating bridge {}'.format(name))
-    cmd = ["ovs-vsctl", "--", "--may-exist", "add-br", name]
+    cmd = ['ovs-vsctl', '--']
+    if not exclusive:
+        cmd.append('--may-exist')
+    cmd.extend(('add-br', name))
+    if brdata:
+        for setcmd in _dict_to_vsctl_set(brdata, 'bridge', name):
+            cmd.extend(setcmd)
```
ajkavanagh: Could we maybe add a log message indicating that the param is deprecated when it  is used.  This will light up in charms that still use it providing an opportunity to remove the paramter at some point?

AurelienLourot: you're talking about `datapath_type`, right? I agree

fnordahl:
Functional validation of the change can be viewed in test results of https://review.opendev.org/#/c/716538/

PR #444, opened 2020-03-30
Add methods for passing & rendering osd-setting

gnuoy:
Adds the following methods for handling osd-settings be passed from
ceph client applications to the osds via the mons.

* get_osd_settings: Get osd settings and check they are whitelisted
  do not conflist with settings recieved from other clients.
* send_osd_settings: Take the osd settings from the clients and pass
  them on to the OSDs
* CephOSDConfContext: Context which consolidates settings provided
  via config flags with ones provided via clients.


charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -1635,5 +1718,49 @@ def __call__(self):
...
+            permitted_sections=permitted_sections)
+        try:
+            self.settings_from_mons = get_osd_settings('mon')
+        except OSDSettingConflict:
+            log(
+                "OSD settings from mons are inconsistent, ignoring them",
+                level=WARNING)
+            self.settings_from_mons = {}
+
+    def filter_osd_from_mon_settings(self):
```
ajkavanagh, line 1747: This function probably needs a docstring to describe what the return values are.

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -1635,5 +1718,49 @@ def __call__(self):
...
+            if key in ceph_conf.get('osd', {}):
+                if ceph_conf['osd'][key] != value:
+                    conflicting_entries[key] = value
+            else:
+                clear_entries[key] = value
+        clear_entries = _order_dict_by_key(clear_entries)
+        conflicting_entries = _order_dict_by_key(conflicting_entries)
+        return ceph_conf, clear_entries, conflicting_entries
+
+    def __call__(self):
```
ajkavanagh, line 1762: Be good to document the return value.

PR #448, opened 2020-04-08
openstack: Move bridge port interface mapping to charm helpers

fnordahl:
This will allow uniform handling of network configuration
across ``neutron-openvswitch``, ``neutron-gateway`` and
the OVN chassis charms.

Note that the handling of this detail in the charm is
pre-deprecated.

The long term goal is for network connectivity detail to be
modelled in the server provisioning layer (such as MAAS) which in
turn will provide a Netplan YAML description that will be used to
drive Open vSwitch.

Until we get to that reality the charm will need to configure
this detail based on application level configuration options.

### DIFF
--- a/charmhelpers/contrib/network/ovs/__init__.py
+++ b/charmhelpers/contrib/network/ovs/__init__.py
@@ -258,6 +258,59 @@ def del_bridge_port(name, port):
     subprocess.check_call(["ip", "link", "set", port, "promisc", "off"])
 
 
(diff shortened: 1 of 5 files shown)


fnordahl:
Note that it will most likely be easier to review the changes on the backdrop of https://review.opendev.org/#/q/project:openstack/charm-neutron-openvswitch+topic:ovn-charm+status:open

PR #449, opened 2020-04-10
Clean-up around adding linux bridges to OVS bridges

AurelienLourot:
Cleans things up around `ifup` for adding a linux bridge to an OVS bridge and handle the fact that this isn't supported anymore.

Also useful for marking the corresponding ports as managed by us.

- https://review.opendev.org/717074
- https://bugs.launchpad.net/charm-neutron-gateway/+bug/1809190


charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -320,25 +325,40 @@ def add_ovsbridge_linuxbridge(name, bridge, ifdata=None):
...
+        # 'ip link set <name> up' as the latter won't parse
+        # /etc/network/interfaces*
+        ifup_cmd = ['ifup', linuxbridge_port]
+        try:
+            subprocess.check_call(ifup_cmd)
+        except FileNotFoundError:
+            # NOTE(lourot): on bionic and newer, 'ifup' isn't installed by
+            # default
+            apt_install('ifupdown', fatal=True)
+            subprocess.check_call(ifup_cmd)
```
ajkavanagh, line 358: I'm fairly against this idea of "automatic" apt install of a library in the c-h code.  This is a dependency that the charm author ought to become aware of and deal with in the pre-install of the charm.

Also, but on bionic what is the alternative to ifup for the user, if it's not installed by default.  Should the code be using that instead as that's the new 'normal' going forwards?

charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -410,6 +434,12 @@ def check_for_eni_source():
         eni.write('\nsource /etc/network/interfaces.d/*')
 
 
+def check_for_eni_source():
+    """DEPRECATED: call setup_eni() instead."""
+    log('DEPRECATION WARNING: call setup_eni() instead.')
```
ajkavanagh, line 439: There's a handy decorator called `@deprecate(...)` from charmhelpers (i.e. from charmhelpers import deprecate) that provides deprecation warnings, etc.  It can also do single use, etc.

charmhelpers/contrib/network/ovs/__init__.py
AurelienLourot: Good point. It looks like netplan.io is the new `/etc/network/interfaces`. I'll give it a try.

AurelienLourot: Actually we have decided meanwhile not to support this anymore: https://review.opendev.org/#/c/728382/

charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -343,6 +345,14 @@ def add_ovsbridge_linuxbridge(name, bridge, ifdata=None):
                20dac08fdcce4b7fda1d07add3b346aa9751cfbc/
                    lib/db-ctl-base.c#L189-L215
     :type ifdata: Optional[Dict[str,Union[str,Dict[str,str]]]]
+    :param portdata: Additional data to attach to port. Similar to ifdata.
+    :type portdata: Optional[Dict[str,Union[str,Dict[str,str]]]]
+
+    WARNING: the `ifup` command (NetworkManager) must be available on the system
```
ajkavanagh, line 351: Shouldn't we detect that it's not available and use the alternative on alter Ubuntu releases?

charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -377,25 +387,43 @@ def add_ovsbridge_linuxbridge(name, bridge, ifdata=None):
...
+        with open('/etc/network/interfaces.d/{}.cfg'.format(
+                linuxbridge_port), 'w') as config:
+            config.write(BRIDGE_TEMPLATE.format(
+                linuxbridge_port=linuxbridge_port,
+                ovsbridge_port=ovsbridge_port, bridge=bridge))
+
+        try:
+            # NOTE(lourot): 'ifup <name>' can't be replaced by
+            # 'ip link set <name> up' as the latter won't parse
+            # /etc/network/interfaces*
```
ajkavanagh, line 413: Okay, I see why this is the case for ifup but ... (see next comment)

charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -377,25 +387,43 @@ def add_ovsbridge_linuxbridge(name, bridge, ifdata=None):
...
+            subprocess.check_call(['ifup', linuxbridge_port])
+        except FileNotFoundError:
+            # NOTE(lourot): on bionic and newer, 'ifup' isn't installed by
+            # default. It has been replaced by netplan.io but we can't use it
+            # yet because of lp:1876730. For the time being, charms using this
+            # have to install 'ifupdown' on bionic and newer. This will however
+            # cause issues when deploying to LXD, see lp:1877594.
```
ajkavanagh, line 420: This seems problematic doesn't it?  i.e. is there not a solution for bionic or newer (which is basically every new install from now on) with LXD?

charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -455,10 +483,14 @@ def get_certificate():
         return None
 
 
-def check_for_eni_source():
-    ''' Juju removes the source line when setting up interfaces,
-    replace if missing '''
+def setup_eni():
+    """Makes sure /etc/network/interfaces.d/ exists and will be parsed.
 
+    Indeed when setting up interfaces, Juju removes from
```
ajkavanagh, line 489: Delete "Indeed ", and replace with "When setting up ..."

charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -455,10 +483,14 @@ def get_certificate():
...
+    /etc/network/interfaces the line sourcing interfaces.d/
+    """
+    if not os.path.exists('/etc/network/interfaces.d'):
+        os.makedirs('/etc/network/interfaces.d', mode=0o755)
     with open('/etc/network/interfaces', 'r') as eni:
         for line in eni:
             if line == 'source /etc/network/interfaces.d/*':
```
ajkavanagh, line 496: I know it's not part of this review, but this search feels really brittle!  Why not something like:

     re.search(r"^source\s+/etc/network/interfaces.d/\*\s*$")

Which would ignore space issues? (obviously, I've not actually tested that regex, but it would be something like that).

AurelienLourot:
As discussed with @thedac we want to be more opinionated here and neither create `/etc/network/interfaces*` nor call `ifup` on focal and newer. I'll rework this PR.

PR #451, opened 2020-04-15
ovs/ovsdb: Improve SimpleOVSDB API

fnordahl:
The current approach is string heavy and could fit better with how
Python developers expect data structures to behave.

As there has been no release of charm-helpers since the initial
commit of the API it should be safe to change it at this point.

A future improvement could still be to add a third dimension to
the data structure to allow users to operate on specific key/value
pairs in a specifc row in a specific table.  At this point in time
the underlying command line argument that would be used does not
support outputting data in JSON format so for that reason I'm
postponing implementation of that facet.

### DIFF
--- a/charmhelpers/contrib/network/ovs/ovsdb.py
+++ b/charmhelpers/contrib/network/ovs/ovsdb.py
@@ -28,74 +28,183 @@ class SimpleOVSDB(object):
     appears to be a bit too involved for our simple use case.
 
     Examples:
-    chassis = SimpleOVSDB('ovn-sbctl', 'chassis')
-    for chs in chassis:
+    sbdb = SimpleOVSDB('ovn-sbctl')
+    for chs in sbdb.chassis:
         print(chs)
 
-    bridges = SimpleOVSDB('ovs-vsctl', 'bridge')
-    for br in bridges:
+    ovsdb = SimpleOVSDB('ovs-vsctl')
+    for br in ovsdb.bridge:
         if br['name'] == 'br-test':
-            bridges.set(br['uuid'], 'external_ids:charm', 'managed')
+            ovsdb.bridge.set(br['uuid'], 'external_ids:charm', 'managed')
     """
 
-    def __init__(self, tool, table):
-        """SimpleOVSDB constructor
+    # For validation we keep a complete map of currently known good tool and
+    # table combinations.  This requires maintenance down the line whenever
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/network/ovs/ovsdb.py
```diff
@@ -28,74 +28,183 @@ class SimpleOVSDB(object):
...
         :type tool: str
-        :param table: Which table to operate on
-        :type table: str
         """
-        if tool not in ('ovs-vsctl', 'ovn-nbctl', 'ovn-sbctl'):
+        if tool not in self._tool_table_map:
             raise RuntimeError(
-                "tool must be one of 'ovs-vsctl', 'ovn-nbctl', 'ovn-sbctl'")
+                'tool must be one of "{}"'.format(self._tool_table_map.keys()))
         self.tool = tool
```
ajkavanagh, line 134: I'm a bit confused as to why there is a `_tool` and a `tool`?

charmhelpers/contrib/network/ovs/ovsdb.py
```diff
@@ -28,74 +28,183 @@ class SimpleOVSDB(object):
...
-
-        :param condition: An optional RFC 7047 5.1 match condition
-        :type condition: Optional[str]
-        :returns: Dictionary with data
-        :rtype: Iterator[Dict[str, ANY]]
+        self._tool = tool
+        self._tableobject = {}
+
+    def __getattr__(self, table):
+        if table not in self._tool_table_map[self._tool]:
```
ajkavanagh, line 139: It took me a little while to work it out, but this class is building a set of singleton `Table` objects on demand?  Whilst that would be useful if the objects held state (which they don't seem to?), all this is doing is keeping the objects around after they have been used until the parent `SimpleOVSDB` goes away ... i.e goes out of scope?  Is this what is intended?

If not, why memoise the object, and instead just hand out a new `Table` object after it's been verified that the table sting is valid?

charmhelpers/contrib/network/ovs/ovsdb.py
fnordahl: Thank you for catching that :+1:

charmhelpers/contrib/network/ovs/ovsdb.py
fnordahl: I guess you caught me red handed in performing premature optimization! Gone. :+1:

PR #452, opened 2020-04-16
Add hook context to unit test helper

gnuoy:
Add support to tests.helper.FakeRelation for setting a implicit
relation context. Also add support for relation_id. This exposed
a mistake in two other unit tests, so fixes to apacheutils tests
included too.

set_relation_context can be used to simulate being in relation hook
context, eg omitting a relation id or unit implies the query is
against the relation id and unit in the current hook context. To set
the context:

    relation = FakeRelation(rel)
    relation.set_relation_context('mysql-svc2/0', 'shared-db:12')

To clear it:

    relation.clear_relation_context()

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -71,9 +71,11 @@ class FakeRelation(object):
         self.relation_get.side_affect = relation.get
         passwd = self.relation_get('password', rid='mysql:0', unit='mysql/0')
 
-    set_relation_context can be used to simulate being in relation hook
-    context, eg omitting a relation id or unit implies the query is against
-    the relation id and unit in the current hook context. To set the context:
+    set_relation_context can be used to simulate being in a relation hook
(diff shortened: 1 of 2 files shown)


tests/helpers.py
```diff
@@ -70,12 +70,41 @@ class FakeRelation(object):
         relation = FakeRelation(rel)
         self.relation_get.side_affect = relation.get
         passwd = self.relation_get('password', rid='mysql:0', unit='mysql/0')
+
+    set_relation_context can be used to simulate being in relation hook
+    context, eg omitting a relation id or unit implies the query is against
```
ChrisMacNaughton, line 75: I think it would be more clear to say: set_relation_context can be used to simulate being in a relation hook context. This allows omitting a relation id or unit when calling relation helpers as the related unit is present

PR #454, opened 2020-04-16
Add common functions for sharing unit state

gnuoy:
Add a set of functions which can be used by openstack charms to
perform common unit state checks and advertise/consume unit state
down a peer relation.

### DIFF
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -2220,7 +2220,7 @@ UNIT_STATES = [UNIT_READY, UNIT_NOTREADY, UNIT_UNKNOWN]
 
 
 def get_peers_unit_state(relation_name='cluster'):
-    """Get peers the state of all peers.
+    """Get the state of all peers.
 
     :param relation_name: Name of relation to check peers on.
     :type relation_name: string
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -2199,6 +2199,12 @@ def get_peer_key(unit_name):
     return 'unit-state-{}'.format(unit_name.replace('/', '-'))
 
 
+UNIT_READY = 'READY'
+UNIT_NOTREADY = 'NOTREADY'
+UNIT_UNKNOWN = 'UNKNOWN'
+UNIT_STATES = [UNIT_READY, UNIT_NOTREADY, UNIT_UNKNOWN]
+
+
 def inform_peers_unit_state(state, relation_name='cluster'):
     """Inform peers of the state of this unit.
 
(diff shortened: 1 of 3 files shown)


charmhelpers/contrib/openstack/utils.py
```diff
@@ -2183,3 +2183,133 @@ def _get_relation_id(rel_type):
...
+    :raises: ValueError
+    """
+    r_ids = relation_ids(relation_name)
+    rids_units = [(r, u) for r in r_ids for u in related_units(r)]
+    unit_states = {}
+    for r_id, unit in rids_units:
+        settings = relation_get(unit=unit, rid=r_id)
+        unit_states[unit] = settings.get(get_peer_key(unit), UNIT_UNKNOWN)
+        if unit_states[unit] not in UNIT_STATES:
+            raise ValueError(
```
ChrisMacNaughton, line 2238: If we're validating on the get, I think that we should validate on the set as well (line 2202)

charmhelpers/contrib/openstack/utils.py
```diff
@@ -2183,3 +2183,133 @@ def _get_relation_id(rel_type):
...
+                "Unit in unknown state {}".format(unit_states[unit]))
+    return unit_states
+
+
+def is_peers_ready(relation_name='cluster'):
```
ChrisMacNaughton, line 2243: `are_peers_ready` please :)

charmhelpers/contrib/openstack/utils.py
```diff
@@ -2183,3 +2183,133 @@ def _get_relation_id(rel_type):
...
+    :type relation_name: string
+    :returns: Whether all units are ready.
+    :rtype: bool
+    """
+    unit_states = get_peers_unit_state(relation_name)
+    return all(v == UNIT_READY for v in unit_states.values())
+
+
+def inform_peers_if_ready(check_unit_ready_func, relation_name='cluster'):
+    """If this unit is ready inform peers.
```
ChrisMacNaughton, line 2256: "Inform peers if this unit is ready" would sound better

charmhelpers/contrib/openstack/utils.py
```diff
@@ -2183,3 +2183,133 @@ def _get_relation_id(rel_type):
...
+    :type check_unit_ready_func: function
+    :param relation_name: Name of relation to check peers on.
+    :type relation_name: string
+    """
+    unit_ready, msg = check_unit_ready_func()
+    if unit_ready:
+        state = UNIT_READY
+    else:
+        state = UNIT_NOTREADY
+    juju_log('Telling peers this unit is: {}'.format(msg), 'DEBUG')
```
ChrisMacNaughton, line 2271: This function is actually just telling peers that the unit is UNIT_READY or UNIT_NOTREADY, I'm not sure that we need the msg from the check_unit_ready_func

charmhelpers/contrib/openstack/utils.py
```diff
@@ -2183,3 +2183,136 @@ def _get_relation_id(rel_type):
...
+
+    The check function should return a tuple (state, message). A state
+    of 'READY' indicates the unit is READY.
+
+    :param check_unit_ready_func: Function to run to check readiness
+    :type check_unit_ready_func: function
+    :param relation_name: Name of relation to check peers on.
+    :type relation_name: string
+    """
+    unit_ready, msg = check_unit_ready_func()
```
ChrisMacNaughton, line 2269: tiny nit that we don't really need to assign `msg` anymore as it's not used, but I'll let it pass :)

PR #456, opened 2020-04-20
Add support for setting application status

gnuoy:
The status-set command can take a 'application' switch which causes
the state and message to be set at the application level rather
than the unit level. This change adds support for that switch.

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -57,6 +57,17 @@ RANGE_WARNING = ('Passing NO_PROXY string that includes a cidr. '
                  'This may not be compatible with software you are '
                  'running in your shell.')
 
+WL_STATE_ACTIVE = 'active'
+WL_STATE_BLOCKED = 'blocked'
+WL_STATE_MAINTENANCE = 'maintenance'
+WL_STATE_WAITING = 'waiting'
+
+WL_STATES = [
+    WL_STATE_ACTIVE,
+    WL_STATE_BLOCKED,
+    WL_STATE_MAINTENANCE,
+    WL_STATE_WAITING]
+
 cache = {}
 
 
@@ -1088,22 +1099,26 @@ def function_tag():
     return os.environ.get('JUJU_FUNCTION_TAG') or action_tag()
 
 
-def status_set(workload_state, message):
+def status_set(workload_state, message, application_status=False):
     """Set the workload state with a message
 
     Use status-set to set the workload state with a message which is visible
     to the user via juju status. If the status-set command is not found then
     assume this is juju < 1.23 and juju-log the message instead.
 
-    workload_state -- valid juju workload state.
-    message        -- status update message
+    workload_state     -- valid juju workload state.
+    message            -- status update message
+    application_status -- Whether this is an application state set
     """
-    valid_states = ['maintenance', 'blocked', 'waiting', 'active']
-    if workload_state not in valid_states:
+    workload_state = workload_state.lower()
+    if workload_state not in WL_STATES:
         raise ValueError(
             '{!r} is not a valid workload state'.format(workload_state)
         )
-    cmd = ['status-set', workload_state, message]
+    cmd = ['status-set']
+    if application_status:
+        cmd.append('--application')
+    cmd.extend([workload_state, message])
     try:
         ret = subprocess.call(cmd)
         if ret == 0:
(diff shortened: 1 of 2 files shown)


charmhelpers/core/hookenv.py
```diff
@@ -57,6 +57,17 @@
...
+WL_STATE_BLOCKED = 'blocked'
+WL_STATE_MAINTENANCE = 'maintenance'
+WL_STATE_WAITING = 'waiting'
+
+WL_STATES = [
+    WL_STATE_ACTIVE,
+    WL_STATE_BLOCKED,
+    WL_STATE_MAINTENANCE,
+    WL_STATE_WAITING]
+
```
ajkavanagh, line 70: As we are Python3 here, why not use an `Enum`: https://docs.python.org/3/library/enum.html

gnuoy: Thanks for the tip, I've switched to an Enum and its much tidier

timClicks:
Awesome. Was literally just looking for this functionality. This change needs a corresponding update to the docs.

charmhelpers/core/hookenv.py
```diff
@@ -57,6 +58,14 @@
                  'This may not be compatible with software you are '
                  'running in your shell.')
 
+
+class WL_STATES(Enum):
```
timClicks, line 62: What does `WL_` stand for?

timClicks: Assuming "workload", why is it not expanded fully?

charmhelpers/core/hookenv.py
```diff
@@ -1088,22 +1097,31 @@ def function_tag():
     return os.environ.get('JUJU_FUNCTION_TAG') or action_tag()
 
 
-def status_set(workload_state, message):
+def status_set(workload_state, message, application_status=False):
```
timClicks, line 1100: IMO there's no need for this parameter to have the `_status` suffix.

charmhelpers/core/hookenv.py
```diff
@@ -1088,22 +1097,31 @@ def function_tag():
...
     """
-    valid_states = ['maintenance', 'blocked', 'waiting', 'active']
-    if workload_state not in valid_states:
+    # Extract the value if workload_state is an Enum
+    try:
+        workload_state = workload_state.value
+    except AttributeError:
+        pass
+    workload_state = workload_state.lower()
+    if workload_state not in [s.lower() for s in WL_STATES.__members__.keys()]:
```
timClicks, line 1117: I recommend implementing `__contains__()` within the enum itself. All of the complexity could then be re-used without becoming distracting.

charmhelpers/core/hookenv.py
```diff
@@ -1088,22 +1097,31 @@ def function_tag():
...
         raise ValueError(
             '{!r} is not a valid workload state'.format(workload_state)
         )
-    cmd = ['status-set', workload_state, message]
+    cmd = ['status-set']
+    if application_status:
+        cmd.append('--application')
+    cmd.extend([workload_state, message])
     try:
         ret = subprocess.call(cmd)
```
timClicks, line 1126: We should be prepared for the hook to fail. `status-set --application` will error when it's called by a non-leader.

Perhaps we should call `is-leader` early and raise an error directly?

charmhelpers/core/hookenv.py
gnuoy: I'll address this in a subsequent PR

charmhelpers/core/hookenv.py
gnuoy: I'll address this too

charmhelpers/core/hookenv.py
gnuoy: I don't think that is the right thing to do. __contains__ is already implemented in enum.EnumMeta and correctly checks if an Enum is a member of the Enum class. What you are proposing seems to be making __contains__ handle strings rather than Enums which feels wrong.

charmhelpers/core/hookenv.py
gnuoy: I agree and I wish the original author (which unfortunately git blame seems to suggest was me) had done that but making that change now would be a change in behaviour on a function that has been around for 5 years. I don't want to inadvertently break any code that is expected a subprocess error when running on a non-leader

PR #457, opened 2020-04-20
Small refactor of the status check functions

gnuoy:
Small refactor of the status check functions. Backwards compatability
is not an issue for this as no charms are currently using them.

### DIFF
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -36,6 +36,7 @@ from charmhelpers.contrib.network import ip
 from charmhelpers.core import unitdata
 
 from charmhelpers.core.hookenv import (
+    WL_STATES,
     action_fail,
     action_set,
     config,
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/utils.py
```diff
@@ -2283,36 +2294,60 @@ def check_api_unit_ready(check_db_ready=True):
...
-    if are_peers_ready():
-        return True, 'All units have passed checks and are ready'
+    if unit_ready and are_peers_ready():
+        app_state = WL_STATES.ACTIVE
+        msg = 'Application is ready'
     else:
-        return False, 'This unit is ready but peers are not'
+        app_state = WL_STATES.WAITING
+        msg = 'Some units not ready'
+    return app_state, msg
```
johnsca, line 2353: This is obscuring both the state (potentially `blocked`) and the message from `get_api_unit_status()`.  If these are not yet used, it seems like it would be better to drop `check_api_application_ready()` and invert the check here.  Something like:

```python
    app_state, msg = get_api_unit_status()
    if app_state == WL_STATES.ACTIVE and not are_peers_ready():
        return WL_STATES.WAITING, 'Some units are not ready'
    return app_state, msg
```

This way, a non-`active` state for this unit (and in particular, a `blocked` state) would take precedence over a peer not being ready.

gnuoy: Thank you for the review, I agree with the points you raised and I'll address them in a subsequent PR.

PR #460, opened 2020-04-21
Handle exceptions in containers when setting sysctl keys

andrewdmcleod:
Prior to focal, when setting some sysctl keys in a container, sysctl returned an exit code of 0. Now, the same error returns a 255. This commit ensures the same behaviour as prior to focal.

### DIFF
--- a/charmhelpers/core/sysctl.py
+++ b/charmhelpers/core/sysctl.py
@@ -17,12 +17,16 @@
 
 import yaml
 
-from subprocess import check_call
+from subprocess import check_call, CalledProcessError
 
 from charmhelpers.core.hookenv import (
     log,
     DEBUG,
     ERROR,
+)
+
+from charmhelpers.core.host import (
+    is_containr,
 )
 
 __author__ = 'Jorge Niedbalski R. <jorge.niedbalski@canonical.com>'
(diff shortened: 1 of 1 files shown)


charmhelpers/core/sysctl.py
```diff
@@ -62,4 +64,10 @@ def create(sysctl_dict, sysctl_file, ignore=False):
     if ignore:
         call.append("-e")
 
-    check_call(call)
+    try:
+        check_call(call)
+    except CalledProcessError as e:
+        if is_container():
+            print("Error setting some systcl keys in this container: {}".format(e.output))
```
javacruft, line 71: Please can we use logging rather than just print here.

javacruft (dismissed):

charmhelpers/core/sysctl.py
```diff
@@ -62,4 +64,11 @@ def create(sysctl_dict, sysctl_file, ignore=False):
     if ignore:
         call.append("-e")
 
-    check_call(call)
+    try:
+        check_call(call)
+    except CalledProcessError as e:
+        if is_container():
+            log("Error setting some systcl keys in this container: {}".format(e.output),
```
javacruft, line 71: typo - sysctl (not systcl) and please can we log this as a warning - its not fatal so it not an error.

PR #464, opened 2020-04-21
Use paramertierised sets for lookups rather than hardcoded lists

timClicks:
Improve the O() behaviour by using sets rather than lists. Also allow callers to supply their own patterns for matching.

Makes the bool_from_string more flexible by allowing callers to opt-out of raising exceptions.

### DIFF
--- a/charmhelpers/core/strutils.py
+++ b/charmhelpers/core/strutils.py
@@ -18,8 +18,10 @@
 import six
 import re
 
+TRUTHY_STRINGS = {'y', 'yes', 'true', 't', 'on'}
+FALSEY_STRINGS = {'n', 'no', 'false', 'f', 'off'}
 
-def bool_from_string(value):
+def bool_from_string(value, truthy_strings=TRUTHY_STRINGS, falsey_strings=FALSEY_STRINGS, assume_false=False):
     """Interpret string value as boolean.
 
     Returns True if value translates to True otherwise False.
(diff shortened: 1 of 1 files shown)


charmhelpers/core/strutils.py
```diff
@@ -32,9 +34,12 @@ def bool_from_string(value):
 
     value = value.strip().lower()
 
-    if value in ['y', 'yes', 'true', 't', 'on']:
+    if value in truthy_strings:
         return True
-    elif value in ['n', 'no', 'false', 'f', 'off']:
+    elif value in falsey_strings:
+        return False
+
```
johnsca, line 41: charmhelpers/core/strutils.py:41:1: W293 blank line contains whitespace

charmhelpers/core/strutils.py
```diff
@@ -18,8 +18,10 @@
 import six
 import re
 
+TRUTHY_STRINGS = {'y', 'yes', 'true', 't', 'on'}
+FALSEY_STRINGS = {'n', 'no', 'false', 'f', 'off'}
 
-def bool_from_string(value):
+def bool_from_string(value, truthy_strings=TRUTHY_STRINGS, falsey_strings=FALSEY_STRINGS, assume_false=False):
```
johnsca, line 24: charmhelpers/core/strutils.py:24:1: E302 expected 2 blank lines, found 1

charmhelpers/core/strutils.py
timClicks: :thinking: I actually added this blank line intentionally because I wanted readers to be able to keep the reporting behaviour to be mentally separate from the prior tests. I'm happy to fold them in together.

johnsca: I don't think the blank line was the issue, it was that it contained whitespace.  I like the new version better, though.

PR #468, opened 2020-05-04
openstack: Add context for SRIOV configuration

fnordahl:
Implement the ``neutron-openvswitch`` charm ``configure_sriov``
helper function as a context.

This is useful both for the interrim and long term use case where
we now will write a interface name and number of VFs to a YAML
file.

As of this writing it will be used with the ``sriov-netplan-shim``
package, long term it can be re-used to augment Netplan config.

Co-Authored-By: James Page <james.page@ubuntu.com>

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -54,7 +54,8 @@ from charmhelpers.core.hookenv import (
     INFO,
     ERROR,
     status_set,
-    network_get_primary_address
+    network_get_primary_address,
+    WARNING,
 )
 
 from charmhelpers.core.sysctl import create as sysctl_create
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/context.py
```diff
@@ -2988,3 +2988,91 @@ def get_ovs_portdata(self, bond):
...
+            for pair in sriov_device_mappings.split()
+        ]
+
+        # create list of devices from sriov_numvfs config option
+        sriov_config_devices = {}
+        try:
+            for ifname, numvfs in (
+                    pair.split(':', 1)
+                    for pair in sriov_numvfs.split()):
+                sriov_config_devices.update({ifname: numvfs})
```
ajkavanagh, line 3025: Perhaps we could re-write this as:

```python
sriov_config_devices = {
    ifname: numvfs for ifname, numvs in (
        pair.split(':', 1) for pair in sriov_numvfs.split())}
```

charmhelpers/contrib/openstack/context.py
```diff
@@ -2988,3 +2988,91 @@ def get_ovs_portdata(self, bond):
...
+                        int(device.sriov_totalvfs))
+                    if numvfs < int(
+                            sriov_config_devices[device.interface_name]):
+                        log('Requested value for sriov-numfs ({}) too high '
+                            'for interface {}. Falling back to interface '
+                            'totalvfs value: {}'
+                            .format(sriov_config_devices[device.interface_name],
+                                    device.interface_name,
+                                    device.sriov_totalvfs),
+                            level=INFO)
```
ajkavanagh, line 3062: This is probably unmaintainable :)  I don't mean that to be unkind; it's just hard to follow.  Could it be split up into functions, or include comments explaining what the code is doing.  e.g. there's a number of uses of `min()` ... it's for a reason, but unless you really understand the domain, it's not clear.

charmhelpers/contrib/openstack/context.py
fnordahl: Ah, cool, I did try to wrangle it into dict comprehension but failed, so thank you for pointing me in the direction of how.

charmhelpers/contrib/openstack/context.py
fnordahl: No worries, that is what reviews are for.

To be fair the code in question it is a lift from charm into charm-helpers exercise, and we need to carry forward the same overloading of some of the configuration options.  The code organization is flavored from it's origin in the charm it comes from.  I'll have a pass at dividing the individual handling into helper methods and add commentary to it.

ajkavanagh: Ah, that makes sense.

PR #469, opened 2020-05-04
Allow specifying port for mysql and shared-db context

thedac:


charmhelpers/contrib/database/mysql.py
```diff
@@ -83,17 +84,19 @@ def __init__(self, rpasswdf_template, upasswdf_template, host='localhost',
...
             password = self.password
         if host is None:
             host = self.host
+        if port is None:
+            port = self.port
 
         log("Opening db connection for %s@%s" % (user, host), level=DEBUG)
         self.connection = MySQLdb.connect(user=user, host=host,
-                                          passwd=password)
+                                          passwd=password, port=port)
```
ajkavanagh, line 99: I'm guessing passing `port=None` here has no effect?

thedac: Correct.

thedac:
Validation of NOT passing port info: https://review.opendev.org/725657

PR #471, opened 2020-05-11
ceph autoscaler: set min pg num for pools

javacruft:
Ensure that the minimum number of pgs is set as appropriate during
pool creation. If the number of pgs used for the initial pool creation
is less that the autoscaler default of 32, set to the lower pg
number to ensure that the autoscaler does not immediately scale
the pool up to 32 pgs.

This avoids pg overloading in smaller test deployments.

Closes-Bug: 1872748

### DIFF
--- a/charmhelpers/contrib/storage/linux/ceph.py
+++ b/charmhelpers/contrib/storage/linux/ceph.py
@@ -92,6 +92,7 @@ DEFAULT_PGS_PER_OSD_TARGET = 100
 DEFAULT_POOL_WEIGHT = 10.0
 LEGACY_PG_COUNT = 200
 DEFAULT_MINIMUM_PGS = 2
+AUTOSCALER_DEFAULT_PGS = 32
 
 
 class OsdPostUpgradeError(Exception):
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -399,16 +400,28 @@ def __init__(self, service, name, pg_num=None, replicas=2,
 
     def create(self):
         if not pool_exists(self.service, self.name):
+            nautilus_or_later = cmp_pkgrevno('ceph-common', '14.2.0') >= 0
             # Create it
-            cmd = ['ceph', '--id', self.service, 'osd', 'pool', 'create',
-                   self.name, str(self.pg_num)]
+            if nautilus_or_later:
```
ChrisMacNaughton, line 405: I wonder if it would be nicer or more clear to do something like:

```python3
cmd = ['ceph', '--id', self.service, 'osd', 'pool', 'create']
if nautilus_or_later:
    cmd.append(
        '--pg-num-min={}'.format(
             min(AUTOSCALER_DEFAULT_PGS, self.pg_num))
cmd.extend([self.name, str(self.pg_num)])
```

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -466,10 +479,24 @@ def create(self):
             k = int(erasure_profile['k'])
             m = int(erasure_profile['m'])
             pgs = self.get_pgs(k + m, self.percent_data)
+            nautilus_or_later = cmp_pkgrevno('ceph-common', '14.2.0') >= 0
             # Create it
-            cmd = ['ceph', '--id', self.service, 'osd', 'pool', 'create',
-                   self.name, str(pgs), str(pgs),
-                   'erasure', self.erasure_code_profile]
+            if nautilus_or_later:
```
ChrisMacNaughton, line 484: Same comment as above

charmhelpers/contrib/storage/linux/ceph.py
javacruft: I'm kinda split on that one - I like the clarity of all of the command that will be executed being between [] even if that does mean duplication of a couple of lines of code.

ChrisMacNaughton: I was thyinking about the potential for future conditions where we want more checks like these; however, we haven't had many changes in this code in years so probably better to be clear now, and worry about the future if things change.

PR #476, opened 2020-05-20
Add useful traces around peer readiness detection

AurelienLourot:
https://bugs.launchpad.net/charm-keystone/+bug/1818113

CI runs: https://review.opendev.org/729781


AurelienLourot:
there are charmstore instabilities at the moment, which is why the tests are failing

PR #488, opened 2020-07-09
Fix ownership for authorized_keys and known_hosts #487

### LINKED ISSUE #487: unison.ssh_authorized_peers creates files with incorrect file ownership
Hi,

Using unison.ssh_authorized_peers(), file ownership for authorized_keys and known_hosts are wrong:

`drwxr-xr-x 3 www-sync www-data 4096 Jul  8 03:54 ..`
`-rw-r--r-- 1 root     root      396 Jul  9 02:55 authorized_keys`
`-rw------- 1 www-sync root     1675 Jul  8 03:54 id_rsa`
`-rw-r--r-- 1 www-sync root      396 Jul  8 03:54 id_rsa.pub`
`-rw-r--r-- 1 root     root      659 Jul  9 02:55 known_hosts`

They should be the 'user' specified. This is also reported against the ubuntu-repository-cache charm in LP:1770071

hloeung:
Fixes #487


stub42 (approved):
All looks good, ta.

PR #489, opened 2020-07-11
Send unit name alongside ceph requests

gnuoy:
Send the unit name alongside ceph request to support cross model
relations.

### DIFF
--- a/charmhelpers/contrib/storage/linux/ceph.py
+++ b/charmhelpers/contrib/storage/linux/ceph.py
@@ -2215,6 +2215,7 @@ def send_request_if_needed(request, relation='ceph'):
         for rid in relation_ids(relation):
             log('Sending request {}'.format(request.request_id), level=DEBUG)
             relation_set(relation_id=rid, broker_req=request.request)
+            relation_set(relation_id=rid, relation_settings={'unit-name': local_unit()})
 
 
 def has_broker_rsp(rid=None, unit=None):


ChrisMacNaughton (approved):
Given that Ceph [still can't figure out remote unit names](https://github.com/openstack/charm-ceph-mon/blob/8eef7d0c39920402ec84c9b739d29834afdc6e1b/hooks/charmhelpers/core/hookenv.py#L292) after this change, I'm tentatively +1 on landing it, with the additional note that further work will be going into the ceph-mon charm to ensure that CMR is disabled by default

johnsca:
It's worth noting that the reason unit / application names aren't exposed across CMR is that they might conflict with other models. Sending the name of the unit over the relation could lead to the same key being used by applications in different models.  This should really use a unique ID; either a generated UUID (which you'd need to store), or just prepending the `JUJU_MODEL_UUID` to the unit name should work.

ChrisMacNaughton:
This has been fairly heavily discussed within the team. The general conclusion is that the Ceph charms will hard-fail by default when somebody tries to use CMR, and that an option will be added to the Ceph charms to enable "sharing" pools cross-model, which this feature enables. The working assumption with the Ceph charms, in current discussions within the team and with users, is that the cross-models would still be internal things, rather than real trust boundaries.

stub42:
I just noticed that 'juju consume' now allows you to specify the local application name, which will allow the remote application name to use be specified at deployment time, rather than needing to transmit it via the relation.

PR #491, opened 2020-07-14
contrib/ovs: Fix deletion of virtual bridge ports

fnordahl:
The OVS `del_bridge_port` helper currently assumes it will be used
with interfaces that actually exist in the system.

This is not always the case. Add arguments that allow the caller
to decide whether a interface link and promiscuous settings should
be touched after removing the port from a OVS bridge. Defaults to
being compatible with the current behaviour.

### DIFF
--- a/charmhelpers/contrib/network/ovs/__init__.py
+++ b/charmhelpers/contrib/network/ovs/__init__.py
@@ -246,20 +246,23 @@ def add_bridge_port(name, port, promisc=False, ifdata=None, exclusive=False,
         subprocess.check_call(["ip", "link", "set", port, "promisc", "off"])
 
 
-def del_bridge_port(name, port):
+def del_bridge_port(name, port, linkdown=True):
     """Delete a port from the named openvswitch bridge
 
     :param name: Name of bridge to remove port from
     :type name: str
     :param port: Name of port to remove
     :type port: str
+    :param linkdown: Whether to set link down on interface (default: True)
+    :type linkdown: bool
     :raises: subprocess.CalledProcessError
     """
     log('Deleting port {} from bridge {}'.format(port, name))
     subprocess.check_call(["ovs-vsctl", "--", "--if-exists", "del-port",
                            name, port])
-    subprocess.check_call(["ip", "link", "set", port, "down"])
-    subprocess.check_call(["ip", "link", "set", port, "promisc", "off"])
+    if linkdown:
+        subprocess.check_call(["ip", "link", "set", port, "down"])
+        subprocess.check_call(["ip", "link", "set", port, "promisc", "off"])
 
 
 def add_bridge_bond(bridge, port, interfaces, portdata=None, ifdatamap=None,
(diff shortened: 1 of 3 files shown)


tests/contrib/network/ovs/test_ovs.py
```diff
@@ -125,6 +125,29 @@ def test_add_bridge_port(self):
...
+        self.check_call.assert_has_calls([
+            mock.call(['ovs-vsctl', '--', '--if-exists', 'del-port',
+                       'test', 'eth1']),
+            mock.call(['ip', 'link', 'set', 'eth1', 'down']),
+            mock.call(['ip', 'link', 'set', 'eth1', 'promisc', 'off'])
+        ])
+        self.assertTrue(self.log.call_count == 1)
+        self.check_call.reset_mock()
+        ovs.del_bridge_port('test', 'eth1', promisc=None)
+        self.check_call.assert_has_calls([
```
AurelienLourot, line 141: `assert_has_calls` tolerates extra calls, so you need an extra check to make sure there hasn't been other calls before and after

charmhelpers/contrib/network/ovs/__init__.py
```diff
@@ -246,20 +246,28 @@ def add_bridge_port(name, port, promisc=False, ifdata=None, exclusive=False,
         subprocess.check_call(["ip", "link", "set", port, "promisc", "off"])
 
 
-def del_bridge_port(name, port):
+def del_bridge_port(name, port, promisc=False, linkdown=True):
```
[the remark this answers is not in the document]
fnordahl: I figured I would get review feedback on that one :) I basically just copied over the ternary behavior of the `add_bridge_port` function.

Actually we could probably just remove the `promisc` argument and do the promisc thing if we do linkdown.

tests/contrib/network/ovs/test_ovs.py
fnordahl: I do check all the permutations and finally do a `assert_called_once` with a reset of the mock in between each round, so I think that should cover all bases.

AurelienLourot: You're right, I think it's ok, but for the record this is what I mean:

```
    def test_del_bridge_port(self):
        self.patch_object(ovs.subprocess, 'check_call')
        self.patch_object(ovs, 'log')
        ovs.del_bridge_port('test', 'eth1')
        self.check_call.assert_has_calls([
            mock.call(['ovs-vsctl', '--', '--if-exists', 'del-port',
                       'test', 'eth1']),
            mock.call(['ip', 'link', 'set', 'eth1', 'down']),
            mock.call(['ip', 'link', 'set', 'eth1', 'promisc', 'off'])
        ])
        self.assertTrue(self.log.call_count == 1)

        # HERE IS WHAT I MEAN: at this point if check_call has also been called with other calls,
        # it will pass because you don't do (self.assertTrue(self.check_call.call_count == 3)

        self.check_call.reset_mock()
        ovs.del_bridge_port('test', 'eth1', linkdown=False)
        self.check_call.assert_called_once_with(
            ['ovs-vsctl', '--', '--if-exists', 'del-port', 'test', 'eth1'])
```

PR #497, opened 2020-07-28
contrib/ceph: Add BlueStore Compression support

fnordahl:
Please review this commit by commit, and please **DO NOT squash on merge**!

### DIFF
--- a/tests/contrib/storage/test_linux_ceph.py
+++ b/tests/contrib/storage/test_linux_ceph.py
@@ -1968,3 +1968,33 @@ class CephUtilsTests(TestCase):
         op = base_op.copy()
         op['max-objects'] = 42
         self.assertEqual(rq.ops, [op])
+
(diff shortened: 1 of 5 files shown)


tests/contrib/storage/test_linux_ceph.py
```diff
@@ -584,6 +584,8 @@ def test_erasure_pool_create(self, get_osds, erasure_profile):
             call(['ceph', '--id', 'admin', 'osd', 'pool', 'create',
                   '--pg-num-min=32', 'test',
                   '2048', '2048', 'erasure', 'default']),
+            call(['ceph', '--id', 'admin', 'osd', 'pool',
+                  'set', 'test', 'target_size_ratio', '1.0']),
```
fnordahl, line 588: Note that this was called after creation of erasure coded pools before the change too, but the order has now changed which means we need to include it in the check.

PR #498, opened 2020-07-29
Add support for Erasure Coded pools

javacruft:
Add support in the ceph module for creation of new erasure coding
profiles and support for setting the allow_ec_overwrites property
on erasure coded pools.  This is required to support RBD use
cases.

Update OpenStack Ceph context to detect use of EC pool type,
providing an appropriate value for the rbd default data pool
for use in ceph.conf.

This last part is a bit of a workaround due to the fact that
OpenStack services don't have native configuration options to
set the metadata and data pool options when creating new RBD
images.

Add support for configuration and validation of the erasure
plugin technique as part of the erasure profile creation.

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -58,6 +58,7 @@ from charmhelpers.core.hookenv import (
     status_set,
     network_get_primary_address,
     WARNING,
+    service_name,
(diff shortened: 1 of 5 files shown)


javacruft:
@fnordahl @gnuoy please can you take a look at this PR.

Issue #500, opened 2020-08-05
cannot use charmhelpers.core.hookenv on python2 due to addition of enum import


afreiberger:
After refreshing charmhelpers on a python2 charm, I ran into issues during import of hookenv and host due to bare import of enum (available in python3.4+)

```
...
install Setting up python-six (1.14.0-2) ...
install Traceback (most recent call last):
install   File "hooks/hooks.py", line 14, in <module>
install     from charmhelpers.core.host import (
install   File "/var/lib/juju/agents/unit-rsyslog-forwarder-ha-0/charm/hooks/charmhelpers/core/host.py", line 37, in <module>
install     from .hookenv import log, INFO, DEBUG, local_unit, charm_name
install   File "/var/lib/juju/agents/unit-rsyslog-forwarder-ha-0/charm/hooks/charmhelpers/core/hookenv.py", line 24, in <module>
install     from enum import Enum
install ImportError: No module named enum
```

I've updated my code to run on python3 to workaround this.  Is there a requirement for python3 for v0.20.16?

Is there a stable charm-helpers tagged release that supports py2?

afreiberger:
introduced by PR #456

ajkavanagh:
This is a bit odd.  I just did this in py2 on a bionic host:

```
Python 2.7.18rc1 (default, Apr  7 2020, 12:05:55)
Type "copyright", "credits" or "license" for more information.

IPython 5.10.0 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: import charmhelpers.core.hookenv

In [2]: ls
bin/           debian/  HACKING.md  Makefile     py2venv/    requirements.txt  setup.cfg  tarmac_tests.sh*       tests/  tox.ini
charmhelpers/  docs/    LICENSE     MANIFEST.in  README.rst  scripts/          setup.py   test-requirements.txt  tools/

In [3]: import enum

In [4]:
```

Is it an old py2 version?

afreiberger:
I don't still have my artifacts, but it may have been xenial and/or trusty.  I ran into it during pretty standard deployment during zaza testing of lp:charm-rsyslog-forwarder-ha.  
https://git.launchpad.net/~afreiberger/charm-rsyslog-forwarder-ha/commit/?id=ad9186b5c9fcce2d386d2459da6d03684330f8a7

[...]

PR #504, opened 2020-08-12
Python2.7 doesn't support Enum (natively) - code around it

ajkavanagh:
As the charm-helpers library still needsd to support Python2.7 charms,
this patchset reverses a recent commit that introduces the Enum class.
Despite this being a great approach, the changes in the patchset are
more compatible with the range of charms that it needs to support.

Closes-Bug: #500

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -21,9 +21,8 @@
 from __future__ import print_function
 import copy
 from distutils.version import LooseVersion
-from enum import Enum
 from functools import wraps
-from collections import namedtuple
+from collections import (namedtuple, OrderedDict)
 import glob
 import os
 import json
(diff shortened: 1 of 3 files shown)


cjwatson:
Did you consider adding something like `enum34; python_version < "3.4"` to `requirements.txt` instead?  That might be a lot simpler; it should be possible to leave the rest of the code as it is if you take that approach.

ajkavanagh:
> Did you consider adding something like `enum34; python_version < "3.4"` to `requirements.txt` instead? That might be a lot simpler; it should be possible to leave the rest of the code as it is if you take that approach.

It's definitely an interesting idea.  However, I'm not sure it would work, as charm-helpers is a lib (rather than app), and so is there a way to signal to the app that it needs an extra module.  Normally this would be via the setup.py `install_requires=`.  I don't know whether `pbr` (which charm-helpers uses in setup.py) picks up requirements.txt for its dependencies, nor whether the "; python_version ..." trick will work with it?  Do you know?

I must admit that I was led down this solution as 'classic' charms vendored in charm-helpers using a sync script which would mean that py27 charms (if there are any left!?) would also need to apt install "python-enum34", and I was seeking to avoid this.

afreiberger:
+1 that requirements.txt workaround won't work for solving this for classic charms trying to keep up with new charmhelpers revisions for things such as added contrib.nrpe library hook updates for max_check_attempts.

That being said, those charms, if possible, should try to port to py3 for hooks as long as they don't have to continue support for trusty.

ajkavanagh:
> +1 that requirements.txt workaround won't work for solving this for classic charms trying to keep up with new charmhelpers revisions for things such as added contrib.nrpe library hook updates for max_check_attempts.
> 
> That being said, those charms, if possible, should try to port to py3 for hooks as long as they don't have to continue support for trusty.

There are definitely no OpenStack py27 classic charms, but the charm-helpers lib is still py27 compatible.  I've been pondering for a while on doing a major version bump and dropping py27 support from charm-helpers.  Thoughts?

We could then create branches for py27, py35 (and soon, py36) as legacy support as those versions are EOL in upstream land.

ajkavanagh:
As charm-helpers is now a py3 only lib as of #651, I'm abandoning/closing this  PR.

PR #509, opened 2020-09-03
ovs: Fix error in patch_ports_on_bridge

fnordahl:
The unit test for patch_ports_on_bridge was not quite good enough
and hid the fact that the function now always raises `ValueError`.

This patch fixes both.


ajkavanagh (approved):
LGTM

PR #513, opened 2020-09-28
Make EC profiles immutable

javacruft:
Changing and existing ec profile using the --force option can have
nasty side effects such as causing OSD's to crash.

If an erasure profile already exists, don't try to update it.

Actions on the ceph-mon charm can be used to create new pools
and profiles, migrate data and then rename existing pools.

Closes-Bug: 1897517


gnuoy (approved):
LGTM, thanks

PR #516, opened 2020-10-06
core/host: Speedup chownr by avoiding a couple of syscalls

pponnuvel:
`chownr` function calls `os.path.exists` and `os.path.lexists` for every file. This can be avoided by checking the error
case that they were meant to handle (i.e. the symlink's target is missing).

As noted in [1898605](https://bugs.launchpad.net/charm-helpers/+bug/1898605), this is particularly noticeable in ceph-osd charms.

Closes-bug: #1898605

Signed-off-by: Ponnuvel Palaniyappan <pponnuvel@gmail.com>

### DIFF
--- a/charmhelpers/core/host.py
+++ b/charmhelpers/core/host.py
@@ -924,9 +924,10 @@ def chownr(path, owner, group, follow_links=True, chowntopdir=False):
     for root, dirs, files in os.walk(path, followlinks=follow_links):
         for name in dirs + files:
             full = os.path.join(root, name)
-            broken_symlink = os.path.lexists(full) and not os.path.exists(full)
-            if not broken_symlink:
+            try:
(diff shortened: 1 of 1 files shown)


pponnuvel:
Ping

ajkavanagh (changes requested):
Unfortunately, charm-helpers still needs to support Py2.7 (until the code is removed for the support) and this will break on 2.7.  (okay from 3.5 onwards).  You could re-do it as per: https://stackoverflow.com/questions/26745283/how-do-i-import-filenotfounderror-from-python-3 to be compatible with both py2.7 and py3

pponnuvel:
Hi @ajkavanagh,
Thanks for the review. Please review https://github.com/juju/charm-helpers/pull/527 where the requested change is done.

PR #520, opened 2020-10-16
Guarantee all SAN addresses are symlinked to certs

thedac:
Prior to this change the algorithm used for requesting certificates was
different from the one used to set up symlinks. This change does the
following:

* Breaks out the SAN IP address gathering into get_certificate_sans
* Changes variable names to better describe what they are for
* Uses the request object to determine which IPs map to which
certificates
* Begins the process of fixing the "int" vs "internal" binding name

Closes-Bug: #1893847

### DIFF
--- a/charmhelpers/contrib/openstack/cert_utils.py
+++ b/charmhelpers/contrib/openstack/cert_utils.py
@@ -34,12 +34,14 @@ from charmhelpers.core.hookenv import (
     WARNING,
 )
 from charmhelpers.contrib.openstack.ip import (
-    ADMIN,
     resolve_address,
     get_vip_in_network,
-    INTERNAL,
-    PUBLIC,
-    ADDRESS_MAP)
+    ADDRESS_MAP,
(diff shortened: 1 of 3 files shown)


charmhelpers/contrib/openstack/cert_utils.py
```diff
@@ -113,63 +115,117 @@ def get_request(self):
         return req
 
 
-def get_certificate_request(json_encode=True):
+def get_certificate_request(json_encode=True, bindings=None):
     """Generate a certificatee requests based on the network confioguration
```
gnuoy, line 120: Docstring for 'bindings'

gnuoy (commented):
One tiny nit, happy to land after that. Thanks for this.

charmhelpers/contrib/openstack/cert_utils.py
thedac: Fixed

gnuoy (approved):

PR #521, opened 2020-10-20
Update haproxy.cfg template for tls backends

xavpaice:
Adds a switch so that if the context for haproxy config includes
backend_options['check_ssl'] then haproxy will run http checks via
https.


ajkavanagh:
Associated charm change (for glance): https://review.opendev.org/c/openstack/charm-glance/+/750224

PR #522, opened 2020-10-27
Correct name of CA cert file

gnuoy:
In the past a CA could be supplied to a charm by one of three different methods:

1. From keystone

1. From ssl* config

1. From vault

Regardless of where the CA came from it was written to `keystone_juju_ca_cert.pem`. The first method has since been removed but the the name of the CA file has remained.

The purpose of this PR is to provide a tactical fix to allow a CA to be supplied via ssl_ca config option and by vault (the certificates management code still needs a complete overhaul). This is achieved by writing out the CA cert received via the certificates relation to a new name. The old file named  `keystone_juju_ca_cert.pem` cannot simply be removed because it may contain a certificate that was supplied via the ssl_ca config option and not from vault. To safely manage this situation the process_certificates method will now ensure that the CA cert from the  ssl_ca config option is written out if ssl_ca is set or the file removed if it is unset.

### DIFF
--- a/charmhelpers/contrib/hahelpers/apache.py
+++ b/charmhelpers/contrib/hahelpers/apache.py
@@ -34,7 +34,9 @@ from charmhelpers.core.hookenv import (
     INFO,
 )
 
-KEYSTONE_CA_CERT_FILE = 'keystone_juju_ca_cert'
+# This file contains the CA cert from the charms ssl_ca configuration
+# option, in future the file name should be updated reflect that.
+CONFIG_CA_CERT_FILE = 'keystone_juju_ca_cert'
 
 
 def get_cert(cn=None):
@@ -85,4 +87,4 @@ def retrieve_ca_cert(cert_file):
 
 
 def install_ca_cert(ca_cert):
-    host.install_ca_cert(ca_cert, KEYSTONE_CA_CERT_FILE)
+    host.install_ca_cert(ca_cert, CONFIG_CA_CERT_FILE)
--- a/charmhelpers/contrib/hahelpers/apache.py
+++ b/charmhelpers/contrib/hahelpers/apache.py
@@ -34,6 +34,8 @@ from charmhelpers.core.hookenv import (
     INFO,
 )
 
+KEYSTONE_CA_CERT_FILE = 'keystone_juju_ca_cert'
+
 
 def get_cert(cn=None):
     # TODO: deal with multiple https endpoints via charm config
@@ -83,4 +85,4 @@ def retrieve_ca_cert(cert_file):
 
 
 def install_ca_cert(ca_cert):
-    host.install_ca_cert(ca_cert, 'keystone_juju_ca_cert')
+    host.install_ca_cert(ca_cert, KEYSTONE_CA_CERT_FILE)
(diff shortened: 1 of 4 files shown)


AurelienLourot (changes requested):
LGTM but what happens with the old file? Does it linger?

gnuoy:
> LGTM but what happens with the old file? Does it linger?

The old file is not touched

AurelienLourot:
Shouldn't we go the extra mile and delete the old file? Any reasons not to?

AurelienLourot (changes requested):
So with your last commit you just went back to using the old name. The intent of the PR was to fix that name. Am I missing something?

gnuoy:
> So with your last commit you just went back to using the old name. The intent of the PR was to fix that name. Am I missing something?

I'm not sure where you are seeing that it is going back to using the old name ?

I originally omitted any attempt to clear up the certificate file using the old name because the file may either contain a certificate derived from the ssl_ca config option or from a certificate supplied from vault. In addition the code that does the installation of certs from the ssl* config options is in a completely different place (side-effect of an apache context). In the end I came around to your pov that the file should be cleaned up if possible but to do that we need to check if ssl_ca is set and to write to the old location if it is, if it is not then remove it.

charmhelpers/contrib/hahelpers/apache.py
```diff
@@ -34,6 +34,8 @@
     INFO,
 )
 
+KEYSTONE_CA_CERT_FILE = 'keystone_juju_ca_cert'
```
AurelienLourot, line 37: The goal of this PR is to stop making use of this filename, so should this be named `OLD_KEYSTONE_CA_CERT_FILE`? 
Also why is this used in `install_ca_cert()`. I'm sure it wasn't like that in earlier version of this PR but because you push-force (it's fine, it's your right :) ) I can't really show evidence of that. But this is basically the source of my confusion: I know see a PR that creates a file named `keystone_juju_ca_cert` although the PR description says it wants to move away from that name. What am I missing?

AurelienLourot: basically I expect this PR to show me some variation of

```
install_ca_cert(<the new name>)
remove_cert(<the old name>)
```

but I'm confused because on line 88 of this file I see `install_ca_cert(<the old name>)`

gnuoy: > The goal of this PR is to stop making use of this filename, so should this be named `OLD_KEYSTONE_CA_CERT_FILE`?
> Also why is this used in `install_ca_cert()`. I'm sure it wasn't like that in earlier version of this PR but because you push-force (it's fine, it's your right :) ) I can't really show evidence of that. But this is basically the source of my confusion: I now see a PR that creates a file named `keystone_juju_ca_cert` although the PR description says it wants to move away from that name. What am I missing?

The push-force was after a rebase, the diff is unchanged

gnuoy: I have updated the commit message which I hope is now a but clearer.

AurelienLourot (approved):
It's much clearer to me now, thanks!

PR #524, opened 2020-11-02
Fix race in checking closed ports on pause

### LINKED ISSUE #523: Racey service pause
When running functional tests I'm seeing intermittent faults of the pause action: `Action 'pause' exited with status 'failed': 'Couldn't pause: Services should be paused but these ports which should be closed, but are open: 4990, 35347, Unit paused.'`. With the keystone charm functests I'd get this about 2 of 5 times.

This message originates in [utils.check_actually_paused()](https://github.com/juju/charm-helpers/blob/master/charmhelpers/contrib/openstack/utils.py#L1460) called via the assessor function created by [make_assess_status_func()](https://github.com/juju/charm-helpers/blob/21f9f239c4f977afb1720243a6f6de4ae876986c/charmhelpers/contrib/openstack/utils.py#L1652) used to verify if a pause action was successful. check_actually_paused() probes ports by [calling `nc -z 0.0.0.0 port`](https://github.com/juju/charm-helpers/blob/21f9f239c4f977afb1720243a6f6de4ae876986c/charmhelpers/contrib/network/ip.py#L552) and checking the returncode.

I believe this is prone to races. The service call to shutdown the service returns before the service shutdown is actually complete. In a loaded system (typical for functests) the time it takes to shutdown the service can be longer than the time it takes to reach the `nc` probe. I've experimentally added some more logging and a retry to the probe function:

```
diff --git a/charmhelpers/contrib/network/ip.py b/charmhelpers/contrib/network/ip.py
index b13277b..b0ef636 100644
--- a/charmhelpers/contrib/network/ip.py
+++ b/charmhelpers/contrib/network/ip.py
@@ -550,7 +550,13 @@ def port_has_listener(address, port):
     Note calls 'zc' via a subprocess shell
     """
     cmd = ['nc', '-z', address, str(port)]
+    netstat = subprocess.check_output("netstat -tnlp", shell=True)
     result = subprocess.call(cmd)
+    log("Checking for listener on  {}:{} yielded {}; netstat: {}".format(address, port, result, netstat))
+    import time; time.sleep(3)
+    netstat = subprocess.check_output("netstat -tnlp", shell=True)
+    result = subprocess.call(cmd)
+    log("Checking AGAIN for listener on  {}:{} yielded {}; netstat: {}".format(address, port, result, netstat))
     return not(bool(result))
``` 

Note the service being shutdown for these is apache2. With the above I'd indeed get cases where apache2 was still listening on the first probe, but would be gone on the second. The systemd unit file actually being run for apache2 is being generated from the resp. SysV `/etc/init.d/apache2` script, which afaict signals apache2 to stop but doesn't block.

To fix this race, check_actually_paused() could do a retry as well

sabaini:
When pausing a service a check is called to verify if services are
stopped and ports are closed. This is prone to racing as service
shutdown is not guaranteed to be synchronous.

This change implements a retry decorator and uses it to retry the
verification function if it's not successful

fixes #523

### DIFF
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -18,6 +18,7 @@ from functools import wraps
 
 import subprocess
 import json
+import operator
 import os
 import sys
 import re
@@ -33,7 +34,7 @@ from charmhelpers import deprecate
 
 from charmhelpers.contrib.network import ip
 
-from charmhelpers.core import unitdata
+from charmhelpers.core import decorators, unitdata
 
 from charmhelpers.core.hookenv import (
     WORKLOAD_STATES,
@@ -1295,7 +1296,7 @@ def _check_listening_on_ports_list(ports):
     Returns a list of ports being listened to and a list of the
     booleans.
 
-    @param ports: LIST or port numbers.
+    @param ports: LIST of port numbers.
     @returns [(port_num, boolean), ...], [boolean]
     """
     ports_open = [port_has_listener('0.0.0.0', p) for p in ports]
@@ -1564,6 +1565,21 @@ def manage_payload_services(action, services=None, charm_func=None):
     return success, messages
 
 
+def make_wait_for_ports_barrier(ports, retry_count=5):
+    """Make a function to wait for port shutdowns.
+
+    Create a function which closes over the provided ports. The function will
+    retry probing ports until they are closed or the retry count has been reached.
+
+    """
+    @decorators.retry_on_predicate(retry_count, operator.not_, base_delay=0.1)
+    def retry_port_check():
+        _, ports_states = _check_listening_on_ports_list(ports)
+        juju_log("Probe ports {}, result: {}".format(ports, ports_states), level="DEBUG")
+        return any(ports_states)
+    return retry_port_check
+
+
 def pause_unit(assess_status_func, services=None, ports=None,
                charm_func=None):
     """Pause a unit by stopping the services and setting 'unit-paused'
@@ -1599,6 +1615,7 @@ def pause_unit(assess_status_func, services=None, ports=None,
         services=services,
         charm_func=charm_func)
     set_unit_paused()
+
     if assess_status_func:
         message = assess_status_func()
         if message:
--- a/charmhelpers/core/decorators.py
+++ b/charmhelpers/core/decorators.py
@@ -53,3 +53,41 @@ def retry_on_exception(num_retries, base_delay=0, exc_type=Exception):
         return _retry_on_exception_inner_2
 
     return _retry_on_exception_inner_1
+
+
+def retry_on_predicate(num_retries, predicate_fun, base_delay=0):
+    """Retry based on return value
+
+    The return value of the decorated function is passed to the given predicate_fun. If the
+    result of the predicate is False, retry the decorated function up to num_retries times
+
+    An exponential backoff up to base_delay^num_retries seconds can be introduced by setting
+    base_delay to a nonzero value. The default is to run with a zero (i.e. no) delay
+
+    :param num_retries: Max. number of retries to perform
+    :type num_retries: int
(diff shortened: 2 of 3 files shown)


charmhelpers/core/decorators.py
```diff
@@ -53,3 +53,34 @@ def _retry_on_exception_inner_2(*args, **kwargs):
...
+    def _retry_on_pred_inner_1(f):
+        def _retry_on_pred_inner_2(*args, **kwargs):
+            retries = num_retries
+            multiplier = 1
+            delay = base_delay
+            while True:
+                result = f(*args, **kwargs)
+                if predicate_fun(result) or retries <= 0:
+                    return result
+                delay *= multiplier
```
ajkavanagh, line 76: delay = 0 (default), so multiplying it will result in 0.  Perhaps base_delay should default to > 0?  Obviously, if this is the intended feature (which I think it is) then it needs to be declared in the docstring.

charmhelpers/core/decorators.py
```diff
@@ -53,3 +53,34 @@ def _retry_on_exception_inner_2(*args, **kwargs):
...
+
+def retry_on_predicate(num_retries, predicate_fun, base_delay=0):
+    """Retry based on return value
+
+    The return value of the decorated function is passed to the given predicate_fun. If the
+    result of the predicate is False, retry the decorated function up to num_retries times
+
+    An exponential backoff up to base_delay^num_retries seconds can be introduced by setting
+    base_delay to a nonzero value
+    """
```
ajkavanagh, line 66: Need do describe the parameters:

:param num_retries: ..
:type num_retries: int

etc.

ajkavanagh (changes requested):
I'm not sure we want to do this.  This helper is used by access_status (IIRC) and thus will do 'lots of work' many times if a service is down.  If there is a race, then the race should be handled at the calling code location if the pausing code needs to determine if the service is down with retries.  i.e. call this in a loop, not make every caller of this code get retries.

sabaini:
> I'm not sure we want to do this. This helper is used by access_status (IIRC) and thus will do 'lots of work' many times if a service is down. If there is a race, then the race should be handled at the calling code location if the pausing code needs to determine if the service is down with retries. i.e. call this in a loop, not make every caller of this code get retries.

Hey @ajkavanagh, thanks for the review!

You're right, the make_assess_status_func() generated func has a code path that also would include check_actually_paused() which I've added the retry-decorator to, and I guess we really don't want to delay the update-status hook runs. Although the retry would only be hit for units which are paused but have an open port/service up that shouldn't be, still could be avoided altogether by moving it to the pause code i.e. in [service_pause()](https://github.com/juju/charm-helpers/blob/21f9f239c4f977afb1720243a6f6de4ae876986c/charmhelpers/core/host.py#L176)

I'll update this PR

charmhelpers/contrib/openstack/utils.py
```diff
@@ -1599,6 +1599,14 @@ def pause_unit(assess_status_func, services=None, ports=None,
...
         charm_func=charm_func)
     set_unit_paused()
+
+    # construct a check function, making it retry with a delay if ports are open
+    @decorators.retry_on_predicate(5, bool, base_delay=0.1)
+    def retry_port_check(ports):
+        _, ports_states = _check_listening_on_ports_list(ports)
+        return not any(ports_states)
+    if ports:
+        retry_port_check(ports)  # block until check is successful or times out
```
ajkavanagh, line 1609: I'm confused; it looks like ports wasn't used before (but still existed) and now does.  So AFAICT it waits for a bit checking the port, and if it's still not closed then it just moves on?  No exception is raised?  This function is now very confusing.  Shouldn't "manage_payload_services" actually do the job of checking that the ports are close ... i.e. it's supposed to manage the payload service?  I think this is blurring what this function is supposed to do?

ajkavanagh (approved):
LGTM

ajkavanagh (approved):
LGTM

PR #525, opened 2020-11-07
get_hostname(): handle exception on no answer from DNS server

mfoliveira:
The function `get_hostname()` raises an exception vs. just `return None`
if DNS reverse query by IP address fails (eg, DNS server fail/no answer.)

In this case, it does not reach `socket.gethostbyaddr()`, which could be
used for a workaround with `/etc/hosts` (eg, while fixing DNS server.)

In Xenial `/etc/hosts` is not taken into account because of the exception.

In Bionic `/etc/hosts` is taken into account before the exception, but if
the hostname/IP is not there, that exception does happen anyway.

So, add a `try/except` block to catch the exception so to try `/etc/hosts`
in Xenial and silence the exception in Bionic / `return None` as planned.

Fixes LP: #1903414

Example:
---

In this trace, the `install` hook of `rabbitmq-server` failed:

```
	2020-11-07 10:15:55 DEBUG install Traceback (most recent call last):
	2020-11-07 10:15:55 DEBUG install   File "/var/lib/juju/agents/unit-rabbitmq-server-8/charm/hooks/install.real", line 989, in <module>
	2020-11-07 10:15:55 DEBUG install     rabbit.assess_status(rabbit.ConfigRenderer(rabbit.CONFIG_FILES))
	2020-11-07 10:15:55 DEBUG install   File "/var/lib/juju/agents/unit-rabbitmq-server-8/charm/hooks/rabbit_utils.py", line 150, in __init__
	2020-11-07 10:15:55 DEBUG install     ctxt.update(svc_context())
	2020-11-07 10:15:55 DEBUG install   File "/var/lib/juju/agents/unit-rabbitmq-server-8/charm/hooks/rabbitmq_context.py", line 137, in __call__
	2020-11-07 10:15:55 DEBUG install     ssl_mode, external_ca = ssl_utils.get_ssl_mode()
	2020-11-07 10:15:55 DEBUG install   File "/var/lib/juju/agents/unit-rabbitmq-server-8/charm/hooks/ssl_utils.py", line 61, in get_ssl_mode
	2020-11-07 10:15:55 DEBUG install     relation_certs = get_relation_cert_data()
	2020-11-07 10:15:55 DEBUG install   File "/var/lib/juju/agents/unit-rabbitmq-server-8/charm/hooks/ssl_utils.py", line 56, in get_relation_cert_data
	2020-11-07 10:15:55 DEBUG install     _, hostname = get_unit_amqp_endpoint_data()
	2020-11-07 10:15:55 DEBUG install   File "/var/lib/juju/agents/unit-rabbitmq-server-8/charm/hooks/ssl_utils.py", line 47, in get_unit_amqp_endpoint_data
	2020-11-07 10:15:55 DEBUG install     return ip, get_hostname(ip)
	2020-11-07 10:15:55 DEBUG install   File "/var/lib/juju/agents/unit-rabbitmq-server-8/charm/charmhelpers/contrib/network/ip.py", line 522, in get_hostname
	2020-11-07 10:15:55 DEBUG install     result = ns_query(rev)
	2020-11-07 10:15:55 DEBUG install   File "/var/lib/juju/agents/unit-rabbitmq-server-8/charm/charmhelpers/contrib/network/ip.py", line 478, in ns_query
	2020-11-07 10:15:55 DEBUG install     answers = dns.resolver.query(address, rtype)
	2020-11-07 10:15:55 DEBUG install   File "/usr/lib/python3/dist-packages/dns/resolver.py", line 979, in query
	2020-11-07 10:15:55 DEBUG install     raise_on_no_answer, source_port)
	2020-11-07 10:15:55 DEBUG install   File "/usr/lib/python3/dist-packages/dns/resolver.py", line 821, in query
	2020-11-07 10:15:55 DEBUG install     raise NoNameservers
	2020-11-07 10:15:55 DEBUG install dns.resolver.NoNameservers
	2020-11-07 10:15:55 ERROR juju.worker.uniter.operation runhook.go:132 hook "install" failed: exit status 1
```

Steps to reproduce/verify:
---

Deploy in Xenial/Bionic, SSH in, and run `python3` / import `get_hostname()`:
```
	$ juju deploy rabbitmq-server rmq --series xenial # or bionic

	$ juju status | grep rmq
	rmq  3.5.7    active      1  rabbitmq-server  jujucharms  106  ubuntu
	rmq/0*  active    idle   3        10.191.59.36    5672/tcp  Unit is ready

	$ juju ssh rmq/0

	$ python3
	>>> import sys
	>>> sys.path.append('/var/lib/juju/agents/unit-rmq-0/charm/')
	>>> from charmhelpers.contrib.network.ip import get_hostname
```
Xenial / original:
```
	>>> get_hostname('10.10.10.10')
	  File "/var/lib/juju/agents/unit-rmq-3/charm/charmhelpers/contrib/network/ip.py", line 522, in get_hostname
	    result = ns_query(rev)
	  File "/var/lib/juju/agents/unit-rmq-3/charm/charmhelpers/contrib/network/ip.py", line 478, in ns_query
	    answers = dns.resolver.query(address, rtype)
	  File "/usr/lib/python3/dist-packages/dns/resolver.py", line 979, in query
	    raise_on_no_answer, source_port)
	  File "/usr/lib/python3/dist-packages/dns/resolver.py", line 821, in query
	    raise NoNameservers
	dns.resolver.NoNameservers

	>>> ^Z
	[1]+  Stopped                 python3

	$ echo '10.10.10.10 test-hostname' | sudo tee -a /etc/hosts
	10.10.10.10 test-hostname

	$ fg
	python3

	>>> get_hostname('10.10.10.10')
	  File "/var/lib/juju/agents/unit-rmq-3/charm/charmhelpers/contrib/network/ip.py", line 522, in get_hostname
	    result = ns_query(rev)
	  File "/var/lib/juju/agents/unit-rmq-3/charm/charmhelpers/contrib/network/ip.py", line 478, in ns_query
	    answers = dns.resolver.query(address, rtype)
	  File "/usr/lib/python3/dist-packages/dns/resolver.py", line 979, in query
	    raise_on_no_answer, source_port)
	  File "/usr/lib/python3/dist-packages/dns/resolver.py", line 821, in query
	    raise NoNameservers
	dns.resolver.NoNameservers

	>>> quit()
```
Xenial / patched:
```
	$ sudo sed -i '/test-hostname/d' /etc/hosts  # remove entry

	>>> get_hostname('10.10.10.10')
	>>>

	>>> ^Z
	[1]+  Stopped                 python3

	$ echo '10.10.10.10 test-hostname' | sudo tee -a /etc/hosts
	10.10.10.10 test-hostname

	$ fg
	python3

	>>> get_hostname('10.10.10.10')
	'test-hostname'

	>>> quit()
```
Bionic / original:
```
	>>> get_hostname('10.10.10.10')
	  File "/var/lib/juju/agents/unit-rmq-0/charm/charmhelpers/contrib/network/ip.py", line 522, in get_hostname
	    result = ns_query(rev)
	  File "/var/lib/juju/agents/unit-rmq-0/charm/charmhelpers/contrib/network/ip.py", line 478, in ns_query
	    answers = dns.resolver.query(address, rtype)
	  File "/usr/lib/python3/dist-packages/dns/resolver.py", line 1132, in query
	    raise_on_no_answer, source_port)
	  File "/usr/lib/python3/dist-packages/dns/resolver.py", line 947, in query
	    raise NoNameservers(request=request, errors=errors)
	dns.resolver.NoNameservers: All nameservers failed to answer the query 10.10.10.10.in-addr.arpa. IN PTR: Server 127.0.0.53 UDP port 53 answered SERVFAIL

	>>> ^Z
	[1]+  Stopped                 python3

	$ echo '10.10.10.10 test-hostname' | sudo tee -a /etc/hosts
	10.10.10.10 test-hostname

	$ fg
	python3

	>>> get_hostname('10.10.10.10')
	'test-hostname'
```
Bionic / patched:
```
	$ sudo sed -i '/test-hostname/d' /etc/hosts  # remove entry

	>>> get_hostname('10.10.10.10')
	>>>

	[1]+  Stopped                 python3

	$ echo '10.10.10.10 test-hostname' | sudo tee -a /etc/hosts
	10.10.10.10 test-hostname

	$ fg
	python3

	>>> get_hostname('10.10.10.10')
	'test-hostname'

	>>> quit()
```

Signed-off-by: Mauricio Faria de Oliveira `<mfo@canonical.com>`

### DIFF
--- a/charmhelpers/contrib/network/ip.py
+++ b/charmhelpers/contrib/network/ip.py
@@ -519,7 +519,10 @@ def get_hostname(address, fqdn=True):
             import dns.reversename
 
         rev = dns.reversename.from_address(address)
-        result = ns_query(rev)
+        try:
+            result = ns_query(rev)
+        except:
+            result = None
 
         if not result:
             try:


ajkavanagh (changes requested):
Unfortunately, I don't think this is the right approach.  charm-helpers is a library, and this is changing the nature of the library function which currently raises an Exception if there is no answer, and instead will now return None.  All the charms that use this function expect (or rely on) an exception being generated if the function fails, rather than returning None.  This could make charms fail in interesting ways as they would now get a value (None) and try to do something with it, rather than the exception either being handled or stopping execution.

It would be better to handle the error in the calling function to wrap `get_hostname()` in an exception handler rather than changing the API of the library.

mfoliveira:
Hi Alex,

Thanks for the quick and detailed review.

I see your point; and looking back, `ns_query()` had a general `try/except` introduced in Oct 2016 commit 6b0675b then commit e64b920  particularized it ~2 weeks later to `dns.resolver.NXDOMAIN` only -- thus all old behavior to change, indeed.

However, currently, the fallback to `socket.gethostbyaddr/name()` in `get_hostname()` and `get_host_ip()` (which allows for `/etc/hosts`) is then not available/exercised at all, for other exceptions (e.g., `NoAnswer`.)

I'm not sure that's intented (i.e., prioritize DNS-only exception over attempt with `/etc/hosts`), since this piece of code is mentioned as "try as last resort" in commit 791dd5e (which predates the other two commits above), and also because there's no way to tell the functions to do it either way.

So, do you think a compromise solution would be both? e.g., add logic to `try/except` the `ns_query()` call, and if throws an exception, try and call `socket.gethostbyname/addr()`, and if this doesn't have any result, then throw the original exception from `ns_query()`.

That should address the inconsistency between Xenial and Bionic, seen above.
In Bionic, `/etc/hosts` _is_ considered before throwing an exception; likely due to DNS differences (`systemd-resolved`, perhaps), and in Xenial, the exception happens first.

This would seems more applicable to a library, I guess.  
Please let me know your opinion/next steps. :)
Thanks again!

mfoliveira:
Hi @ajkavanagh 

Just following up on this -- do you think we should try to fix the inconsistency between Xenial/Bionic?

(i.e., Bionic uses `/etc/hosts` implicity in the DNS query likely bcz of `systemd-resoved` before trying `socket.gethostbyname/addr()`, while Xenial raises the exception in the DNS query/has no `systemd-resolved`.)

cheers,
Mauricio

ajkavanagh:
Hi @mfoliveira 

Sorry for taking such a while to get back to this.  Having taken another look at this function, based on your observations, it's really quite broken isn't it!  It's great that you are looking into and coming up with a solution.  So w.r.t. your questions raised:

> So, do you think a compromise solution would be both? e.g., add logic to try/except the ns_query() call, and if throws an exception, try and call socket.gethostbyname/addr(), and if this doesn't have any result, then throw the original exception from ns_query().

Yes, this is probably a good idea; i.e. it will raise the exception if things go badly.  It looks like depending code expects to see this.  It will also "fix" the issue of it raising an exception or returning `None` depending on which bit of the lookup is failing.  So, that's a good thing.

> Just following up on this -- do you think we should try to fix the inconsistency between Xenial/Bionic?
>
> (i.e., Bionic uses /etc/hosts implicity in the DNS query likely bcz of systemd-resoved before trying socket.gethostbyname/addr(), while Xenial raises the exception in the DNS query/has no systemd-resolved.)

Yes, if you can in a logically consistent way, it's probably a good idea to do that.

Thanks again for looking into this; I'd missed the brokenness in the function (raising an Exception or returning None) that had been introduced over time.  It'd be good to get some consistency back.

Thanks
Alex.

mfoliveira:
Hi @ajkavanagh 

Thanks for your review comments; no worries w.r.t. delay. :)

So, yes, it's possible to fix the inconsistency between Trusty/Xenial and Bionic/Focal about `/etc/hosts` in a logically consistent way.
Since this is somewhat different from the original approach in this PR, I'll send another one and close this, so we keep history.

Please see PR #555.

Note, the `/etc/hosts` inconsistency is unrelated to the `return None` issue.
The former happens on the call to `ns_query()` (it raises an exception or not depending on the distro release).
The latter happens _if_ `ns_query()` returns None (i.e., in case of `NXDOMAIN`) and _if_ `socket.gethostbyaddr/name()` fails too.

Fixing the former doesn't incur behavior changes, so I've gone ahead and did it in #555.

But fixing the latter _is_ a behavior change, so I'll need your blessing on that :)
Technically it's a very simple change on top of #555, which I can push there, depending on your input.
(and if it possibly causes problems, that's a simpler revert; which doesn't pull out the other improvement.)

The thing w/ that behavior change is that callers that maybe didn't handle exceptions, and were happy w/ getting `return None` for the case they got a non-existing domain,  would now get exceptions.

I'm not very familiar w/ many charms, so can't speak to the standard usage pattern of `get_hostname()` and `get_host_ip()` -- so I'd have to ask for your input on whether to do that other change or not.

Thanks for all your help on this!

PR #526, opened 2020-11-10
Updates for the Cinder Ceph Replication spec

oprinmarius:
These changes are part of the implementation of the [Cinder Ceph Replication](https://opendev.org/openstack/charm-specs/src/commit/40367522bf03cb18cc4d1c0d055aab6065f41fcc/specs/wallaby/backlog/cinder-ceph-replication.rst) spec:

- Update the broker request handler to take into consideration the new rbd-mirroring-mode flag.
   Add 'rbd-mirroring-mode' as part of the 'create-pool' operation 
   This value is going to be parsed by the ceph-rbd-mirror charm when enabling pool mirroring.
   It defaults to 'pool', since this was the only mode handled previously by the ceph-rbd-mirror charm.

- Update CephContext to ignore advertised default RBD features from the relation with ceph-mon, if the client charm implements the new 'rbd-mirroring-mode' flag, and that's set to 'image'.

- Add Ceph relation name parameter to the CephContext constructor. The current implementation assumes that the clients will name their relation 'ceph'. This will remove the hard-coded value, and make it the default parameter value. Thus, we will maintain the backwards compatibility.
This is going to be useful for charms that implement the ceph-client interface, but the relation name is not ceph.

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -788,13 +788,14 @@ class CephContext(OSContextGenerator):
         ctxt = {
             'use_syslog': str(config('use-syslog')).lower()
         }
+        rbd_mirroring_mode = config('rbd-mirroring-mode')
         for rid in relation_ids('ceph'):
             for unit in related_units(rid):
                 if not ctxt.get('auth'):
                     ctxt['auth'] = relation_get('auth', rid=rid, unit=unit)
                 if not ctxt.get('key'):
                     ctxt['key'] = relation_get('key', rid=rid, unit=unit)
-                if not ctxt.get('rbd_features'):
+                if rbd_mirroring_mode != "image" and not ctxt.get('rbd_features'):
                     default_features = relation_get('rbd-features', rid=rid, unit=unit)
                     if default_features is not None:
                         ctxt['rbd_features'] = default_features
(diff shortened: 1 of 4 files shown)


AurelienLourot:
CI validation (still ongoing) that this change remains backward-compatible, to be on the safe side:
- inside charm-ceph-rbd-mirror: https://review.opendev.org/762355
- inside charm-nova-compute: https://review.opendev.org/762358

ionutbalutoiu:
> CI validation (still ongoing) that this change remains backward-compatible, to be on the safe side:
> 
> * inside charm-ceph-rbd-mirror: https://review.opendev.org/762355
> * inside charm-nova-compute: https://review.opendev.org/762358

Hello Aurelien!

Could you please do the same for `ceph-mon` and `cinder-ceph` to validate the backwards-compatibility ? We will ultimately have to sync charm helpers there as well.

However, the `nova-compute` charm doesn't need any changes from here for the Ceph Replication spec implementation. The only change needed to `nova-compute` is https://github.com/ionutbalutoiu/charm-nova-compute/commit/78acc4fa058f2a346b5f83fd1d5049f42e4c8575, which is not part of the charm-helpers.

AurelienLourot:
Sure :)
- ceph-mon: https://review.opendev.org/762485
- cinder-ceph: https://review.opendev.org/762486

AurelienLourot (approved):
LGTM, thanks! Happy to land it after successful https://review.opendev.org/#/c/762485/

AurelienLourot:
Happy to merge now but it looks like I'm lacking permissions to do so. Digging.

wolsen (approved):
All looks good to me

thedac (approved):
Seconded. +1

PR #527, opened 2020-11-10
core/host: Speedup chownr by avoiding a couple of syscalls

pponnuvel:
*Closed https://github.com/juju/charm-helpers/pull/516 and did the requested changes here.*

chownr function calls os.path.exists and os.path.lexists for every file. This can be avoided by checking the error case that they were meant to handle (i.e. the symlink's target is missing).

As noted in [1898605](https://bugs.launchpad.net/charm-helpers/+bug/1898605), this is particularly noticeable in ceph-osd charms.

Closes-bug: #1898605

Signed-off-by: Ponnuvel Palaniyappan <pponnuvel@gmail.com>


ajkavanagh (approved):
LGTM

PR #532, opened 2020-11-19
Partially revert "Updates for the Cinder Ceph Replication spec (#526)"

AurelienLourot:
Partially revert "Updates for the Cinder Ceph Replication spec (#526)"
This partially reverts commit ead529b.

When pulled into charm-ceph-radosgw the reverted change
leads to both an incomplete context when rendering ceph.conf
and a missing 'mon' relation.

@ionutbalutoiu @oprinmarius Please re-create the reverted change as a new pull request and validate it against charm-ceph-radosgw, thanks!

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -777,13 +777,10 @@ class AMQPContext(OSContextGenerator):
 
 class CephContext(OSContextGenerator):
     """Generates context for /etc/ceph/ceph.conf templates."""
-
-    def __init__(self, ceph_relation='ceph'):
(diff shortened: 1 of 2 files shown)


AurelienLourot:
In that charm, a class `MonContext` inheriting `CephContext` and manually setting `interfaces` is used, which now doesn't work anymore with that reverted change.

thedac (dismissed):
+1 to the revert.

ionutbalutoiu:
+1 for the revert.

We'll be shipping these reverted changes as part of the updates to the `cinder-ceph` charm.

Thank-you for unblocking the charm helpers.

thedac (approved):
Still +1

PR #538, opened 2020-11-25
Add support for pmd-socket-cores

gnuoy:
The OpenStack charms allow a user to specify how many cores in
each NUMA socket will be dedicated to non-datapath DPDK threads
via the dpdk-socket-cores config option. This change adds support
for a new config option, pmd-socket-cores. pmd-socket-cores
specifies how many cores in each NUMA socket will be dedicated to
packet processing.


ajkavanagh (approved):
Since this is 'ready-to-go' and passes its tests, let's merge this as it enables other feature development on the OVN charms.  (https://bugs.launchpad.net/charm-ovn-chassis/+bug/1905416)

PR #539, opened 2020-11-26
Add max_check_attempts support for nrpe checks

xavpaice:
Add support to nrpe for optional attribute max_check_attempts.  If
supplied, max_check_attempts is added to the NRPE check when consumed by
the Nagios host, overriding the default.

### DIFF
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -142,7 +142,7 @@ define service {{
 }}
 """)
 
-    def __init__(self, shortname, description, check_cmd):
+    def __init__(self, shortname, description, check_cmd, max_check_attempts=None):
         super(Check, self).__init__()
         # XXX: could be better to calculate this from the service name
         if not re.match(self.shortname_re, shortname):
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -327,6 +328,11 @@ def write(self):
             nrpe_monitors[nrpecheck.shortname] = {
                 "command": nrpecheck.command,
             }
+            # If we were passed max_check_attempts, add that to the relation data
+            try:
+                nrpe_monitors[nrpecheck.shortname]['max_check_attempts'] = nrpecheck.max_check_attempts
```
ChrisMacNaughton, line 333: If I'm reading this code right, then the value of 'max_check_attempts' in the nrpe_monitors will be explicitly set to None, and the only test confirms that adding configuration for this new option works. I'd like to see a test with this not being set along with being set to confirm that it works, as expected, in both situations.

xavpaice: Thanks for this - it might be that copying the entire test is overkill, but the more I thought about it the more I thought it was worthwhile.  I've also updated to add in modification of the template which is exported to /var for remote Nagios hosts to collect, though I don't know if anyone still uses that until we remove it entirely we should maintain it as current.

ChrisMacNaughton (approved):
Looks reasonable to me

Issue #540, opened 2020-11-30
distributed_wait maintenance message could use status updates when set to long values.


afreiberger:
In charmhelpers/contrib/hahelpers/cluster.py distributed_wait sets a maintenance message listing how long the charm is waiting before it continues to it's next step.

When dealing with longer wait times, it would be helpful for operator view for this to count down by 15 or 30 second increments rather than saying "sleeping for X seconds" for the entire X seconds.  This becomes more necessary when modulo_nodes is increased or wait time is increased beyond the typical settings of 3 nodes and 30 second each.

For example, if I have a known_wait of 5 minutes, my third unit waits 10 minutes.  I'd like to know an updated time still to be slept in between 10 minutes and the end of the wait state.

Ticking the message to remaining time every 15, 30, or 60 seconds would be very helpful.

PR #541, opened 2020-11-30
Add elapsed time status updates for distributed_wait

afreiberger:
addresses issue #540 to provide operator feedback on elapsed time of long distributed wait times.


afreiberger:
I think this may break tests, as it would no longer match 0 second sleep for leader unit, as the sleep would be bypassed as remaining = 0

afreiberger:
updated leader test to expect no waiting.

charmhelpers/contrib/hahelpers/cluster.py
```diff
@@ -404,11 +404,25 @@ def distributed_wait(modulo=None, wait=None, operation_name='operation'):
...
-    msg = "Waiting {} seconds for {} ...".format(calculated_wait,
-                                                 operation_name)
-    log(msg, DEBUG)
-    status_set('maintenance', msg)
-    time.sleep(calculated_wait)
+
+    # update status every 60 seconds with wait time elapsed
+    remaining_wait = calculated_wait
+    elapsed = 0
+    while remaining_wait:
```
ajkavanagh, line 411: I think this would be more robust if it said:

```python
while remaining_wait > 0:
    ...
```

I know that by inspection the code will work, it's just the defensive programmer in me!

tests/contrib/hahelpers/test_cluster_utils.py
```diff
@@ -531,7 +531,7 @@ def test_distributed_wait(self, log, modulo_distribution, sleep,
         is_leader.return_value = True
         cluster_utils.distributed_wait(modulo=9, wait=23)
         modulo_distribution.assert_not_called()
-        sleep.assert_called_with(0)
+        sleep.assert_not_called(0)
```
ajkavanagh, line 534: It would probably be good to check that the looping does actually work here.  i.e. verify that it works for an odd number of "not divisible by" 60 seconds wait.

ajkavanagh (commented):
This is a nice patch which would enhance the operator's experience.  Just a couple of minor observations.  Thanks.

Issue #544, opened 2020-12-08
ubuntu _add_apt_repository needs to pass both http_proxy and https_proxy for some PPA keys to be found


afreiberger:
Add-apt-repository requires both http_proxy and https_proxy settings for some PPAs.

This line integrates the https proxy variable into the call used by a charm to add-apt-repository, but while some keys are served by keyserver.ubuntu.com via https, some are also pointing to http endpoints and failing to be proxied.

https://github.com/juju/charm-helpers/blob/198c0e3a27e7af40ab9f811b81c4e51809fe9722/charmhelpers/fetch/ubuntu.py#L649

As an example, I'm able to add ppa:telegraf-devs/ppa from the CLI if I set both http_proxy and https_proxy, but it fails on just https_proxy being set.

```
ubuntu@host:~$ sudo https_proxy=http://10.0.0.1:8000/ add-apt-repository ppa:telegraf-devs/ppa
 
 More info: https://launchpad.net/~telegraf-devs/+archive/ubuntu/ppa
Press [ENTER] to continue or ctrl-c to cancel adding it

gpg: keyring `/tmp/tmphx3xn4br/secring.gpg' created
gpg: keyring `/tmp/tmphx3xn4br/pubring.gpg' created
gpg: requesting key C94406F5 from hkp server keyserver.ubuntu.com
?: [fd 4]: read error: Connection reset by peer
gpgkeys: HTTP fetch error 7: couldn't connect: eof
gpg: no valid OpenPGP data found.
gpg: Total number processed: 0
gpg: keyserver communications error: keyserver unreachable
gpg: keyserver communications error: public key not found
gpg: keyserver receive failed: public key not found
Failed to add key.
ubuntu@host:~$ sudo http_proxy=http://10.0.0.1:8000/ https_proxy=http://10.0.0.1:8000/ add-apt-repository ppa:telegraf-devs/ppa
 
 More info: https://launchpad.net/~telegraf-devs/+archive/ubuntu/ppa
Press [ENTER] to continue or ctrl-c to cancel adding it

gpg: keyring `/tmp/tmp42e10sgd/secring.gpg' created
gpg: keyring `/tmp/tmp42e10sgd/pubring.gpg' created
gpg: requesting key C94406F5 from hkp server keyserver.ubuntu.com
gpg: /tmp/tmp42e10sgd/trustdb.gpg: trustdb created
gpg: key C94406F5: public key "Launchpad PPA for Telegraf Devs" imported
[...]

PR #545, opened 2020-12-08
Add http_proxy to environment for add_apt_repository to address #544

afreiberger:
It was recently found that some references in the ubuntu keyserver point to http:// hosted GPG keys for PPAs which will require add-apt-repository to have both https_proxy and http_proxy variables set.

See issue #544 for further details.


ajkavanagh (approved):
LGTM - thanks for the patch/fix.

PR #546, opened 2020-12-14
 Add 'is_ssl_enabled' for bug 1851610 to check if SSL module is enabled

mullumaus:
Add 'is_ssl_enable' for bug 1851610 to check if SSL module is enabled or not

The gerrit review link:  https://review.opendev.org/c/openstack/charm-neutron-api/+/766883

### DIFF
--- a/charmhelpers/contrib/hardening/audits/apache.py
+++ b/charmhelpers/contrib/hardening/audits/apache.py
@@ -98,3 +98,8 @@ class DisabledModuleAudit(BaseAudit):
     def _restart_apache():
         """Restarts the apache process"""
         subprocess.check_output(['service', 'apache2', 'restart'])
+
+    @staticmethod
+    def is_ssl_enabled():
+        """Check if SSL module is enabled or not"""
+        return 'ssl' in DisabledModuleAudit._get_loaded_modules()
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/hardening/audits/apache.py
```diff
@@ -98,3 +98,8 @@ def _disable_module(module):
     def _restart_apache():
         """Restarts the apache process"""
         subprocess.check_output(['service', 'apache2', 'restart'])
+
+    @staticmethod
+    def is_ssl_enable():
```
wolsen, line 103: The method should be named `is_ssl_enabled` vs `is_ssl_enable`.

wolsen (changes requested):
In addition to the inline comment, can we please add a unit test?

charmhelpers/contrib/hardening/audits/apache.py
mullumaus: Thanks Billy, I removed the method to 'is_ssl_enabled' and added a unit test for it in tests/contrib/hardening/audits/test_apache_audits.py

ajkavanagh (approved):
LGTM - just need @wolsen to take another look.

wolsen (approved):

PR #550, opened 2020-12-17
Add connect_timeout to the MySQLHelper class

ajkavanagh:
This is needed for the mysql-router charm as the connect method can hang
indefinitely without it if the associated local router fails to connect.
This happens during an update-status hook.

LP Bug: 1907250


thedac (approved):
This tests well and looks like a great addition. Landing.

PR #551, opened 2020-12-18
Add support to cert_utils for bespoke bindings

gnuoy:
The cert_utils use ADDRESS_MAP as a definitive list of
bindings in various places. This PR allows a charm to extend that
map with its own bindings. It also allows the 'config' and 'override' keys to
be absent. For some bindings there is no corresponding charm config options
so it is reasonable for them to be absent.

* Update get_certificate_request, get_certificate_sans and create_ip_cert_links
  to accept an address_map kwarg to allow bespoke bindings to be resolved.
* Fix bug in get_certificate_request which removes an IP from _sans even
  if it has not yet been added to the certificate request (occurs if an ip
  is in _sans but if hostname_override is None)
* Update _get_address_override to accept an address_map kwarg which can then
  be used to search for override key.
* Update _get_address_override to allow the override key to be missing and
  return None if it is.
* Update resolve_address to accept an address_map kwarg.
* Update resolve_address to allow the 'config' key to be missing from the
  address map.

### DIFF
--- a/charmhelpers/contrib/openstack/cert_utils.py
+++ b/charmhelpers/contrib/openstack/cert_utils.py
@@ -120,7 +120,7 @@ class CertRequest(object):
         return req
 
 
-def get_certificate_request(json_encode=True, bindings=None):
+def get_certificate_request(json_encode=True, bindings=None, address_map=None):
     """Generate a certificate requests based on the network configuration
 
     :param json_encode: Encode request in JSON or not. Used for setting
@@ -129,9 +129,13 @@ def get_certificate_request(json_encode=True, bindings=None):
     :param bindings: List of bindings to check in addition to default api
                      bindings.
     :type bindings: list of strings
+    :param address_map: A collection of dictionaries specifying how different
+                        bindings should be handled.
+    :type address_map: Dict[str, Dict[str, str]]
     :returns: CertRequest request as dictionary or JSON string.
     :rtype: Union[dict, json]
     """
+    address_map = address_map or ADDRESS_MAP
     if bindings:
         # Add default API bindings to bindings list
         bindings = set(bindings + get_default_api_bindings())
@@ -141,15 +145,19 @@ def get_certificate_request(json_encode=True, bindings=None):
     req = CertRequest(json_encode=json_encode)
     req.add_hostname_cn()
     # Add os-hostname entries
-    _sans = get_certificate_sans()
-
+    _sans = get_certificate_sans(bindings=bindings, address_map=address_map)
     # Handle specific hostnames per binding
     for binding in bindings:
-        hostname_override = config(ADDRESS_MAP[binding]['override'])
+        hostname_override = None
+        override = address_map[binding].get('override')
+        if override:
+            hostname_override = config(override)
         try:
-            net_addr = resolve_address(endpoint_type=binding)
+            net_addr = resolve_address(
+                endpoint_type=binding,
+                address_map=address_map)
             ip = network_get_primary_address(
-                ADDRESS_MAP[binding]['binding'])
+                address_map[binding]['binding'])
             addresses = [net_addr, ip]
             vip = get_vip_in_network(resolve_network_cidr(ip))
             if vip:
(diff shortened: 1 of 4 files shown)


charmhelpers/contrib/openstack/cert_utils.py
```diff
@@ -160,12 +168,12 @@ def get_certificate_request(json_encode=True, bindings=None):
                     binding,
                     hostname_override,
                     addresses)
-            # Remove hostname specific addresses from _sans
```
thedac, line 163: This does seem to be a bug. I proved it to myself with a unit test.

thedac (commented):
See my thoughts on the charm PR. I really think the address_map change is orthogonal to the bug. It is an interesting addition but I had to "see" the problem without it. I think we should treat it separately.

The indentation problem san removal is a good fix.

We can discuss more when we get back from break.

gnuoy:
I don't think the change proposed here is orthogonal to the bug targeted. I think it may appear that way as there are a number of different issues at play. 

Firstly when deploying as described in https://bugs.launchpad.net/charm-mysql-innodb-cluster/+bug/1908306 the db-router address is missing (and incorrectly contains the ip addresses of its peers) in the certificate request. As you say in the corresponding PR to this change *1 we could simplify the mysql-innodb-cluster change to use `instance.cluster_address` rather than `instance.cluster_peer_addresses`. That issue is not part of this PR.

With the certificate request fixed this leads on to the next bug, the supplied certificate has a sans entry for the db-router address but still no symlink is created. The charm explicitly calls `cert_utils.create_ip_cert_links` but `create_ip_cert_links` derives which ip symlinks need creating by calling `cert_utils.get_certificate_request`. `cert_utils.get_certificate_request` builds the certificate request by examining the bindings defined in `charmhelpers.contrib.openstack.ip.ADDRESS_MAP`. Neither db-router or cluster endpoints are part of that map and so are not included in the certificate request and so `create_ip_cert_links` does not create symlinks for the bindings.

Fundamentally charms should reuse the same code to construct the request when sending the cert request and when responding to the supplied certificate. To achieve this, and to avoid bugs like this happening in the future in other charms, the functions in cert_utils should support including additional endpoints in a certificate request. The way I implemented that was to allow a charm to supply an address map which includes all the bindings it wants included in a certificate request. This leads to this charm helper change. The idea here being that a charm should be able to specify any additional bindings it wants to include in a certificate request. In the case of mysql-innodb-cluster I would like to create a follow up PR to remove manual construction of the certificate request *2 and replace it with a call to `cert_utils.get_certificate_request` with an address map which defines additional bindings. 

With the previous issues fixed this led to the final bug in `get_certificate_request` which removes an IP from sans even if it has not yet been added to the certificate request (occurs if an ip is in the sans list but no hostname_override has been set)


*1 https://review.opendev.org/c/openstack/charm-mysql-innodb-cluster/+/767757
*2 https://opendev.org/openstack/charm-mysql-innodb-cluster/src/branch/master/src/reactive/mysql_innodb_cluster_handlers.py#L323-L331


To reproduce deploy bundle https://paste.ubuntu.com/p/t3bJMPmqhv/ with network spaces defined for internal, admin, storage, public and data. Then run `zaza.openstack.charm_tests.vault.setup.auto_initialize_no_validation_no_wait`. IP symlinks will be missing from `/etc/mysql/tls/mysql-innodb-cluster/` on the mysql-innodb-cluster units for the db-router and cluster IPs.

charmhelpers/contrib/openstack/cert_utils.py
gnuoy: fwiw worth it is covered by this addition to the unit tests *1, without the change '10.30.0.2' is removed from the certificate request.


*1 https://github.com/juju/charm-helpers/pull/551/files#diff-facadc69ca25c0fccf4cc86904ab4e30e32088dbf213543521cc1c6bcf222ee3R136

thedac:
I am about to relent on this. I need to test one more nuance to the problem.

FWIW, I was originally trying to avoid the need for ADDRESS_MAP with the bindings parameter. We could do the same by adding it to the create_ip_cert_links. My reasoning is that many many non-API charms will not use overrides the way ADDRESS_MAP does.

Anyway, I want to test one more thing, and then let's discuss.

thedac:
Just for a reference point: https://github.com/juju/charm-helpers/pull/556

gnuoy:
Superseded by https://github.com/juju/charm-helpers/commit/27b3f59ddeaefff6f1b4a269959e522cb42c1639

PR #552, opened 2020-12-19
Run actual tests of _calculate_workers and _num_cpus

nobuto-m:
Previously it succeeded always regardless of any values set. Also,
handle corner cases properly such as worker-multiplier=0.0 because the
number of workers cannot be zero.


wolsen (approved):
LGTM

nobuto-m:
@wolsen Looks like you merged the other way around? I see your merge master -> my branch instead of my branch -> master.

nobuto-m:
Reverted the merge commit so it's ready for review/merge again.

PR #553, opened 2020-12-21
Cap the number of workers always when multiplier is not set

nobuto-m:
**This pull request depends on #552.**

Previously the cap is only applied to the units in containers. However,
services in a bare metal also requires a sensible cap. Otherwise,
nova-compute, for example, will have 256 workers for nova-api-metadata
out of the box which is an overkill with the following system.

32 cores * 2 threads/core * 2 sockets * 2 (default multiplier) = 256

Let's cap the number of workers as 4 always, then let operators override
it with an explicit config to worker-multiplier.

Closes-Bug: [LP: #1843011](https://bugs.launchpad.net/charm-nova-compute/+bug/1843011)

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -74,7 +74,6 @@ from charmhelpers.core.host import (
     pwgen,
     lsb_release,
     CompareHostReleases,
-    is_container,
 )
 from charmhelpers.contrib.hahelpers.cluster import (
     determine_apache_port,
@@ -1581,8 +1580,7 @@ class BindHostContext(OSContextGenerator):
             return {'bind_host': '0.0.0.0'}
 
 
-MAX_DEFAULT_WORKERS = 16
-MAX_DEFAULT_WORKERS_CONTAINER = 4
+MAX_DEFAULT_WORKERS = 4
 DEFAULT_MULTIPLIER = 2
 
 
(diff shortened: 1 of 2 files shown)


wolsen (dismissed):
LGTM

wolsen (dismissed):
This limits worker counts to a sensible default and allows the ability to override it as necessary. 

I'm on the fence about changing the value for MAX_DEFAULT_WORKERS and introducing a new constant for the same value (i.e. MAX_DEFAULT_WORKERS_CONTAINER). It may cause unintended side-effects in charms using this. However, in all reference of charms I *know* to be using this, it will not be a problem, and arguably it shouldn't be a problem as they would reference the constant intentionally.

charmhelpers/contrib/openstack/context.py
```diff
@@ -1581,7 +1581,8 @@ def __call__(self):
             return {'bind_host': '0.0.0.0'}
 
 
-MAX_DEFAULT_WORKERS = 4
+MAX_DEFAULT_WORKERS = 16
+MAX_DEFAULT_WORKERS_CONTAINER = 4
```
ajkavanagh, line 1585: I think having two defaults is slightly confusing. The default is to set a max workers to some sensible value, if the operator of the charm doesn't set the worker-multiplier.  If they do, then neither maximum will be used.  I think conversations will be easier if it goes "the default max is 4, if you need to change that, then please set the worker-multiplier".

nobuto-m: I actually like this approach for simplicity. I will wait for other feedback and will update the PR next week.

ajkavanagh (approved):
I'm +2 on this; looks good.

nobuto-m:
Hi, is there anything left I need to do to get it merged? Thanks,

PR #555, opened 2021-01-04
get_hostname(), get_host_ip(): try socket.gethostbyaddr/name() on DNS exceptions

mfoliveira:
Currently, `get_hostname()` and `get_host_ip()` do not handle exceptions
when calling `ns_query()` thus are susceptible to exceptions raised by
`dns.resolver.query()`, except for `dns.resolver.NXDOMAIN` (handled.)

When that happens, they do not reach `socket.gethostbyaddr/name()` and
thus ignore `/etc/hosts` _only on the releases w/out `systemd-resolved`_
(which apparently does that under the covers.)

This is an inconsistency between releases with//out `systemd-resolved`
(e.g., Trusty/Xenial vs Bionic/Focal.)

In order to fix that, handle and store the exception from `ns_query()`,
then try `socket.gethostbyaddr/name()` inconditionally as previously,
and _if it also fails_, raise that exception instead.

There is no behavior change for all other code paths/cases, and this
case gets fixed/improvement (Trusty/Xenial on par with Bionic/Focal.)

Note: the usage of `reraise()` instead of just `raise` is to smooth out
differences between python2/python3 (exception chaining is py3-only)
and get a consistent backtrace that shows previous functions on py2.

Cases:
-----

The variables: whether `ns_query()` gets a name/address, `NXDOMAIN`
exception, or other exceptions; and whether `gethostbyaddr/name()`
gets an exception or not (i.e., doesn't/does find in `/etc/hosts`).

The result: what `get_hostname() / get_host_ip()` return/raise.

1) `ns_query()` gets name/address answer (returns answer).
   Result: answer.

2) `ns_query()` gets `NXDOMAIN` exception (returns None):

  2.1) `socket.gethostbyaddr/name()` finds an entry in `/etc/hosts`.
       Result: entry.

  2.2) `socket.gethostbyaddr/name()` finds NO entry in `/etc/hosts`.
       Result: None.

3) `ns_query()` gets another exception (dep. on release): [PATCH]

- Before)

  Result: exception; does not try `socket.gethostbyaddr/name()`.

- After)

  3.1) `socket.gethostbyaddr/name()` finds an entry in `/etc/hosts`.
       Result: entry.

  3.2) `socket.gethostbyaddr/name()` finds NO entry in `/etc/hosts`.
       Result: exception (original exception, not the new one.)

Tests:
-----

Tested on Trusty/Xenial/Bionic/Focal with both python2/python3
(with//out an entry in `/etc/hosts` for the address/name in test.)

1) Not changed; not tested.

2) These hit `NXDOMAIN` in `ns_query()`, not exception, returns None;
and may raise exceptions in `socket.gethostbyaddr/name()` based on
`/etc/hosts`:
- `get_hostname('1.2.3.4')`
- `get_host_ip('example.nxd')`

3) These may raise exceptions in `ns_query()` (depends on release)
or `socket.gethostbyaddr/name()`, based on `/etc/hosts`:
- `get_hostname('10.10.10.10')`
- `get_host_ip('test-hostname')`

3) This can generate exception in `ns_query()` (indep of release)
- `get_host_ip('test-hostname')` // with DNS packets dropped, eg,
       // `sudo iptables -A|-D OUTPUT -p udp --dport 53 -j DROP`

Signed-off-by: Mauricio Faria de Oliveira <mfo@canonical.com>

### DIFF
--- a/charmhelpers/contrib/network/ip.py
+++ b/charmhelpers/contrib/network/ip.py
@@ -17,8 +17,10 @@ import re
 import subprocess
 import six
 import socket
+import sys
 
 from functools import partial
+from six import reraise
 
 from charmhelpers.fetch import apt_install, apt_update
 from charmhelpers.core.hookenv import (
@@ -493,14 +495,25 @@ def get_host_ip(hostname, fallback=None):
     if is_ip(hostname):
         return hostname
 
-    ip_addr = ns_query(hostname)
-    if not ip_addr:
-        try:
-            ip_addr = socket.gethostbyname(hostname)
-        except Exception:
-            log("Failed to resolve hostname '%s'" % (hostname),
-                level=WARNING)
-            return fallback
+    ip_addr = None
+    exc_value = None
+    try:
+        ip_addr = ns_query(hostname)
+    except Exception:
+        # If DNS lookup failed, try nsswitch (may check /etc/hosts).
+        # If that fails too, raise this exception/traceback instead.
+        (exc_type, exc_value, exc_traceback) = sys.exc_info()
+    finally:
+        if not ip_addr:  # Hit exception (non-NXDOMAIN) or no lookup/answer.
+            try:
+                ip_addr = socket.gethostbyname(hostname)
+            except Exception:
+                log("Failed to resolve hostname '%s'" % (hostname),
+                    level=WARNING)
+                if fallback:
+                    return fallback
+                if exc_value:
+                    reraise(exc_type, exc_value, exc_traceback)
     return ip_addr
 
 
@@ -519,14 +532,23 @@ def get_hostname(address, fqdn=True):
                 apt_install("python3-dnspython", fatal=True)
             import dns.reversename
 
+        result = None
+        exc_value = None
         rev = dns.reversename.from_address(address)
-        result = ns_query(rev)
-
-        if not result:
-            try:
-                result = socket.gethostbyaddr(address)[0]
-            except Exception:
-                return None
+        try:
+            result = ns_query(rev)
+        except Exception:
+            # If reverse DNS lookup failed, try nsswitch (may check /etc/hosts)
+            # If that fails too, raise this exception/traceback instead.
+            (exc_type, exc_value, exc_traceback) = sys.exc_info()
+        finally:
+            if not result:  # Hit exception (non-NXDOMAIN) or no lookup/answer.
+                try:
+                    result = socket.gethostbyaddr(address)[0]
+                except Exception:
+                    if exc_value:
+                        reraise(exc_type, exc_value, exc_traceback)
+                    return None
     else:
         result = address
 


mfoliveira:
This has been tested before/after patching with this script.
The output has been compared with `vimdiff` and is as expected.
Tested: py2/py3 on T/X/B/F.

```
#!/bin/sh

for TEST in \
  "get_hostname('10.10.10.10')" \
  "get_host_ip('test-hostname')" \
  "get_hostname('1.2.3.4')" \
  "get_host_ip('example.nxd')" \
  "get_host_ip('TEST-HOSTNAME')"
do
        for HOSTS in 'No' 'Yes'
        do
                if [ "$HOSTS" = 'No' ]; then
                        sudo sed -i '/10.10.10.10 test-hostname/d' /etc/hosts # delete entry
                        sudo sed -i '/1.2.3.4 example.nxd/d'       /etc/hosts # delete entry
                else
                        sudo sed -i '$a 10.10.10.10 test-hostname' /etc/hosts # append entry
                        sudo sed -i '$a 1.2.3.4 example.nxd'       /etc/hosts # append entry
                fi

                if echo "$TEST" | grep -q 'TEST-HOSTNAME'; then # use uppercase to flag this
                        sudo iptables -A OUTPUT -p udp --dport 53 -j DROP # block DNS
                fi

                echo
                echo "- Test: $TEST"
                echo "- Entry in /etc/hosts: $HOSTS"

                for PYTHON in 'python2' 'python3'
                do
                        # systemd-resolved: force clear DNS cache
                        if which systemd-resolve >/dev/null 2>&1; then
                                sudo systemctl restart systemd-resolved.service >/dev/null 2>&1
                        fi

                        echo
                        echo "  - Python: $PYTHON"
                        $PYTHON -c "from charmhelpers.contrib.network.ip import get_hostname, get_host_ip; print($TEST)" 2>&1 \
                          | sed 's,^,    ,'
                done

                if echo "$TEST" | grep -q 'TEST-HOSTNAME'; then # use uppercase to flag this
                        sudo iptables -D OUTPUT -p udp --dport 53 -j DROP >/dev/null 2>&1 # unblock DNS
                fi

                if [ "$HOSTS" = 'Yes' ]; then
                        sudo sed -i '/10.10.10.10 test-hostname/d' /etc/hosts # delete entry
                        sudo sed -i '/1.2.3.4 example.nxd/d'       /etc/hosts # delete entry
                fi
        done
done
```

mfoliveira:
Updated patch to fix PEP8 errors.

mfoliveira:
The previous PEP8 failures are now passing.

The remaining failure doens't seem to be related to this patch.

```
$ tox -c tox.ini -e $ENV
pep8 create: /home/travis/build/juju/charm-helpers/.tox/pep8
pep8 installdeps: -r/home/travis/build/juju/charm-helpers/test-requirements.txt
ERROR: invocation failed (exit code 1), logfile: /home/travis/build/juju/charm-helpers/.tox/pep8/log/pep8-1.log
```

ajkavanagh:
> The previous PEP8 failures are now passing.
> 
> The remaining failure doens't seem to be related to this patch.
> 
> ```
> $ tox -c tox.ini -e $ENV
> pep8 create: /home/travis/build/juju/charm-helpers/.tox/pep8
> pep8 installdeps: -r/home/travis/build/juju/charm-helpers/test-requirements.txt
> ERROR: invocation failed (exit code 1), logfile: /home/travis/build/juju/charm-helpers/.tox/pep8/log/pep8-1.log
> ```

Digging into this, it looks like pip 20.3.1 strikes again.  The latest version of pip has a new resolver which is causing issues across a whole range of projects.  Extremely annoying, but the work did apparently need to be done to pip.  It is exposing all kinds of problems with dependencies and requirements files.

It may be necessary to downgrade pip to < 20.3 to get around this whilst a more permanent fix is done.  This is what we have done in other tox.ini files:

```tox.ini
# NOTES:
# * We avoid the new dependency resolver by pinning pip < 20.3, see
#   https://github.com/pypa/pip/issues/9187
# * Pinning dependencies requires tox >= 3.2.0, see
#   https://tox.readthedocs.io/en/latest/config.html#conf-requires
# * It is also necessary to pin virtualenv as a newer virtualenv would still
#   lead to fetching the latest pip in the func* tox targets, see
#   https://stackoverflow.com/a/38133283
requires = pip < 20.3
           virtualenv < 20.0
# NOTE: https://wiki.canonical.com/engineering/OpenStack/InstallLatestToxOnOsci
minversion = 3.2.0
```

This goes at the end of the `[tox]` section at the top of the tox.ini.  If you put this in temporarily to see if it will pass its tests?  I'll raise a bug about this separately so that we can track what's happening with it.

ajkavanagh:
Once #558 lands, merge from / rebase onto master and the tests will pass.

mfoliveira:
Rebased to get the `tox.ini` changes from #558, which are in there now, and pushed.

mfoliveira:
> Once #558 lands, merge from / rebase onto master and the tests will pass.

Thanks for looking into this; and indeed, all tests pass now.

mfoliveira:
Hi @ajkavanagh 

Just following up in case you think this still makes sense.
Quick summary in my last comment in #525, if that helps.

Thanks!

ajkavanagh:
> Hi @ajkavanagh
> 
> Just following up in case you think this still makes sense.
> Quick summary in my last comment in #525, if that helps.
> 
> Thanks!

Unfortunately, I needs another rebase; this is actually a good time to get this in as we are at the start of the cycle.  Thanks.

mfoliveira:
@ajkavanagh
> Unfortunately, I needs another rebase; this is actually a good time to get this in as we are at the start of the cycle. Thanks.

No problem; done!
Cool, nice timing.

BTW, this PR doesn't have the behavior-changing bits mentioned in the previous PR; if you'd like that in (even if just for reviewing and possibly dropping it later), just let me know and I'll add its patch.

Thanks!

charmhelpers/contrib/network/ip.py
```diff
@@ -493,14 +495,24 @@ def get_host_ip(hostname, fallback=None):
...
+    finally:
+        if not ip_addr:  # Hit exception (non-NXDOMAIN) or no lookup/answer.
+            try:
+                ip_addr = socket.gethostbyname(hostname)
+            except Exception:
+                log("Failed to resolve hostname '%s'" % (hostname),
+                    level=WARNING)
+                if exc_value:
+                    reraise(exc_type, exc_value, exc_traceback)
+                return fallback
```
ajkavanagh, line 515: I think this is a bit awkward for what it is doing.  The sequence is (if I understand the code):

1. ip_addr = ns_query(hostname)
2. ip_addr = socket.gethostbyname(hostname)

So this translates, for me, to:

```python
try:
    return ns_query(hostname)
except Exception as e:  # seems a bit too broad, maybe try to reign it in to a particular set of exceptions?
    try:
        return socket.gethostbyname(hostname)
    except Exception as f:
        log("failed to resolve hostname by either ns_query ({}) or socket.gethostbyname() ({})".format(str(e), str(f)))
        return fallback
```

Because, otherwise the new code in the patch does exactly what the old code does, but in a more complicated way.

charmhelpers/contrib/network/ip.py
```diff
@@ -519,14 +531,23 @@ def get_hostname(address, fqdn=True):
...
+            # If that fails too, raise this exception/traceback instead.
+            (exc_type, exc_value, exc_traceback) = sys.exc_info()
+        finally:
+            if not result:  # Hit exception (non-NXDOMAIN) or no lookup/answer.
+                try:
+                    result = socket.gethostbyaddr(address)[0]
+                except Exception:
+                    if exc_value:
+                        reraise(exc_type, exc_value, exc_traceback)
+                    return None
```
ajkavanagh, line 550: Again does this do exactly the same thing?

ajkavanagh (changes requested):
I'm a bit perplexed really; I must be missing something.  The revised code does exactly the same thing?  (or I'm misreading it, which is equally likely?)

charmhelpers/contrib/network/ip.py
mfoliveira: Uh oh, I've been caught trying to increase my LOC metrics. x)

So, it's not _exactly_ the same thing, because we now handle an exception in `ns_query()`.
That was the issue on T/X, which prevented even trying `gethostby{name,addr}()` for `/etc/hosts`.

The additional complication is, if we get to try `gethostby{name,addr}()` and it fails/raises exception,
we had discussed in the previous PR that it should reraise the exception generated in `ns_query()`
(to keep existing behavior in the library.)

If it's OK to change the behavior now (it wasn't back when we discussed this, if I understood it correctly)
and which exceptions/values are (not) raised/returned to the caller, then absolutely, the suggestion above is much simpler.

I can adjust the code according to what's the right direction now; just let me know.

charmhelpers/contrib/network/ip.py
mfoliveira: Likewise.

charmhelpers/contrib/network/ip.py
ajkavanagh: Okay, I think I've caught up.  (I may not have!)

So, the code is doing this?

```python
try:
    return ns_query(hostname)
except Exception as e:  # seems a bit too broad, maybe try to reign it in to a particular set of exceptions?
    t, v, tb = sys.exc_info()
    try:
        return socket.gethostbyname(hostname)
    except Exception as f:
        log("failed to resolve hostname by either ns_query ({}) or socket.gethostbyname() ({})".format(str(e), str(f)))
        if fallback:
            return fallback
        six.reraise(t, v, tb)
```

It would be nice not to have to use `six`, but the lib is *still* py27 compat.

charmhelpers/contrib/network/ip.py
ajkavanagh: Ah, indeed, see above.

charmhelpers/contrib/network/ip.py
mfoliveira: Hi @ajkavanagh 

Sorry, it's been a busy 2 weeks, and I had to review my own code and choices here. :)

So, the code is _not exactly_ doing that; let me try to clarify.
The complexity comes from sticking to no behavior changes. :/

- We need that `finally:` block, because in the original code `if not ip_addr:` runs inconditionally, not just in case of unhandled exception in `ns_query()` (it can return `None` for `NXDOMAIN`).

- On the `fallback` check/return before `reraise`: good point.
If `fallback` is set, we should not `reraise`, but use that instead.

- Regarding `six`, I've used it to keep the backtraces in py2 consistent or at least useful; py3 has exception chaining, but IIRC in py2 (which doesn't) if we just `raise` the original exception, we lose the backtrace altogether; bad enough for debugging.
And then while I was trying to use a python exception thing one way or another, it would be a syntax failure or other problem in either python version.
So I eventually gave up on syntatic sugar and chose `six` (with this verbose `t, c, tb = exc_info()` approach.)

Hope this helps!

ajkavanagh (approved):
LGTM

ajkavanagh:
@mfoliveira hey there; this got lost in the weeds; let's get it rebased/updated and then we can merge it.  Now is a great time as it's at the start of the next development cycle.  Thanks!

mfoliveira:
Hey @ajkavanagh , no worries, this is very low priority, as it's a corner case and its affected base (Trusty/Xenial) should  decrease. :)

I've updated the patch with the suggested change to use `fallback` if set, before `reraise()`.

I'll try and re-do the testing. I see the development cycles are 3 months for charms, but for my own planning, by when would you need this fully tested? (I plan to do it at most by next week, but just in case something comes up and this slips.)

Thanks!

ajkavanagh:
> Hey @ajkavanagh , no worries, this is very low priority, as it's a corner case and its affected base (Trusty/Xenial) should decrease. :)
> 
> I've updated the patch with the suggested change to use `fallback` if set, before `reraise()`.
> 
> I'll try and re-do the testing. I see the development cycles are 3 months for charms, but for my own planning, by when would you need this fully tested? (I plan to do it at most by next week, but just in case something comes up and this slips.)

Whenever you can fit it in.  It's absolutely the best time in the next few weeks as we are at the beginning of the cycle so the code will get the most testing and see if there are any errors, etc across the whole OpenStack charm estate.  Charm-helpers is used by lots of charms and most groups are are at the beginning of their dev cycles too.

Thanks again for the change.

mfoliveira:
Sorry for the delay in getting back to this.

I'll close this out, as it's handling a corner case in releases with low and decreasing usage (Trusty/Xenial), and it introduces regression risk for (newer) releases that aren't impacted.

Thanks for the review cycles and input, it's definitely been helpful and actually applied in other contributions (and hopefully in more, future contributions too!)

PR #556, opened 2021-01-06
Use bindings for certificate requests

thedac:
Ignore ADDRESS_MAP KeyErrors

Just for a reference.

With a couple of changes to the charm and one to charms.openstack.charm.classes.get_certificate_requests, this works using the centralized helpers.


ajkavanagh:
Failing due to a pip 20.3.1 resolver error; #558 is a temporary fix, so when that lands  merge from / rebase on master and this should pass.

gnuoy (approved):
LGTM, thanks for all your work on his!

PR #559, opened 2021-01-06
Update ubuntu_apt_pkg.py

murraybd:
The first part of dpkg-query's output is the "Desired" status of the package which can be one of "Unknown/Install/Remove/Purge/Hold". Where "Hold" means "A package marked to be on hold is kept on the same version" which is a variation of installed. Subsequently, a package with status of Hold should be considered as installed.


murraybd:
This would fix https://bugs.launchpad.net/charm-helpers/+bug/1908147

thedac (approved):
This looks good. Just waiting on the Travis unit tests and pep8 to land.

PR #560, opened 2021-01-07
Add support for new openstack-release package

ChrisMacNaughton:
The new openstack-release package can be used to determine
the OpenStack codename that is installed on a given machine
when it is available.


ChrisMacNaughton:
For reference sake, the /etc/openstack-release file used in the PR looks like

```
OPENSTACK_CODENAME=wallaby
```

and comes from https://git.launchpad.net/~ubuntu-openstack-dev/ubuntu/+source/openstack-release/tree/etc/openstack-release

AurelienLourot (approved):

PR #561, opened 2021-01-14
Stop getting {public,private}-address from unit-get

gnuoy:
In Juju 2.8rc3 `unit-get public-address` became unreliable
(Bug #1910973). Since getting an address this was is deprecated
switch the OpenStack functions to prefer network-get. However,
fallback to the old method to support old versions of Juju for
the time being.


thedac (approved):
Looks good. Landing.

PR #562, opened 2021-01-17
Fix (new) local_address() to work with pre-2.9 juju

ajkavanagh:
The local_address() function will fallback on unit-get public-address
if the network-get --primary-address juju-info fails.
However, it appears that it does fail with NoNetworkBinding rather than
notimplemented (which is probably just an error).  This fixes things so
that it will fallback if network-get fails.


AurelienLourot (approved):

ChrisMacNaughton (approved):

gnuoy:
Ideally we would not use `network-get --primary-address ` as it is deprecated, but bug https://bugs.launchpad.net/juju/+bug/1863916 stops us from doing that.

gnuoy (approved):

PR #563, opened 2021-01-28
Add JUJU_DEPARTING_UNIT support

freyes:
Juju 2.8 introduced the new environment variable JUJU_DEPARTING_UNIT
which holds the name of the unit that's being removed, this allows
charm authors to remove the unit from the cluster during the execution
of the *-relation-departed hook.

https://discourse.charmhub.io/t/juju-2-8-0-release-notes/

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -226,6 +226,17 @@ def relation_id(relation_name=None, service_or_unit=None):
         raise ValueError('Must specify neither or both of relation_name and service_or_unit')
 
 
+def departing_unit():
+    """The departing unit for the current relation hook.
+
+    Available since juju 2.8.
+
+    :returns: the departing unit, or None if the information isn't available.
+    :rtype: Optional[str]
+    """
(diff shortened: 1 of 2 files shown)


charmhelpers/core/hookenv.py
```diff
@@ -226,6 +226,13 @@ def relation_id(relation_name=None, service_or_unit=None):
         raise ValueError('Must specify neither or both of relation_name and service_or_unit')
 
 
+def departing_unit():
+    """The departing unit for the current relation hook.
+
+    Available since juju 2.8"""
```
ajkavanagh, line 232: Please could we start to add type information for the params and return values for new functions (or if we touch old ones).  It really helps with debugging to see the intended types that a function takes and returns.

e.g. in this case:

```
:returns: the departing unit, or None if the information isn't available.
:rtype: Optional[str]
```

freyes: done.

niedbalski (approved):

PR #565, opened 2021-02-02
Issue564: add None check for max_check_attempts in nrpe check

### LINKED ISSUE #564: nrpe.write will add max_check_attempts: null which will cause nagios to break
In commit https://github.com/juju/charm-helpers/commit/198c0e3a27e7af40ab9f811b81c4e51809fe9722 , 
a new kwarg `max_check_attempts=None` was introduced to `nrpe.Check` class.
However, it doesn't check the None condition, which will always render `max_check_attempts: null` in check cfg and cause nagios to break with latest charmhelpers.

We need to add None check to ensure `max_check_attempts` only be rendered when it's not None.

guoqiao:
In commit 198c0e3, a new kwarg max_check_attempts=None was introduced to nrpe.Check class.
However, it doesn't check the None condition, which will always render max_check_attempts: null
in check cfg and cause nagios to break with latest charmhelpers.

This patch add None check to ensure max_check_attempts only be rendered when it's not None.

Closes #564 

Signed-off-by: Joe Guo <joe.guo@canonical.com>

### DIFF
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -337,10 +337,8 @@ class NRPE(object):
                 "command": nrpecheck.command,
             }
             # If we were passed max_check_attempts, add that to the relation data
-            try:
+            if nrpecheck.max_check_attempts is not None:
                 nrpe_monitors[nrpecheck.shortname]['max_check_attempts'] = nrpecheck.max_check_attempts
-            except AttributeError:
-                pass
 
         # update-status hooks are configured to firing every 5 minutes by
         # default. When nagios-nrpe-server is restarted, the nagios server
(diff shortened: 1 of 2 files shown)


wallyworld (approved):
Thanks.

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -337,10 +337,8 @@ def write(self):
                 "command": nrpecheck.command,
             }
             # If we were passed max_check_attempts, add that to the relation data
-            try:
+            if nrpecheck.max_check_attempts:
```
ajkavanagh, line 340: This check is probably correct, but I just need to check that it's not legal for `max_check_attempts` to be `0` (or if that is being disallowed).  More formally, the code should probably read:

```python
if nrpecheck.max_check_attempts is not None:
    ...
```

i.e. explicitly check for the Optional `None` to indicate that it is disabled, prior to 'say' verifying that it is in the correct bounds (if `0` is not a valid number ... which say could be infinite checks??)

guoqiao: updated, thanks.

ajkavanagh (approved):
LGTM from a code perspective.

guoqiao:
Hi openstack team,
Can we get this merged, please ?

PR #566, opened 2021-02-03
Supporting application level relation data

xtrusia:
Adding app parameter to relation_get and relation_set
to support application level relation data in the future


charmhelpers/core/hookenv.py
```diff
@@ -468,13 +468,16 @@ def config(scope=None):
...
+    if app is True:
+        _args.append('--app')
     if rid:
         _args.append('-r')
         _args.append(rid)
     _args.append(attribute or '-')
+    # unit or application name
     if unit:
         _args.append(unit)
     try:
```
johnsca, line 483: I know this convention matches the underlying `relation-get` call, but it feels really awkward.  I'm fine w/ this as-is, but I'd also like to put forth this suggested alternative UX:

```suggestion
def relation_get(attribute=None, unit=None, rid=None, app=None):
    """Get relation information"""
    _args = ['relation-get', '--format=json']
    if app is not None:
        _args.append('--app')
    if rid:
        _args.append('-r')
        _args.append(rid)
    _args.append(attribute or '-')
    # unit or application name
    if unit or app:
        _args.append(unit or app)
    try:
```

Thus, you could call it like it this:

```python
relation_get(rid=rid, app='foo')
relation_get(rid=rid, unit='foo/0')
```

johnsca (changes requested):
See in-line comment, which you can accept or reject, but I would like to see test coverage of this added [around here](https://github.com/juju/charm-helpers/blob/master/tests/core/test_hookenv.py#L800).

johnsca (commented):
Thanks, LGTM but I would still like to see test coverage added.

PR #567, opened 2021-02-05
add function to retrieve version of installed package

hemanthnakkina:
Currently version of installed package can be retrieved by using
apt_cache[pkg] but this takes longer execution time due to apt-cache
show --no-all-versions call [1]
To reduce the execution time, new function get_installed_version is
added which retrieves the information using dpkg-query without
building the Package object

[1] https://github.com/hemanthnakkina/charm-helpers/blob/master/charmhelpers/fetch/ubuntu_apt_pkg.py#L79

Partial-Bug: #1913992


pponnuvel (approved):
LGTM

javacruft (approved):

PR #568, opened 2021-02-09
Use HA peers as memcache cluster

nicolasbock:
Charms using `MemcacheContext`deploy local `memcached` services on
each unit but are not utilizing a `oslo_cache.memcache_pool`. This can
lead to inconsistently cached values and consequently to incorrect
query results.

Closes-Bug: #1771114

Signed-off-by: Nicolas Bock <nicolas.bock@canonical.com>

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -73,7 +73,6 @@ from charmhelpers.core.host import (
     write_file,
     pwgen,
     lsb_release,
-    CompareHostReleases,
     is_container,
 )
 from charmhelpers.contrib.hahelpers.cluster import (
@@ -2116,6 +2115,7 @@ class MemcacheContext(OSContextGenerator):
     This context provides options for configuring a local memcache client and
     server for both IPv4 and IPv6
     """
+    interfaces = ['memcache']
 
     def __init__(self, package=None):
         """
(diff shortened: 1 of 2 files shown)


nicolasbock:
@wolsen Could you have a quick peek at this PR? It's working in my test environment and fixes the issue described in https://bugs.launchpad.net/charm-keystone/+bug/1771114. Thanks!

charmhelpers/contrib/openstack/context.py
```diff
@@ -2129,29 +2129,17 @@ def __call__(self):
         ctxt = {}
         ctxt['use_memcache'] = enable_memcache(package=self.package)
         if ctxt['use_memcache']:
-            # Trusty version of memcached does not support ::1 as a listen
-            # address so use host file entry instead
-            release = lsb_release()['DISTRIB_CODENAME'].lower()
-            if is_ipv6_disabled():
```
nicolasbock, line 2135: The IPv6 context test is failing right now. I am going to work on fixing it.

charmhelpers/contrib/openstack/templates/memcached.conf
```diff
@@ -38,7 +38,7 @@ logfile /var/log/memcached.log
 # Specify which IP address to listen on. The default is to listen on all IP addresses
 # This parameter is one of the only security measures that memcached has, so make sure
 # it's listening on a firewalled interface.
--l {{ memcache_server }}
+-l {{ memcache_listen_address }}
```
ChrisMacNaughton, line 41: I'm fairly uncomfortable with this change. By doing this, the memcached instance is being opened to, potentially, the world.

A better solution, in my mind, would be to actually relate to a coordinator memcached[1] server to handle this, rather than make all units implicitly members of a globally usable memcached service.

[1]: Designate uses a [copordinator memcached](https://github.com/openstack/charm-designate/blob/master/src/tests/bundles/bionic-rocky.yaml#L68-L69) to handle sharing cache between units.

ChrisMacNaughton (changes requested):

charmhelpers/contrib/openstack/templates/memcached.conf
nicolasbock: Looking at how this is implemented I noticed that the `memcached` instances deployed by the memcached charm are just as open:

https://git.launchpad.net/charm-memcached/tree/templates/memcached.conf#n46

What am I missing?

(I am not opposed to your suggestion at all. I just would like to understand 😄 In fact I think to factor out the memcached logic into separate units is a great move regardless of the security implications).

ChrisMacNaughton: The memcached charm actually manages access control[]via ufw](https://git.launchpad.net/charm-memcached/tree/hooks/memcached_utils.py#n22) though, which the OpenStack Charm's internal memcached does not

nicolasbock: Oh nice! Thanks for pointing that out, I had missed that. I'll get to work then :smile:

nicolasbock: I have updated the change here and added a relation to the `memcached` charm. This change requires a change in `metadata.xml` on the charm side which I will submit separately.

nicolasbock: I will revert the changes to `memcached.conf` in another iteration of this PR.

nicolasbock:
@ChrisMacNaughton Could you have another look? I added a relation to `memcache` similar to what the Designate charm is doing. I also modified the unit tests so that the memcache context is tested properly. In the process I removed the IPv6 part since in its current form it's not testing whether there is an IPv6 network on which the memcached instances can be reached. Thanks!

charmhelpers/contrib/openstack/context.py
```diff
@@ -2129,29 +2129,17 @@ def __call__(self):
         ctxt = {}
         ctxt['use_memcache'] = enable_memcache(package=self.package)
```
ChrisMacNaughton, line 2130: if I'm reading this right, this change still installs memcached locally on the unit, and then tries to use the remote cluster?

Also, what happens in the non-HA setup where a remote memcached is unnecessary? Does this change allow the charm to use a local memcached when there is only a single unit but be added to a remote memcached for better HA?

ChrisMacNaughton (changes requested):

charmhelpers/contrib/openstack/context.py
nicolasbock: Yes, that's correct. To be honest I am not quite sure how to address this, since I found that the Designate charm uses two `memcached` instances, one that is using the `memcached` relation (for `[coordination]`) and one that isn't (for `[keystone_authotoken]`). I have to go through this code a bit more carefully.

I am not sure I understand your second comment. If the charm requires the `memcache` relation then the operator needs to deploy a `memcached` unit. So in even in a non-HA setup we end up with a remote `memcached` instance.

Thanks!

dosaboy: Could we perhaps make this conditional on whether the relation actually exists and if not just revert back to the local memcache method?

dosaboy:
For the reason explained in https://bugs.launchpad.net/charm-keystone/+bug/1771114/comments/8 we are wont use this approach anymore. ill go ahead and close this proposal and if we decide to revisit this approach later we can always reopen it.

PR #569, opened 2021-02-12
Deprecate local_address

thedac:
network-get juju-info is likely to alphanumerically sort addresses and
return the first rather than the "default" binding in the bundle.
Replace all calls to local_address with network-get and an explicit
binding.

Partial-Bug: #1915578

### DIFF
--- a/charmhelpers/contrib/openstack/ip.py
+++ b/charmhelpers/contrib/openstack/ip.py
@@ -15,6 +15,7 @@
 from charmhelpers.core.hookenv import (
     NoNetworkBinding,
     config,
+    log,
     unit_get,
     service_name,
     network_get_primary_address,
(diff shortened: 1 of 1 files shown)


charmhelpers/contrib/openstack/ip.py
```diff
@@ -139,9 +142,21 @@ def local_address(unit_get_fallback='public-address'):
     :returns: IP Address
     :rtype: str
     """
+    log(
+        "local_address is deprecated. Please find the call and replace it with "
+        "a network-get call with an explicit binding.", "WARNING")
```
ajkavanagh, line 147: If you wanted to be REALLY interesting/nasty here, you could use `trace_back` and print the stack at this point to tell them where the call is :)  It might make the logs a bit messy, but would highlight the issue.

thedac:
Out of date

PR #572, opened 2021-02-18
Extend restart_on_changed_helper

gnuoy:


charmhelpers/core/host.py
```diff
@@ -730,37 +730,77 @@ def wrapped_f(*args, **kwargs):
...
     @param lambda_f: function to call.
     @param restart_map: {file: [service, ...]}
     @param stopstart: whether to stop, start or restart a service
     @param restart_functions: nonstandard functions to use to restart services
                               {svc: func, ...}
+    @param restart_gate_f: A function used to check if the restart is
+                            permitted.
+    @param post_svc_restart_f: A function run after a service has
+                               restarted.
+    @param pre_restarts_wait_f: A function callled before any restarts.
```
ajkavanagh, line 771: I'd switch these all (and the previous ones) to `:param xyz:` and `:type xyz:` so that the functions can be declared.  e.g. `:type restart_gate_f` Callable[[str], boolean]`.  Also the name (I think) would be better as `can_restart_now_f`  .. i.e. it's used as a inverted check (returns False if you can't restart)?

charmhelpers/core/host.py
```diff
@@ -730,37 +730,77 @@ def wrapped_f(*args, **kwargs):
...
+    changed_files = []
+    restarts = []
     # create a list of lists of the services to restart
-    restarts = [restart_map[path]
-                for path in restart_map
-                if path_hash(path) != checksums[path]]
+    for path in restart_map:
+        if path_hash(path) != checksums[path]:
+            restarts.append(restart_map[path])
+            changed_files.append(path)
```
ajkavanagh, line 784: There's a bit of awkwardness in getting a simple list of changed files here.  Why not a `defaultdict[list]` where the key is the service_name.  e.g.:

```python
restarts = []
changed_files = defaultdict(list)
for path, service in restart_map.items():
    if path_hash(path) != checksums[path]:
        restarts.append(service)
        changed_files[service].append(path)

# and then later

if restart_gate_f:
     if not restart_gate_f(service_name, changed_files[service_name]):
        continue
```

This way you don't have to re-find the path <-> service mappings again?

charmhelpers/core/host.py
```diff
@@ -730,37 +730,84 @@ def wrapped_f(*args, **kwargs):
...
+    changed_files = defaultdict(list)
+    restarts = []
     # create a list of lists of the services to restart
-    restarts = [restart_map[path]
-                for path in restart_map
-                if path_hash(path) != checksums[path]]
+    for path in restart_map:
+        if path_hash(path) != checksums[path]:
+            restarts.append(restart_map[path])
+            changed_files[service].append(path)
```
ajkavanagh, line 792: I don't think this is quite correct.  I think it might/should be:

```python
for path, service in restart_map.items():
    if path_hash(path) != checksums[path]:
        restarts.append(service)
        changed_files[service].append(path)
```

e.g. the `for path, service in restart_map.items()` bit is missing.

ajkavanagh (commented):
Please see inline comment.

charmhelpers/core/host.py
gnuoy: I'm probably missing something obvious here but it looks to me that what I'm proposing matches the behaviour of the previous dict comprehension.

gnuoy: ah, yes, hurrah for pier review

ajkavanagh (approved):
Looks good to me; do we want to have a test run where it is used prior to merging it?

gnuoy:
> Looks good to me; do we want to have a test run where it is used prior to merging it?

I have a test here *1 which uses this branch (along with other changes)

*1 https://openstack-ci-reports.ubuntu.com/artifacts/test_charm_pipeline_func_smoke/openstack/charm-neutron-openvswitch/777568/2/22160/consoleText.test_charm_func_smoke_21453.txt

PR #581, opened 2021-03-01
Fix error log notification of openstack-release pkg

ajkavanagh:
Basically, kill the error log messsage for the openstack-release pkg
being missing, which appears due to the attempt to install it.  However,
it's not actually a bug if it isn't there (yet).

Also, add some docstrings to the functions so that future travellers can
understand what is going on.

### DIFF
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -483,9 +483,27 @@ def get_swift_codename(version):
     return None
 
 
-@deprecate("moved to charmhelpers.contrib.openstack.utils.get_installed_os_version()", "2021-01", log=juju_log)
 def get_os_codename_package(package, fatal=True):
-    '''Derive OpenStack release codename from an installed package.'''
+    """Derive OpenStack release codename from an installed package.
+
+    Initially, see of the openstack-release pkg is available (by trying to
+    install it) and use it instead.
+
+    If it isn't then it falls back to the existing method of checking the
+    version of the package passed and then resolving the version from that
+    using lookup tables.
+
+    Note: if possible, charms should use get_installed_os_version() to
+    determine the version of the "openstack-release" pkg.
+
+    :param package: the package to test for version information.
+    :type package: str
+    :param fatal: If True (default), then die via error_out()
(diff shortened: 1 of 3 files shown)


ChrisMacNaughton (dismissed):
Looks reasonable to me

charmhelpers/contrib/openstack/utils.py
```diff
@@ -483,9 +483,27 @@ def get_swift_codename(version):
     return None
 
 
-@deprecate("moved to charmhelpers.contrib.openstack.utils.get_installed_os_version()", "2021-01", log=juju_log)
 def get_os_codename_package(package, fatal=True):
-    '''Derive OpenStack release codename from an installed package.'''
+    """Derive OpenStack release codename from an installed package.
+
+    Initially, see of the openstack-release pkg is available (by trying to
```
AurelienLourot, line 489: s/of/if/

charmhelpers/contrib/openstack/utils.py
```diff
@@ -483,9 +483,27 @@ def get_swift_codename(version):
...
+    version of the package passed and then resolving the version from that
+    using lookup tables.
+
+    Note: if possible, charms should use get_installed_os_version() to
+    determine the version of the "openstack-release" pkg.
+
+    :param package: the package to test for version information.
+    :type package: str
+    :param fatal: If True (default), then die via error_out()
+        be determined.
```
AurelienLourot, line 502: It may be my English, but I don't understand "die via error_out() be determined". Is a word missing?

AurelienLourot (changes requested):
LGTM, just two nitpicks, thanks!

charmhelpers/contrib/openstack/utils.py
ajkavanagh: Looks like a copy-paste error type thing - I'll fix it.

charmhelpers/contrib/openstack/utils.py
ajkavanagh: Good catch

AurelienLourot (approved):

PR #582, opened 2021-03-02
Add hirsute and wallaby identifiers for OpenStack

ajkavanagh:
This patch adds the hirsute to wallaby mapping for 'distro' for hirsute
and the wallaby version numbers for the core packages.

### DIFF
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -151,6 +151,7 @@ OPENSTACK_RELEASES = (
     'train',
     'ussuri',
     'victoria',
+    'wallaby',
 )
 
 UBUNTU_OPENSTACK_RELEASE = OrderedDict([
@@ -173,6 +174,7 @@ UBUNTU_OPENSTACK_RELEASE = OrderedDict([
     ('eoan', 'train'),
     ('focal', 'ussuri'),
     ('groovy', 'victoria'),
+    ('hirsute', 'wallaby'),
 ])
 
 
@@ -196,6 +198,7 @@ OPENSTACK_CODENAMES = OrderedDict([
     ('2019.2', 'train'),
     ('2020.1', 'ussuri'),
     ('2020.2', 'victoria'),
+    ('2021.1', 'wallaby'),
 ])
 
 # The ugly duckling - must list releases oldest to newest
@@ -238,6 +241,8 @@ SWIFT_CODENAMES = OrderedDict([
         ['2.24.0', '2.25.0']),
     ('victoria',
         ['2.25.0', '2.26.0']),
+    ('wallaby',
+        ['2.26.0']),
 ])
 
 # >= Liberty version->codename mapping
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/utils.py
```diff
@@ -235,6 +238,8 @@
         ['2.24.0', '2.25.0']),
     ('victoria',
         ['2.25.0', '2.26.0']),
+    ('wallaby',
```
ChrisMacNaughton, line 241: The point of the new openstack-release package is that this should be unnecessary. on Hirsute+, the openstakc-package should be both available and installed on all openstack installs, and should tell is that this is running Wallaby

charmhelpers/contrib/openstack/utils.py
```diff
@@ -277,6 +284,7 @@
         ('15', 'train'),
         ('16', 'ussuri'),
         ('17', 'victoria'),
+        ('18', 'cinder'),
```
ChrisMacNaughton, line 287: Typo, but also see comment about new package

ChrisMacNaughton (changes requested):

charmhelpers/contrib/openstack/utils.py
ajkavanagh: But we still need this to work for focal, don't we (I mean right NOW).

charmhelpers/contrib/openstack/utils.py
ajkavanagh: Yeash, yes, definitely a typo!

charmhelpers/contrib/openstack/utils.py
ChrisMacNaughton: On focal-wallaby, the package should be present _in the cloud-archive_ for Wallaby.

ChrisMacNaughton:
```
$ cmadison openstack-release
 openstack-release | 1.0ubuntu1~cloud0 | wallaby-proposed | focal-proposed | source
 ```

ajkavanagh: I still wonder, for completeness, whether this should be added?

ajkavanagh:
@ChrisMacNaughton are we in the "I don't think we need this PR" at the moment state?

coreycb (changes requested):
Do we still need all of these changes for wallaby+ now that we have the new openstack-release package?

charmhelpers/contrib/openstack/utils.py
coreycb: For the dicts we no longer need to update (ie. the ones that openstack-release replaces) I think we should add a comment for wallaby+ to say that openstack-release package determines the openstack code name.

coreycb:
> @ChrisMacNaughton are we in the "I don't think we need this PR" at the moment state?

We may only need the change to CLOUD_ARCHIVE_POCKETS in this PR.

ChrisMacNaughton:
> (update: please ignore - looks like the comment was already made above) Do we still need all of these changes for wallaby+ now that we have the new openstack-release package?

We do need to learn about the Wallaby codename, we just don't need the version number -> codename mapping that was originally in this change.

ChrisMacNaughton (approved):
LGTM

AurelienLourot (approved):

coreycb (approved):

PR #584, opened 2021-03-04
Add helper for sequencing charm status checks

przemeklal:
This helper allows to sequence status check functions so that they all
get a chance to run.

Signed-off-by: Przemysław Lal <przemyslaw.lal@canonical.com>

### DIFF
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -2474,3 +2474,26 @@ def get_api_application_status():
             msg = 'Some units are not ready'
     juju_log(msg, 'DEBUG')
     return app_state, msg
+
+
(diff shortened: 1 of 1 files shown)


przemeklal:
This code is already used here: https://github.com/openstack/charm-neutron-gateway/blob/master/hooks/neutron_utils.py#L1161
Now that the same behavior is being reimplemented for charm-neutron-openvswitch, it makes sense to move this code to charm-helpers to avoid duplication.

coreycb (approved):
LGTM

ajkavanagh:
@przemeklal please could you do a follow-up PR for this that adds tests for this.  I can't decide whether to revert this or ask for a new PR?  Which would you prefer?

przemeklal:
I'll follow up with a new PR.

PR #587, opened 2021-03-11
Decorator/context manager version of restart_on_change

ajkavanagh:
This is the start of an example of how to turn the restart_on_change
decorator into a context manager and decorator.


ajkavanagh:
Closing as it's probably out of time; charm-helpers is now a deprecated library as everything Juju has/is moving to the ops framework that doesn't use charm-helpers.

PR #588, opened 2021-03-11
Add 'heartbeat_timeout_threshold` option

nicolasbock:
This change adds the `heartbeat_timeout_threshold` option to the
`oslo_messaging_rabbit` section.

Closes-Bug: #1918673

Signed-off-by: Nicolas Bock <nicolas.bock@canonical.com>


charmhelpers/contrib/openstack/context.py
```diff
@@ -768,6 +768,8 @@ def __call__(self):
         if send_notifications_to_logs:
             ctxt['send_notifications_to_logs'] = send_notifications_to_logs
 
+        ctxt['rabbit_heartbeat_timeout_threshold'] = conf.get('rabbit_heartbeat_timeout_threshold')
```
freyes, line 771: in the charm's configuration we should use `rabbit-heartbeat-timeout-threshold` to keep the consistency.

freyes (changes requested):

charmhelpers/contrib/openstack/context.py
nicolasbock: done

freyes:
LGTM.

charmhelpers/contrib/openstack/context.py
```diff
@@ -768,6 +768,8 @@ def __call__(self):
         if send_notifications_to_logs:
             ctxt['send_notifications_to_logs'] = send_notifications_to_logs
 
+        ctxt['rabbit_heartbeat_timeout_threshold'] = conf.get('rabbit-heartbeat-timeout-threshold')
```
ajkavanagh, line 771: All of the above versions of conf.get() include the default 'None' parameter.  For consistency, we should probably include that, please.

ajkavanagh (changes requested):
Thanks for the patch; please see the comment about the default value of None.  Thanks.

charmhelpers/contrib/openstack/context.py
nicolasbock: done

gnuoy:
Looks like unit tests need updating

PR #590, opened 2021-03-19
contrib/openstack: Add helpers for principal subordinate package information exchange

fnordahl:
For principal - subordinate plugin type relations where the
principal Python payload imports code from packages managed by a
subordinate, upgrades can be problematic.

This change will allow a subordinate charm that have opted into the
feature to inform its principal about all implemented release -
packages combinations ahead of time. With this information in place
the principal can do the upgrade in one operation without risk of
charm relation RPC type processing at a critical moment.

Related-Bug: #1806111

### DIFF
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -2201,6 +2201,23 @@ def container_scoped_relations():
     return relations
 
 
+def container_scoped_relation_get(attribute=None):
+    """Get relation data from all container scoped relations.
+
+    :param attribute: Name of attribute to get
+    :type attribute: Optional[str]
+    :returns: Iterator with relation data
+    :rtype: Iterator[Optional[any]]
+    """
+    for endpoint_name in container_scoped_relations():
+        for rid in relation_ids(endpoint_name):
+            for unit in related_units(rid):
+                yield relation_get(
+                    attribute=attribute,
+                    unit=unit,
+                    rid=rid)
+
+
 def is_db_ready(use_current_context=False, rel_name=None):
     """Check remote database is ready to be used.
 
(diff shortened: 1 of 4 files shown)


charmhelpers/contrib/openstack/utils.py
```diff
@@ -2497,3 +2514,33 @@ def _inner_sequenced_functions(configs):
...
+    :returns: Packages to install and packages to purge or None
+    :rtype: SubordinatePackages[set,set]
+    """
+    install = set()
+    purge = set()
+
+    for rdata in container_scoped_relation_get('releases-packages-map'):
+        rp_map = json.loads(rdata or '{}')
+        for release in sorted(rp_map.keys(), reverse=True):
+            if (CompareOpenStackReleases(release) <= os_release and
```
AurelienLourot, line 2538: I think this deserves a comment because the sorting with reverse=True and the comparison here aren't crystal-clear. Are we taking here the highest subordinate release smaller or equal the principal release? or are we taking the smallest subordinate release greater or equal the principal release? Also why? (I can't really decide what's best myself and why) thanks :)

EDIT: I have now read the unit tests, so I now understand that this will give the most recent earlier release info (i.e. what I called "the highest subordinate release smaller or equal the principal release") for each subordinate. But I'm not sure why this is good. Why not taking the exact same release info, and if not available, we would end up waiting for the subordinate to upgrade? I'm sure if I read through the linked bug and wrap my head around it I will understand but maybe this can be summed up in a few words in the code here? Thanks for your patience

AurelienLourot (changes requested):
This looks good, thanks! I think we're one inline-comment away from having it merged :)

charmhelpers/contrib/openstack/utils.py
AurelienLourot: oh I think I know: say the subordinate is implemented today and gives info up to wallaby. then it works so well that we don't touch it for cycle X, we want the principle of release X to still regard the information provided for wallaby as valid. Am I right?

fnordahl: In short, yes. Just like with the selection of templates or charm implementation, if nothing has changed since past release R we still want to use the information from past release R now. And since you're inclined to ask the question that is justification enough for me to add a comment, coming up.

charmhelpers/contrib/openstack/reactive.py
```diff
@@ -0,0 +1,38 @@
...
+"""Library of base classes for common reactive endpoints and relations."""
+
+import json
+
+import charms.reactive as reactive
+
+import charmhelpers.core.hookenv as hookenv
+
+
+class OpenStackRelationBase(reactive.RelationBase):
```
thedac, line 23: Why RelationBase rather than Endpoint?

thedac (changes requested):
Charmhelpers really *really* feels like the wrong place for this. Wouldn't a new interface or at the very least charms.openstack be a better place for this code?

I also don't think naming it reactive.py is a good idea. 

I would love to discuss this further. I have more to review but I wanted to get this out before someone lands this.

charmhelpers/contrib/openstack/reactive.py
fnordahl: The change is to be used with the [keystone-domain-backend](https://github.com/openstack/charm-interface-keystone-domain-backend) and [cinder-backend](https://github.com/openstack/charm-interface-cinder-backend) interfaces which both are RelationBase interfaces. As soon as we need to use this with Endpoint-based interfaces we should add a separate definition for that.

fnordahl:
> Charmhelpers really _really_ feels like the wrong place for this. Wouldn't a new interface or at the very least charms.openstack be a better place for this code?

So the goal is to have somewhere to host base classes for interface code, that be `RelationBase`, `Endpoint` or what we may find around the next corner.

The code is relevant for multiple interfaces, any interface used for a subordinate relation could make use of it. Adding a separate endpoint+interface for this is not an option imho.

> I also don't think naming it reactive.py is a good idea.
> 
> I would love to discuss this further. I have more to review but I wanted to get this out before someone lands this.

Discussion and suggestions are welcome!

AurelienLourot (approved):
My comment has been addressed, thanks! Not landing just yet: waiting for the ongoing design conversation to settle.

fnordahl:
As discussed offline let's remove the interface class from c-h for now, a compromise could be to add it to layer-openstack, but for the time being we'll just stash it in the specific interface code.

thedac (approved):
Looks good.

PR #591, opened 2021-03-26
Enable pausable_restart_on_change as cntxt manager

gnuoy:
Enable pausable_restart_on_change to be used as both
a decorator and context manager.


gnuoy:
PR to test no regression in using it as a decorator ttps://review.opendev.org/c/openstack/charm-glance/+/783354

AurelienLourot:
@gnuoy the [associated Gerrit review](https://review.opendev.org/c/openstack/charm-glance/+/783354) was abandoned as "not needed". Does it mean we don't need this change anymore? Or does it mean that the associated review did its job at validating this work and it's still wanted?

Issue #367, opened 2019-08-27
Allow custom environment for jinja templates


chris-sanders:
Jinja allows for several customization options when creating the environment to processes templates. Specifically being able to change the start/end characters for variables and blocks is needed when processing templates of templates and should be exposed in the template module.

mkalcok:
(@chris-sanders) I think that with successful merge of #592 . This issue can be closed.

auria:
@ajkavanagh FYI, I think this issue can be closed. A PR was merged a few back, thanks!

ajkavanagh:
As advised by @auria I'm closing this issue.

PR #592, opened 2021-03-29
Allow passing additional jinja2.Environment variables to `render` function

mkalcok:
This should solve #367

### DIFF
--- a/charmhelpers/contrib/templating/jinja.py
+++ b/charmhelpers/contrib/templating/jinja.py
@@ -31,8 +31,21 @@ except ImportError:
 DEFAULT_TEMPLATES_DIR = 'templates'
 
 
-def render(template_name, context, template_dir=DEFAULT_TEMPLATES_DIR):
+def render(template_name, context, template_dir=DEFAULT_TEMPLATES_DIR,
+           jinja_env_args=None):
+    """
+    Render jinja2 template with provided context.
+
+    :param template_name: name of the jinja template file
+    :param context: template context
+    :param template_dir: directory in which the template file is located
+    :param jinja_env_args: additional arguments passed to the
+                           jinja2.Environment. Expected dict with format
+                           {'arg_name': 'arg_value'}
+    :return: Rendered template as a string
+    """
+    env_kwargs = jinja_env_args or {}
     templates = jinja2.Environment(
-        loader=jinja2.FileSystemLoader(template_dir))
+        loader=jinja2.FileSystemLoader(template_dir), **env_kwargs)
     template = templates.get_template(template_name)
     return template.render(context)
--- a/tests/contrib/templating/test_jinja.py
+++ b/tests/contrib/templating/test_jinja.py
@@ -12,6 +12,8 @@ SIMPLE_TEMPLATE = "{{ somevar }}"
 
 LOOP_TEMPLATE = "{% for i in somevar %}{{ i }}{% endfor %}"
 
+CUSTOM_DELIM_TEMPLATE = "{{ not_var }} << template_var >>"
+
 
 class Jinja2Test(TestCase):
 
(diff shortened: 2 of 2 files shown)


charmhelpers/contrib/templating/jinja.py
```diff
@@ -31,8 +31,21 @@
 DEFAULT_TEMPLATES_DIR = 'templates'
 
 
-def render(template_name, context, template_dir=DEFAULT_TEMPLATES_DIR):
+def render(template_name, context, template_dir=DEFAULT_TEMPLATES_DIR,
+           jinja_env_args=None):
```
auria, line 35: Change lgtm. I wonder if `jinja_env_var=None` could be replaced by `jinja_env_args={}` or directly by `**env_kwargs`. On line 47, there is no input validation so the behavior would be the same if the dict key does not exist in `jinja2.Environment`. The way params are passed to this `render(...)` helper would also change.

This is not a blocker to me, though.

ajkavanagh: This would be a really *bad* idea.  Setting the optional parameter to `None` **is** the pattern to indicate that nothing has been passed.

The reason is that Python passes parameters by reference, if if the default for a parameter is a boxed (i.e. a reference to a type) like `Dict` or `List` then *only on the first call* is the default param initialised.  Thus, it acts like a local cache and is the same dictionary on each invocation:

```
$ ipython3
Python 3.8.5 (default, Jan 27 2021, 15:41:15)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: def a(x, y={}):
   ...:     y[x] = 1
   ...:     print(y)
   ...:

In [2]: a('hello')
{'hello': 1}

In [3]: a('goodbye')
{'hello': 1, 'goodbye': 1}
```

So this is that correct pattern.

mkalcok: I was thinking about both of those approaches.

`jinja_env_args={}`. I generally try to avoid using mutable default argument values even in cases like this where it would be theoretically safe. (@ajkavanagh has more thorough explanation, thanks)

`**env_kwargs`. I thought that this could potentially create some confusion because `Environment.render()` (for which this function is basically a wrapper) also takes `**kwargs` but for a different purpose and I can see how someone could think that `**kwargs` in `charmhelpers.render` and `jinja.render` behave the same. I don't know, maybe I'm overthinking this one...

ajkavanagh: (@mkalcok sorry for hijacking your review ...)

Just to follow-up on my comment, the initialization of the defaults happens at *function definition* not at first call, like I erroneously said.  This means that expensive defaults happen when the *module is loaded*:

```python
In [13]: class B(): 
    ...:     def __init__(self): 
    ...:         self.x = 5 
    ...:         print("hi from B() init") 
    ...:          
    ...:                                                                                        

In [14]: def b(x=B()): 
    ...:     print("hi from b() function") 
    ...:                                                                                        
hi from B() init

In [15]: b()                                                                                    
hi from b() function
```

But there's also a really important point here.  As python passes *everything* by reference, any type with interior mutability (e.g. Dict, List, instantiated objects) is vulnerable to both the "default kwarg" problem, but also the "in-out" return problem.  "Simple" types like ints don't have interior mutability and, thus, aren't susceptible.

So the pattern, generally, when accepting an interior-mutable type into a function that that function is going to modify is to make a copy of it, so as not to contaminate the caller's 'version' of the variable.  e.g.

```python
def f(some_list, some_dict):
    my_list = some_list[:]
    my_list.append(something)
    my_dict = some_dict.copy()  # shallow copy only!
    my_dict['extra'] = thing
    # etc.
```

For classes, it's a bit muddier (as there's no default 'copy' for them), but the strategy/pattern should always be "avoid modifying the callers parameters".

Again, sorry for hijacking the thread, but just wanted to be clear.  Thanks!

auria: Cristal clear. +1 on the approach.

ajkavanagh (approved):
LGTM

PR #593, opened 2021-03-29
Add code to manage deferred hooks

gnuoy:
Add code to manage deferred hooks and is_hook_allowed to
combine checks that restarts are allowed and pause is not set.


thedac (approved):
This looks great.

PR #594, opened 2021-03-31
Add helper for dpdk pmd cpu mask

baycarbone:
Added helper to convert a cpu set in list format into a hex formatted
cpu mask for pmd-cpu-mask and added associated unit test.

Added support for caret operator in _parse_cpu_list(cpulist) and updated
associated unit test.

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -2560,14 +2560,18 @@ class OVSDPDKDeviceContext(OSContextGenerator):
         :rtype: List[int]
         """
         cores = []
-        ranges = cpulist.split(',')
-        for cpu_range in ranges:
-            if "-" in cpu_range:
-                cpu_min_max = cpu_range.split('-')
-                cores += range(int(cpu_min_max[0]),
-                               int(cpu_min_max[1]) + 1)
-            else:
-                cores.append(int(cpu_range))
+        if cpulist and re.match(r"^[0-9,\-^]*$", cpulist):
+            ranges = cpulist.split(',')
+            for cpu_range in ranges:
+                if "-" in cpu_range:
+                    cpu_min_max = cpu_range.split('-')
+                    cores += range(int(cpu_min_max[0]),
+                                   int(cpu_min_max[1]) + 1)
+                elif "^" in cpu_range:
+                    cpu_rm = cpu_range.split('^')
+                    cores.remove(int(cpu_rm[1]))
+                else:
+                    cores.append(int(cpu_range))
         return cores
 
     def _numa_node_cores(self):
@@ -2586,36 +2590,32 @@ class OVSDPDKDeviceContext(OSContextGenerator):
 
     def cpu_mask(self):
         """Get hex formatted CPU mask
-
         The mask is based on using the first config:dpdk-socket-cores
         cores of each NUMA node in the unit.
         :returns: hex formatted CPU mask
         :rtype: str
         """
-        return self.cpu_masks()['dpdk_lcore_mask']
-
-    def cpu_masks(self):
-        """Get hex formatted CPU masks
-
-        The mask is based on using the first config:dpdk-socket-cores
-        cores of each NUMA node in the unit, followed by the
-        next config:pmd-socket-cores
-
-        :returns: Dict of hex formatted CPU masks
-        :rtype: Dict[str, str]
-        """
-        num_lcores = config('dpdk-socket-cores')
-        pmd_cores = config('pmd-socket-cores')
-        lcore_mask = 0
-        pmd_mask = 0
+        num_cores = config('dpdk-socket-cores')
+        mask = 0
         for cores in self._numa_node_cores().values():
-            for core in cores[:num_lcores]:
-                lcore_mask = lcore_mask | 1 << core
-            for core in cores[num_lcores:][:pmd_cores]:
-                pmd_mask = pmd_mask | 1 << core
-        return {
-            'pmd_cpu_mask': format(pmd_mask, '#04x'),
-            'dpdk_lcore_mask': format(lcore_mask, '#04x')}
+            for core in cores[:num_cores]:
+                mask = mask | 1 << core
+        return format(mask, '#04x')
+
+    @classmethod
+    def pmd_cpu_mask(cls):
+        """Get hex formatted pmd CPU mask
+
+        The mask is based on config:pmd-cpu-set.
+        :returns: hex formatted CPU mask
+        :rtype: str
+        """
+        mask = 0
+        cpu_list = cls._parse_cpu_list(config('pmd-cpu-set'))
+        if cpu_list:
+            for core in cpu_list:
+                mask = mask | 1 << core
+        return format(mask, '#x')
 
     def socket_memory(self):
         """Formatted list of socket memory configuration per socket.
@@ -2694,6 +2694,7 @@ class OVSDPDKDeviceContext(OSContextGenerator):
             ctxt['device_whitelist'] = self.device_whitelist()
             ctxt['socket_memory'] = self.socket_memory()
             ctxt['cpu_mask'] = self.cpu_mask()
+            ctxt['pmd_cpu_mask'] = self.pmd_cpu_mask()
         return ctxt
 
 
--- a/tests/contrib/openstack/test_os_contexts.py
+++ b/tests/contrib/openstack/test_os_contexts.py
@@ -4431,6 +4431,10 @@ class MockPCIDevice(object):
 TEST_CPULIST_1 = "0-3"
 TEST_CPULIST_2 = "0-7,16-23"
 TEST_CPULIST_3 = "0,4,8,12,16,20,24"
+TEST_CPULIST_4 = "0-7,^5,16-23,^20,^22"
+TEST_CPULIST_5 = "0-7,^5,16-23,wrong,$;^20,^22"
+TEST_CPULIST_6 = "wrong:format"
+TEST_CPULIST_7 = "0-7;16-23"
 DPDK_DATA_PORTS = (
     "br-phynet3:fe:16:41:df:23:fe "
     "br-phynet1:fe:16:41:df:23:fd "
@@ -4555,6 +4559,15 @@ class TestOVSDPDKDeviceContext(tests.utils.BaseTestCase):
                           16, 17, 18, 19, 20, 21, 22, 23])
         self.assertEqual(self.target._parse_cpu_list(TEST_CPULIST_3),
                          [0, 4, 8, 12, 16, 20, 24])
+        self.assertEqual(self.target._parse_cpu_list(TEST_CPULIST_4),
+                         [0, 1, 2, 3, 4, 6, 7,
+                          16, 17, 18, 19, 21, 23])
+        self.assertEqual(self.target._parse_cpu_list(TEST_CPULIST_5),
+                         [])
+        self.assertEqual(self.target._parse_cpu_list(TEST_CPULIST_6),
+                         [])
+        self.assertEqual(self.target._parse_cpu_list(TEST_CPULIST_7),
+                         [])
 
     def test__numa_node_cores(self):
         self.patch_target('_parse_cpu_list')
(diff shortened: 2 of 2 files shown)


charmhelpers/contrib/openstack/context.py
```diff
@@ -2577,6 +2580,22 @@ def cpu_mask(self):
                 mask = mask | 1 << core
         return format(mask, '#04x')
 
+    def pmd_cpu_mask(self):
```
ajkavanagh, line 2583: This looks like a `@classmethod`?  i.e. it doesn't seem to use any of the instance members of the class.  As a class method it would access `_parse_cpu_list()` via the `cls` method param (instead of `self`).  Thanks.

baycarbone: updated PR with @classmethod and use of cls.

charmhelpers/contrib/openstack/context.py
```diff
@@ -2577,6 +2580,23 @@ def cpu_mask(self):
                 mask = mask | 1 << core
         return format(mask, '#04x')
 
+    @classmethod
+    def pmd_cpu_mask(cls):
+        """Get hex formatted pmd CPU mask
+
+        The mask is based on config:pmd-cpu-set.
+        :returns: hex formatted CPU mask
+        :rtype: str
```
fnordahl, line 2589: The function as presented actually has rtype `Optional[str]`, however, do bear in mind that returning that `None` has high risk of blowing up somewhere else, so I'd consider an empty string instead if that makes sense.

tests/contrib/openstack/test_os_contexts.py
```diff
@@ -4395,6 +4395,7 @@ def __init__(self, address):
 TEST_CPULIST_1 = "0-3"
 TEST_CPULIST_2 = "0-7,16-23"
 TEST_CPULIST_3 = "0,4,8,12,16,20,24"
+TEST_CPULIST_4 = "0-7,^5,16-23,^20,^22"
```
fnordahl, line 4398: Would it also make sense to add some not well formed strings and test if the code behaves as you would expect?

fnordahl (changes requested):
Thank you for your work on this, it looks pretty good and I have a couple of comments in-line.

tests/contrib/openstack/test_os_contexts.py
baycarbone: I can but I know it will not behave as expected ;-). I re-used _parse_cpu_list(cpulist) as is but I can make it more robust if needed.

ajkavanagh: > I can but I know it will not behave as expected ;-). I re-used _parse_cpu_list(cpulist) as is but I can make it more robust if needed.

The concepts of "assume all data is corrupt" and thus "parse and reject malformed data" come to mind here.  Fuzzy testing tends to break code for a reason :)  I agree with @fnordahl that some invalid strings should be used to test the effectiveness of parsing the data out of the string.

charmhelpers/contrib/openstack/context.py
fnordahl: That would be handled here:
https://github.com/openstack-charmers/charm-layer-ovn/blob/0898177a1f278cb83f911eac70e4a0af7a0dc120/lib/charms/ovn_charm.py#L518-L521

The charm code removes the key from the other_config field when no value is set.

In Python the Boolean check on a variable returns false both for `None` and `''` (the empty string)

ajkavanagh: Just to hijack @fnordahl 's comment a little bit, there *is* an important distinction between `None` (the (no) type) and other types (e.g. `str`).  It specifically is to indicate that *no value* was appropriate, and is intended to either blow up code that doesn't know this (i.e. as a library function return value), or (more correctly) to allow callers to detect the difference between a false-y value and no value.  i.e. we use:

```python
     if some_var is None:
        pass
```

to detect the latter state.

I'm not sure (in the context of this code), which is the better, but @fnordahl is correct in that the *docstring* is incorrect.  The return value *is* Optional[str].

tests/contrib/openstack/test_os_contexts.py
baycarbone: No problem, will do!

baycarbone:
made the changes but a small question before I push, should I rebase and squash my commits once the review process is done and everything looks good to merge or should I rebase and squash each time I update my PR?

charmhelpers/contrib/openstack/context.py
baycarbone: used the empty string as suggested by Frode thus kept return type to str.

charmhelpers/contrib/openstack/context.py
baycarbone: Changed it back as it now calls cpu_mask().

charmhelpers/contrib/openstack/context.py
baycarbone: I've changed the behavior in case the option is not set to instead use the dpdk-lcore-mask shifted by 1.

tests/contrib/openstack/test_os_contexts.py
baycarbone: Tests added with invalid data.

baycarbone:
found a bug, fixing it, then will update

charmhelpers/contrib/openstack/context.py
```diff
@@ -2562,6 +2566,18 @@ def _numa_node_cores(self):
                 nodes[index] = self._parse_cpu_list(cpulist.read().strip())
         return nodes
 
+    def _lowest_cpu_mask(self, num_cores=1):
```
baycarbone, line 2569: Extracted computation of lowest mask in order to be able to re-use it for the default mask generation of pmd-cpu-mask.

tests/contrib/openstack/test_os_contexts.py
```diff
@@ -4455,8 +4459,8 @@ def test_resolve_pci_from_mapping_config(self):
 }
 
 NUMA_CORES_MULTI = {
-    '0': [0, 1, 2, 3],
-    '1': [4, 5, 6, 7]
+    '0': [0, 2, 4, 6],
+    '1': [1, 3, 5, 7]
```
baycarbone, line 4463: I think the CPU list per NUMA in the test data was incorrect. Here is an example of CPU list with 2 NUMA nodes:
# numactl -H
available: 2 nodes (0-1)
node 0 cpus: 0 2 4 6 8 10 12 14
node 1 cpus: 1 3 5 7 9 11 13 15

tests/contrib/openstack/test_os_contexts.py
```diff
@@ -4549,12 +4562,54 @@ def test_cpu_mask(self):
...
 
         self._numa_node_cores.return_value = NUMA_CORES_MULTI
-        self.assertEqual(self.target.cpu_mask(), '0x11')
+        self.assertEqual(self.target.cpu_mask(), '0x03')
 
         self.config.side_effect = lambda x: {
             'dpdk-socket-cores': 2,
         }.get(x)
-        self.assertEqual(self.target.cpu_mask(), '0x33')
+        self.assertEqual(self.target.cpu_mask(), '0x0f')
```
baycarbone, line 4570: Corrected test based on correction to NUMA_CORES_MULTI

tests/contrib/openstack/test_os_contexts.py
```diff
@@ -4549,12 +4562,54 @@ def test_cpu_mask(self):
         self.assertEqual(self.target.cpu_mask(), '0x01')
 
         self._numa_node_cores.return_value = NUMA_CORES_MULTI
-        self.assertEqual(self.target.cpu_mask(), '0x11')
+        self.assertEqual(self.target.cpu_mask(), '0x03')
```
baycarbone, line 4565: Corrected test based on correction to NUMA_CORES_MULTI

baycarbone:
Here are manual functional tests results: https://docs.google.com/document/d/1sQDPx8ar_kjm15Qhe95rTxFw4OEli4z9dWaiyF4NZ30/edit?usp=sharing

charmhelpers/contrib/openstack/context.py
```diff
@@ -2577,6 +2581,26 @@ def cpu_mask(self):
...
+        cpu_list = self._parse_cpu_list(config('pmd-cpu-set'))
+        if cpu_list:
+            for core in cpu_list:
+                mask = mask | 1 << core
+        else:
+            # The default mask is based on the next cpu of each NUMA node
+            # following the dpdk-lcore-mask in order to avoid overlap.
+            num_cores = config('dpdk-socket-cores')
+            for cores in self._numa_node_cores().values():
+                mask = mask | 1 << cores[num_cores]
```
fnordahl, line 2601: Would you mind enlightening me the reasoning behind this being the best default behavior? Any reason why we could/should not use the [Open vSwitch default behavior](https://github.com/openvswitch/ovs/blob/f686957c9667ae962fb8fc003be2a5482e380d75/vswitchd/vswitch.xml#L338-L339) in the event of this value not being set by the end user of the charm?

fnordahl (commented):
Hello @baycarbone, I finally got back to looking at this again and I have questions.

charmhelpers/contrib/openstack/context.py
javacruft: I think I see the logic here - this ensures that the PMD and DPDK lcore don't overlap - the PMD is placed on the core after the allocation for the DPDK lcore

fnordahl: > I think I see the logic here - this ensures that the PMD and DPDK lcore don't overlap - the PMD is placed on the core after the allocation for the DPDK lcore

I don't disagree with that, my problem with the current proposal is that if the user does not specify this config the charm does not need to populate the OVS config at all, [OVS will create a PMD thread per NUMA node and pin it to a available core all on its own.](https://github.com/openvswitch/ovs/blob/5db8aa39d9b7f6317f0d450e503de53865fa5593/vswitchd/vswitch.xml#L354-L355)

Since the original author has apparently disappeared and we want this resolved I guess we just need to fork the change.

baycarbone: @fnordahl, I did not go away ;). In fact I had updated the initial [bug](https://bugs.launchpad.net/charm-ovn-chassis/+bug/1905416/comments/7) with additional information following this merged [PR](https://github.com/juju/charm-helpers/pull/538) which was following the same idea as this PR but was not compatible with my [layer-ovn PR](https://github.com/openstack-charmers/charm-layer-ovn/pull/41) which actually exposes the config-option to the ovn-chassis charm.

fnordahl: Excellent. So the thing is that I would still like this change to land.

While adding a automatic get-me-a-number-of-pmd cores option is great, the reality is that if you're interesting in tuning these parameters, you most likely will have very specific needs. I.e. the system will most likely have a set of CPU cores isolated, and you would want PMD threads spawned on those specific cores.

So I would very much like this PR to land as well. If you have a spare moment to update the offending section, i.e. don't set a mask if the config is not populated, I think we should be good to go.

baycarbone: Happy to make the change, though I now recall the reason behind this default behavior so will share it just in case:
In the [link](https://github.com/openvswitch/ovs/blob/5db8aa39d9b7f6317f0d450e503de53865fa5593/vswitchd/vswitch.xml#L354-L355) you mentioned it states:
> If not specified, one pmd thread will be created for each numa node
   and pinned to **_any_** available core on the numa node by default.

But I also came across this [link](https://developers.redhat.com/blog/2017/06/28/ovs-dpdk-parameters-dealing-with-multi-numa#dpdk_init) which mentions:
> OVS-DPDK performance typically scales well with additional cores, so it is common that users set pmd-cpu-mask but if it is not set, _**the default is that the lowest core on each NUMA node will be used for datapath packet processing**_.

As a result, if defaults are used for both dpdk-socket-cores and pmd-cpu-set, then you would end up with an overlap between dpdk-lcore-mask and pmd-cpu-mask (and the ovn-chassis charm would go into blocked state according to my layer-ovn PR).

fnordahl: Thanks for the update, in that case we should stop the lcore mask from being configured into OVS unless it's actually set by the end user and require both to be configured together when set. We'll deal with that in charm code

fnordahl: And to be clear, that means we want the code to handle a new `pmd-cpu-set` configuration option (i.e. this PR) to land, sans the default mask when config not set. I.e. if user does not provide config let OVS defaults (which may develop over time) run the show.

baycarbone: Change has been made, I've also removed the previous implementation brought by 3b7a86420629cc45ced7f78fa3706d99f514448e to avoid confusion. I checked the ovn and ovs charms and I didn't see that config option being used yet so I assumed it was safe to remove.

fnordahl (approved):
LGTM, thank you for working on this!

PR #595, opened 2021-04-06
Allow rbd mirroring mode to be set to none

wolsen:
Allow the rbd-mirroring-mode for a Ceph pool to be specified as
'none'. This allows a charm to be able to explicitly disable rbd
mirroring when it knows it will be creating a pool type which is
not supported.

The example for this, is when erasure coded pools are being created
for RBD use. Two pools are created, one for erasure coded data and
one for RBD metadata. Neither pool should be mirrored, so the charm
can explicitly specify this.

Closes-Bug: #1898504


AurelienLourot (approved):
LGTM, thanks! We need another approval.

PR #596, opened 2021-04-09
Allow rbd mirroring mode to be set to none

wolsen:
Allow the rbd-mirroring-mode for a Ceph pool to be specified as
'none'. This allows a charm to be able to explicitly disable rbd
mirroring when it knows it will be creating a pool type which is
not supported.

The example for this, is when erasure coded pools are being created
for RBD use. Two pools are created, one for erasure coded data and
one for RBD metadata. Neither pool should be mirrored, so the charm
can explicitly specify this.

Closes-Bug: #1898504
(cherry picked from commit c3bd490b67a53d86a40382ef644e41acf67d0b6a)

### DIFF
--- a/charmhelpers/contrib/storage/linux/ceph.py
+++ b/charmhelpers/contrib/storage/linux/ceph.py
@@ -268,7 +268,7 @@ class BasePool(object):
         'compression-max-blob-size': (int, None),
         'compression-max-blob-size-hdd': (int, None),
         'compression-max-blob-size-ssd': (int, None),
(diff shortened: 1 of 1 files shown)


ajkavanagh:
H @wolsen what's the status of this change, please? It's targeted against stable/21.04 which is now EOL (due to charmhub migration), and thus I was wondering if we can close it?  Also the associated gerrit review at https://review.opendev.org/c/openstack/charm-cinder-ceph/+/784871 and the bug at https://bugs.launchpad.net/charm-ceph-rbd-mirror/+bug/1898504?  Thanks!

PR #599, opened 2021-04-14
Fix duplicating other charms deferred event messages

gnuoy:
Stop charms from duplicating each others deferred event messages when
they are deployed on the same unit.

This includes a fix to ensure the policy_requestor_name and
policy_requestor_type are read properly from the event file.

Closes bug https://bugs.launchpad.net/charm-nova-compute/+bug/1923739


AurelienLourot (dismissed):

ajkavanagh (approved):
LGTM

PR #600, opened 2021-04-14
Fix duplicating other charms deferred event messages (#599)

gnuoy:
* Fix duplicating other charms deferred event messages

Stop charms from duplicating each others deferred event messages when
they are deployed on the same unit.

Closes bug https://bugs.launchpad.net/charm-nova-compute/+bug/1923739


thedac (approved):
Landing.

PR #601, opened 2021-04-22
Add checks for openvswitch and helper function to enable the checks

afreiberger:
This commit provides an nrpe script to check for errors in ```ovs-vsctl show``` output to be shared across several openstack networking charms or any other charm that may wish to add openvswitch monitoring.

There is also a helper function in contrib.charmsupport.nrpe to add_openvswitch_checks which will setup the necessary sudoers rights for the nagios user to introspect the running openvswitch process.

The check_openvswitch.py script relies upon nagios_plugins3 module which is delivered with charm-nrpe.  The add_openvswitch_checks method should not be used outside of the context of an nrpe-relation hook.

### DIFF
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -520,7 +520,7 @@ def add_openvswitch_checks(nrpe, unit_name):
     enable_sudo_for_openvswitch_checks()
     nrpe.add_check(
         shortname='openvswitch',
-        description='Check Open vSwitch {%s}' % unit_name,
+        description='Check Open vSwitch {{{0}}}'.format(unit_name),
         check_cmd='check_openvswitch.py')
 
 
@@ -528,7 +528,9 @@ def enable_sudo_for_openvswitch_checks():
     sudoers_dir = "/etc/sudoers.d"
     sudoers_mode = 0o100440
     ovs_sudoers_file = "99-check_openvswitch"
-    ovs_sudoers_entry = "nagios ALL=(root) NOPASSWD: /usr/bin/ovs-vsctl show"
+    ovs_sudoers_entry = ("# Juju owned - do not edit #\n"
+                         "nagios ALL=(root) NOPASSWD: /usr/bin/ovs-vsctl "
+                         "--format=json list Interface\n")
     dest = os.path.join(sudoers_dir, ovs_sudoers_file)
     try:
         with open(dest, "w") as sudoer_file:
--- a/charmhelpers/contrib/openstack/files/check_openvswitch.py
+++ b/charmhelpers/contrib/openstack/files/check_openvswitch.py
@@ -1,42 +1,62 @@
 #!/usr/bin/env python3
-# -*- coding: us-ascii -*-
-"""Check for issues with NVME hardware devices."""
+"""Check for known error statuses in OVS.
+
+This script currently checks through the Interface table in OVSDB
+for errors. The script is named generically to allow for expanded
+status checks in the future.
+"""
 
 import argparse
-import re
+import json
 import subprocess
 import sys
 
+# This depends on nagios_plugin3 being installed by charm-nrpe
 from nagios_plugin3 import CriticalError, UnknownError, try_check
 
 
-def parse_ovs_status():
-    """Check for errors in 'ovs-vsctl show' output."""
+def parse_ovs_interface_errors():
+    """Check for errors in OVSDB Interface table."""
     try:
-        cmd = ["/usr/bin/sudo", "/usr/bin/ovs-vsctl", "show"]
+        cmd = [
+            "/usr/bin/sudo",
+            "/usr/bin/ovs-vsctl",
+            "--format=json",
+            "list",
+            "Interface",
+        ]
         ovs_output = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
     except subprocess.CalledProcessError as e:
         raise UnknownError(
-            "UNKNOWN: Failed to query ovs: {}".format(
-                e.output.decode(errors="ignore").rstrip()
+            "UNKNOWN: OVS command '{}' failed.  Output: {}".format(
+                " ".join(cmd), e.output.decode(errors="ignore").rstrip(),
             )
         )
+    ovs_interfaces = json.loads(ovs_output.decode(errors="ignore"))
+    ovs_interface_errors = []
+    for i in ovs_interfaces["data"]:
+        # OVSDB internal data is formatted per RFC 7047 5.1
+        iface = dict(zip(ovs_interfaces["headings"], i))
+        error = iface["error"]
+        if isinstance(error, list) and len(error) == 2 and error[0] == "set":
+            # deserialize the set data into csv string elements
+            error = ",".join(error[1])
+        if error:
+            ovs_interface_errors.append(
+                "Error on iface {}: {}".format(iface["name"], error)
+            )
 
-    ovs_vsctl_show_errors = []
-    ovs_error_re = re.compile(r"^.*error: (?P<message>.+)$", re.I)
-    for line in ovs_output.decode(errors="ignore").splitlines():
-        m = ovs_error_re.match(line)
-        if m:
-            ovs_vsctl_show_errors.append(m.group("message"))
-
-    if ovs_vsctl_show_errors:
-        numerrs = len(ovs_vsctl_show_errors)
+    if ovs_interface_errors:
         raise CriticalError(
-            "CRITICAL: Found {} error(s) in ovs-vsctl show: "
-            "{}".format(numerrs, ", ".join(ovs_vsctl_show_errors))
+            "CRITICAL: Found {} interface error(s) in OVSDB: "
+            "{}".format(len(ovs_interface_errors), ", ".join(ovs_interface_errors))
         )
 
-    print("OK: no errors found in openvswitch")
+    print(
+        "OK: No errors found across {} interfaces in openvswitch".format(
+            len(ovs_interfaces["data"])
+        )
+    )
 
 
 def parse_args(argv=None):
@@ -44,8 +64,8 @@ def parse_args(argv=None):
     parser = argparse.ArgumentParser(
         prog="check_openvswitch",
         description=(
-            "this program checks openvswitch status and outputs an "
-            "appropriate Nagios status line"
+            "this program checks openvswitch interface status and outputs "
+            "an appropriate Nagios status line"
         ),
         formatter_class=argparse.ArgumentDefaultsHelpFormatter,
     )
@@ -55,7 +75,7 @@ def parse_args(argv=None):
 def main(argv):
     """Define main subroutine."""
     parse_args(argv)
-    try_check(parse_ovs_status)
+    try_check(parse_ovs_interface_errors)
 
 
 if __name__ == "__main__":


charmhelpers/contrib/openstack/files/check_openvswitch.py
```diff
@@ -0,0 +1,62 @@
...
+                e.output.decode(errors="ignore").rstrip()
+            )
+        )
+
+    ovs_vsctl_show_errors = []
+    ovs_error_re = re.compile(r"^.*error: (?P<message>.+)$", re.I)
+    for line in ovs_output.decode(errors="ignore").splitlines():
+        m = ovs_error_re.match(line)
+        if m:
+            ovs_vsctl_show_errors.append(m.group("message"))
```
ajkavanagh, line 30: Rather than using a regex, the `ovs-vsctl` command does support outputing in json (`--format=json`).  Is there a reason for not doing that (perhaps that the errors would appear in different nodes in different versions??)  Just wondering how to make it less magic.

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -509,6 +510,37 @@ def add_haproxy_checks(nrpe, unit_name):
...
+    try:
+        with open(dest, "w") as sudoer_file:
+            sudoer_file.write(ovs_sudoers_entry)
+        os.chmod(dest, sudoers_mode)
+        os.chown(dest, uid=0, gid=0)
+    except (OSError, IOError) as e:
+        log("Failed to setup sudoers file for check_openvswitch: {}".format(e))
+    else:
+        log("Sudoers file for check_openvswitch installed: {}".format(dest), DEBUG)
+
```
ajkavanagh, line 542: Would it be possible to add unit tests for these two functions?

ajkavanagh (commented):
Thanks for the patch; just a couple of comments around testing and extracting data from the ovs-vsctl command.

charmhelpers/contrib/openstack/files/check_openvswitch.py
afreiberger: ovs-vsctl show does not support the --format option (the args are parsed, but the output is not differentiated).

```
root@ruling-manta:/home/ubuntu# ovs-vsctl -f json show
fcaf57e2-8667-4972-b687-169789d1d15d
    Bridge "br0"
        Port "dpdk-p1"
            Interface "dpdk-p1"
                type: dpdk
                options: {dpdk-devargs="0000:01:00.1"}
                error: "could not open network device dpdk-p1 (Address family not supported by protocol)"
        Port "br0"
            Interface "br0"
                type: internal
        Port "dpdk-p0"
            Interface "dpdk-p0"
                type: dpdk
                options: {dpdk-devargs="0000:01:00.0"}
                error: "could not open network device dpdk-p0 (Address family not supported by protocol)"
    ovs_version: "2.9.8"
```

I've also investigated using the json output for ```ovs-vsctl list Interfaces```, but that takes as much processing as using csv output:

```
{"data":[[["uuid","9bc2d04d-0069-4a76-918a-b9ac61d4c5ce"],["set",[]],["map",[]],["map",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],"could not open network device dpdk-p1 (Address family not supported by protocol)",["map",[]],["set",[]],0,0,["set",[]],["set",[]],["set",[]],["set",[]],["map",[]],["set",[]],["set",[]],["set",[]],["set",[]],"dpdk-p1",-1,["set",[]],["map",[["dpdk-devargs","0000:01:00.1"]]],["map",[]],["map",[]],["map",[]],"dpdk"],[["uuid","2d8758ca-510f-4399-ab84-8c1f3a33061e"],"down",["map",[]],["map",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],["map",[]],4,0,0,["set",[]],0,["set",[]],"down",["map",[]],["set",[]],"c2:bd:86:7f:fa:4d",1500,["set",[]],"br0",65534,["set",[]],["map",[]],["map",[]],["map",[["collisions",0],["rx_bytes",0],["rx_crc_err",0],["rx_dropped",2],["rx_errors",0],["rx_frame_err",0],["rx_over_err",0],["rx_packets",0],["tx_bytes",0],["tx_dropped",0],["tx_errors",0],["tx_packets",0]]],["map",[["driver_name","openvswitch"]]],"internal"],[["uuid","d67191aa-bdae-4ed5-95c8-86778276052e"],["set",[]],["map",[]],["map",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],["set",[]],"could not open network device dpdk-p0 (Address family not supported by protocol)",["map",[]],["set",[]],0,0,["set",[]],["set",[]],["set",[]],["set",[]],["map",[]],["set",[]],["set",[]],["set",[]],["set",[]],"dpdk-p0",-1,["set",[]],["map",[["dpdk-devargs","0000:01:00.0"]]],["map",[]],["map",[]],["map",[]],"dpdk"]],"headings":["_uuid","admin_state","bfd","bfd_status","cfm_fault","cfm_fault_status","cfm_flap_count","cfm_health","cfm_mpid","cfm_remote_mpids","cfm_remote_opstate","duplex","error","external_ids","ifindex","ingress_policing_burst","ingress_policing_rate","lacp_current","link_resets","link_speed","link_state","lldp","mac","mac_in_use","mtu","mtu_request","name","ofport","ofport_request","options","other_config","statistics","status","type"]}
```

Very odd that the project uses "headings" for indexing the data model instead of making it key-value oriented output as would be expected of json.  Trying to use this instead of parsing with regex, I get the following that requires some additional processing for values that are empty, as every Interface has an errors key, and if it's blank, the value of that list index is a list that contains the data type, "set", and the empty set, [].

```
>>> error_index = data["headings"].index("error")
>>> for interface in data["data"]:
...     print(interface[error_index])
... 
could not open network device dpdk-p1 (Address family not supported by protocol)
['set', []]
could not open network device dpdk-p0 (Address family not supported by protocol)
```

This could certainly be used instead of regex with an ```if interface[error_index] != list(['set', []]):``` but I chose the simpler to read regex, and am also hoping to catch errors from 'ovs-vsctl show' that may not be Interface related (though for the current requirement, limiting to checking for Interface errors would suffice).

Do you have advice regarding readability of code vs using something other than regex in a situation like this?  The regex, to me, seemed more elegant and readable vs the additional handling of missing "error" index as well as the handling of the ['set', []].

charmhelpers/contrib/charmsupport/nrpe.py
afreiberger: I'll see if I can adapt the current unit testing around this module to also test these functions, however, my plan was to add functional testing to the consuming charms since these are both external callout functions and the tests would simply be to ensure the functions are called with expected inputs, since they'd all be mocked.

afreiberger: I've added unittests and they are passing (though I'm seeing failures on test_check_restart_timestamps (tests.contrib.openstack.test_deferred_events.DeferredCharmServiceEventsTestCase))

```
  File "/home/ubuntu/charm-helpers/tests/contrib/openstack/test_deferred_events.py", line 326, in test_check_restart_timestamps
    self.assertFalse(clear_deferred_restarts.called)
AssertionError: True is not false
```

This is also erroring in stable/21.04 branch, so I'm ignoring for this PR.

afreiberger: Looks like that error is my test env, but OSCI appears okay.

charmhelpers/contrib/openstack/files/check_openvswitch.py
afreiberger: After investigating all of the other tables, Interfaces is the only table that has the error column, so I'll write this more deterministically and be able to include potentially vital interface information in the notification.

ajkavanagh: I didn't realize it wasn't a complete implemention (the --format option being missing).  I always worry about using regex's on human-readable/consumable output as it is prone to be changed (on a whim sometimes!) and so it can make the code brittle.

I think from your explanations, it's fine to go with regex as a pragmatic solution as long as all error conditions are handled.  I'll go back and look again.

afreiberger: I've updated with latest commit to use the Interface table json.

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -509,6 +510,37 @@ def add_haproxy_checks(nrpe, unit_name):
...
+    """
+    Add checks for openvswitch
+
+    :param NRPE nrpe: NRPE object to add check to
+    :param str unit_name: Unit name to use in check description
+    """
+    enable_sudo_for_openvswitch_checks()
+    nrpe.add_check(
+        shortname='openvswitch',
+        description='Check Open vSwitch {%s}' % unit_name,
```
jtroup, line 523: How about 'Check for interface errors in Open vSwitch'?

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -509,6 +510,37 @@ def add_haproxy_checks(nrpe, unit_name):
...
+        check_cmd='check_openvswitch.py')
```
jtroup, line 524: How about make the name more explicit, e.g. check_ovs_interfaces.py or check_ovs_ifaces.py?

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -509,6 +510,37 @@ def add_haproxy_checks(nrpe, unit_name):
...
+
+
+def enable_sudo_for_openvswitch_checks():
+    sudoers_dir = "/etc/sudoers.d"
+    sudoers_mode = 0o100440
+    ovs_sudoers_file = "99-check_openvswitch"
+    ovs_sudoers_entry = "nagios ALL=(root) NOPASSWD: /usr/bin/ovs-vsctl --format=json list Interface"
```
jtroup, line 531: It'd be nice to include a 'Juju owned - do not edit' header, if possible?

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -509,6 +510,37 @@ def add_haproxy_checks(nrpe, unit_name):
...
+    nrpe.add_check(
+        shortname='openvswitch',
+        description='Check Open vSwitch {%s}' % unit_name,
+        check_cmd='check_openvswitch.py')
+
+
+def enable_sudo_for_openvswitch_checks():
+    sudoers_dir = "/etc/sudoers.d"
+    sudoers_mode = 0o100440
+    ovs_sudoers_file = "99-check_openvswitch"
```
jtroup, line 530: As above, suggest check_ovs_interfaces; unless the plan is to expand in the future?

charmhelpers/contrib/openstack/files/check_openvswitch.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
+# -*- coding: us-ascii -*-
+"""Check for OVS interface errors."""
+
+import argparse
+import json
+import subprocess
+import sys
+
+# dependence on nagios_plugin3 being installed by charm-nrpe
```
jtroup, line 10: Perhaps: s/dependence/This depends/

charmhelpers/contrib/openstack/files/check_openvswitch.py
```diff
@@ -0,0 +1,67 @@
...
+from nagios_plugin3 import CriticalError, UnknownError, try_check
+
+
+def parse_ovs_status():
+    """Check for errors in 'ovs-vsctl show' output."""
```
jtroup, line 15: It's not actually running 'show' anymore :-)

charmhelpers/contrib/openstack/files/check_openvswitch.py
```diff
@@ -0,0 +1,67 @@
...
+    try:
+        cmd = ["/usr/bin/sudo", "/usr/bin/ovs-vsctl", "--format=json", "list", "Interface"]
+        ovs_output = subprocess.check_output(cmd, stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError as e:
+        raise UnknownError(
+            "UNKNOWN: Failed to query ovs: {}".format(
```
jtroup, line 21: Being more explicit about what failed, would be nice, e.g. s/Failed to query ovs/ovs-vsctl list Interface failed/ as this makes it less likely the Operator has to dig up the charm code to figure out what failed?

charmhelpers/contrib/openstack/files/check_openvswitch.py
```diff
@@ -0,0 +1,67 @@
...
+            ovs_interface_errors.append("Error on iface {}: {}".format(iface["name"], error))
+
+    if ovs_interface_errors:
+        numerrs = len(ovs_interface_errors)
+        raise CriticalError(
+            "CRITICAL: Found {} error(s) in OVSDB: "
+            "{}".format(numerrs, ", ".join(ovs_interface_errors))
+        )
+
+    print("OK: no errors found in openvswitch")
```
jtroup, line 44: Perhaps 'no interface errors'?

charmhelpers/contrib/openstack/files/check_openvswitch.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def parse_args(argv=None):
+    """Process CLI arguments."""
+    parser = argparse.ArgumentParser(
+        prog="check_openvswitch",
+        description=(
+            "this program checks openvswitch status and outputs an "
```
jtroup, line 52: s/status/interface status/

charmhelpers/contrib/openstack/files/check_openvswitch.py
```diff
@@ -0,0 +1,67 @@
...
+            "{}".format(numerrs, ", ".join(ovs_interface_errors))
+        )
+
+    print("OK: no errors found in openvswitch")
+
+
+def parse_args(argv=None):
+    """Process CLI arguments."""
+    parser = argparse.ArgumentParser(
+        prog="check_openvswitch",
```
jtroup, line 50: As previous

charmhelpers/contrib/openstack/files/check_openvswitch.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python3
+# -*- coding: us-ascii -*-
```
jtroup, line 2: Not a big deal, but any particular reason why?

charmhelpers/contrib/openstack/files/check_openvswitch.py
```diff
@@ -0,0 +1,67 @@
...
+        iface = dict(zip(ovs_interfaces["headings"], i))
+        error = iface["error"]
+        if isinstance(error, list) and len(error) == 2 and error[0] == "set":
+            # deserialize the set data into csv string elements
+            error = ",".join(error[1])
+        if error:
+            ovs_interface_errors.append("Error on iface {}: {}".format(iface["name"], error))
+
+    if ovs_interface_errors:
+        numerrs = len(ovs_interface_errors)
```
jtroup, line 38: I thought this was a typo for 'numbers' at first - num_errs would be more consistent with other variable names?

charmhelpers/contrib/openstack/files/check_openvswitch.py
jtroup: As a followup thought; I always love nagios checks which have a side effect of giving you other information.  For example wouldn't it be cool it said:

OK: no errors found in NN interfaces in Open vSwitch

And the cost would be a simple counter?

jtroup (commented):
A bunch of trivial feedback; feel free to ignore it all if you like.  Just happy to see this check be added.

charmhelpers/contrib/charmsupport/nrpe.py
jtroup: Also, .format() for consistency?

charmhelpers/contrib/charmsupport/nrpe.py
afreiberger: I was definitely feeling this would be the start of something that could be expanded to more checks as we identify them.  The interface errors are just MVP for the current need.

charmhelpers/contrib/charmsupport/nrpe.py
afreiberger: +1

charmhelpers/contrib/openstack/files/check_openvswitch.py
afreiberger: This came from an hw-health check script that I used as a shell.  it can be removed.

charmhelpers/contrib/openstack/files/check_openvswitch.py
afreiberger: Found it's just as readable just using the len() call in the format.

PR #604, opened 2021-05-11
contrib/ovs/ovsdb: Allow passing extra args to tool

fnordahl:
When using SimpleOVSDB with a non-local database it may be useful
to pass extra arguments to the tool.


ajkavanagh (approved):
LGTM

PR #609, opened 2021-05-25
Added watermark_scale_factor tuning

brettmilford:
Adds a method for determining an appropriate vm.watermark_scale_factor
value for baremetal systems.

Implements: spec memory-fragmentation-tuning

https://review.opendev.org/c/openstack/charm-specs/+/792383


charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,82 @@
...
+    :returns: watermark_scale_factor
+    :type: int
+    """
+
+    try:
+        with open('/proc/meminfo', 'r') as f:
+            line = f.readline()
+            while line != '':
+                if "MemTotal" in line:
+                    mem_total = int(P.search(line).group())  # int([v for v in line.split(' ') if P.match(v)][0])
```
ajkavanagh, line 45: Is the comment left over?

charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,82 @@
...
+    """Calculates optimal vm.watermark_scale_factor value
+
+    :returns: watermark_scale_factor
+    :type: int
+    """
+
+    try:
+        with open('/proc/meminfo', 'r') as f:
+            line = f.readline()
+            while line != '':
```
ajkavanagh, line 43: Why not just:

```python
while line:
    ...
```

I.e. rely on the truthiness of Python's tests.

charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,82 @@
...
+                if "MemTotal" in line:
+                    mem_total = int(P.search(line).group())  # int([v for v in line.split(' ') if P.match(v)][0])
+                    break
+    except OSError as e:
+        log(f"Failed to read /proc/meminfo in calculating watermark_scale_factor: {e}", ERROR)
+        raise e
```
ajkavanagh, line 49: What happens if (for a bizarre reason) "MemTotal" is never found in the line?  the variable will never get set, and the function will fail further down?  What if a blank line is incorporated into the output prior to  "MemTotal" at some point?

In fact as a I read back the whole section looks a bit strange to me?  Why not something like:

```python
with open(...) as f:
    for line in f:
        if "MemTotal" in line:
            ...
            break
    else:
        raise RuntimeException("Couldn't find MemTotal??")
```

charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,82 @@
...
+
+    try:
+        normal_managed_pages = []
+        with open('/proc/zoneinfo', 'r') as f:
+            for line in f.read().splitlines():
+                if "Node" in line and "zone" in line:
+                    zone = [v for v in line.split(' ')
+                            if v in ["DMA", "DMA32", "Normal", "Movable", "Device"]][0]
+                    # node = int(P.search(line).group())  # int([v for v in line.split(' ') if P.match(v)][0].rstrip(','))
```
ajkavanagh, line 58: Left over comment?

charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,82 @@
...
+                    mem_total = int(P.search(line).group())  # int([v for v in line.split(' ') if P.match(v)][0])
+                    break
+    except OSError as e:
+        log(f"Failed to read /proc/meminfo in calculating watermark_scale_factor: {e}", ERROR)
+        raise e
+
+    try:
+        normal_managed_pages = []
+        with open('/proc/zoneinfo', 'r') as f:
+            for line in f.read().splitlines():
```
ajkavanagh, line 54: if /proc/zoneinfo is a ascii/utf-u file, the  `for line in f:` will do the same.

charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,82 @@
...
+                if "Node" in line and "zone" in line:
+                    zone = [v for v in line.split(' ')
+                            if v in ["DMA", "DMA32", "Normal", "Movable", "Device"]][0]
+                    # node = int(P.search(line).group())  # int([v for v in line.split(' ') if P.match(v)][0].rstrip(','))
+
+                if zone == "Normal" and "managed" in line:
+                    managed = int([v for v in line.split(' ') if P.match(v)][0])
+                    normal_managed_pages.append(managed)
+    except OSError as e:
+        log(f"Failed to read /proc/zoneinfo in calculating watermark_scale_factor: {e}", ERROR)
```
ajkavanagh, line 64: Unfortunately, f-strings aren't supported on Py35 and we have to support that still.  Please could you change to `.format()`.

charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,82 @@
...
+        raise e
+
+    wmark = min([watermark_scale_factor(mem_total, managed_pages)
+                 for managed_pages in normal_managed_pages])
+    log(f"vm.watermark_scale_factor: {wmark}", DEBUG)
```
ajkavanagh, line 69: As above.

charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,82 @@
...
+
+                if zone == "Normal" and "managed" in line:
+                    managed = int([v for v in line.split(' ') if P.match(v)][0])
+                    normal_managed_pages.append(managed)
+    except OSError as e:
+        log(f"Failed to read /proc/zoneinfo in calculating watermark_scale_factor: {e}", ERROR)
+        raise e
+
+    wmark = min([watermark_scale_factor(mem_total, managed_pages)
+                 for managed_pages in normal_managed_pages])
```
ajkavanagh, line 68: If `normal_managed_pages` is an empty list, this breaks as `min` will throw a `ValueError` (I think) for an empty list.  Is this a scenario that can happen?

ajkavanagh (changes requested):
Thank you for the patch for the additional functionality.  I have a few inline comments (please see against the file), and also I think the tests need to be expanded to test the `calculate_watermark_scale_factor()` function (i.e. only the `watermark_scale_factor()` function is tested).

It would also be good to actually feed it test fixtures for the `/proc/zoneinfo` and `/proc/meminfo/`.

Also, technically, the library still supports Py27 charms, and definitely supports Py35 (xenial) charms.

Thanks.

brettmilford:
Thanks for the feedback @ajkavanagh.
I've tried implementing your suggestions and improving testing in the latest commit.

charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,100 @@
...
+                    normal_managed_pages.append(managed)
+    except OSError as e:
+        log("Failed to read /proc/zoneinfo in calculating watermark_scale_factor: {}".format(e), ERROR)
+        raise e
+
+    return normal_managed_pages
+
+
+def watermark_scale_factor(mem_total, managed_pages):
+    # if < 16G ram return default
```
ajkavanagh, line 92: docstring please for the function detailing the parameters and return value (inc. types)

charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,100 @@
...
+        normal_managed_pages = []
+        with open('/proc/zoneinfo', 'r') as f:
+            for line in f:
+                if "Node" in line and "zone" in line:
+                    zone = [v for v in line.split(' ')
+                            if v in ["DMA", "DMA32", "Normal", "Movable", "Device"]][0]
+
+                if zone == "Normal" and "managed" in line:
+                    managed = int([v for v in line.split(' ') if P.match(v)][0])
+                    normal_managed_pages.append(managed)
```
ajkavanagh, line 83: So a few issues with the code here:

1. `zone` is potentially accessed at line 81 before it is assigned at 78 (due to the if statement).  Before the loop, a
  `zone = None` ought to be done to eliminate a variable not found type error.
2. The code is using the string in `zone` as a boolean to indicate whether it is parsing a "Node <d>, zone Normal" section in the `/proc/zoneinfo` output.  Can we make this more obvious by:
  * A comment on what the block of code is doing.
  * Using a boolean to indicate that it is in the 'Normal Zone'

I might do something like the following:

```python
with open(...) as f:
    in_zone_normal = False
    # regex to search for strings that look like "Node 0, zone    Normal" and last string to group 1
    normal_zone_matcher = re.compiile(r"^Node\s\d+\s+zone\s+(\S+)$")
    # regex to match to a number at the end of the line.
    managed_matcher = re.compile(r"\s+managed\s+(\d+)$")
    for line in f:
        # See if the line is at the beginning of a Node section, capturing it to group 1
        match = normal_zone_matcher.search(line)
        if match:
            in_zone_normal = match.group(1) == 'Normal'
        if in_zone_normal:
            # match the number at the end of "     managed    3840" into group 1.
             managed_match = managed_matcher.search(line)
             if managed_match:
                normal_managed_pages.append(managed_match.group(1))
                # skip the rest of the section
                in_zone_normal = False
```

Note, I'm only using a regex to make it clear that we are searching for a particular type of line to then start sub-searching for the 'managed' part in the line.

tests/contrib/sysctl/test_watermark_scale_factor.py
```diff
@@ -0,0 +1,619 @@
+#!/usr/bin/env python
+
+from charmhelpers.contrib.sysctl.watermark_scale_factor import watermark_scale_factor, calculate_watermark_scale_factor, get_memtotal, get_normal_managed_pages
```
ajkavanagh, line 3: Please could we reformat this line to:

```python
from charmhelpers.contrib.sysctl.watermark_scale_factor import (
    watermark_scale_factor,
    calculate_watermark_scale_factor,
    get_memtotal,
    get_normal_managed_pages,
)
```

tests/contrib/sysctl/test_watermark_scale_factor.py
```diff
@@ -0,0 +1,619 @@
...
+HugePages_Rsvd:        0
+HugePages_Surp:        0
+Hugepagesize:       2048 kB
+Hugetlb:               0 kB
+DirectMap4k:    85061760 kB
+DirectMap2M:    14542848 kB
+DirectMap1G:     2097152 kB
+"""
+
+PROC_ZONEINFO = """
```
ajkavanagh, line 68: Although it is excellent to use real data, it might be appropriate to cut this down a bit please?

ajkavanagh (changes requested):
Very much more readable this time; much improved.  Unfortunately(!) this let me see some more areas (mostly docstrings) and also a suggestion on improving the code around the searching or proc/zoneinfo.

Many thanks.

brettmilford:
Hi Alex,
Thanks again for your extensive feedback.
I much prefer your approach to parsing /proc/zoneinfo so I've largely used that.

The heuristics for calculating vm.watermark_scale_factor might change based on [this charm spec](https://review.opendev.org/c/openstack/charm-specs/+/792383), and subsequent reviews. 
Also currently the unit tests for `get_memtotal` and `get_normal_managed_pages` are failing as mock_open doesn't appear to be using read_data as I expected.

ajkavanagh (dismissed):
LGTM - let me know if it's ready for merging; you mentioned that there might be a change?  Also, thanks very much for your work on this and the changes.

brettmilford:
@ajkavanagh I think this should now be ready for merging. The associated spec is approved and the tests are passing for me in my env with the latest commit.

ajkavanagh:
> @ajkavanagh I think this should now be ready for merging. The associated spec is approved and the tests are passing for me in my env with the latest commit.

Thanks @brettmilford ; I'm back from leave now and have just kicked off the tests.

brettmilford:
Hey @ajkavanagh, can you please run the tests again. TY

charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,126 @@
...
+    :param managed_pages: Number of managed pages
+    :type managed_pages: int
+    :returns: normal_managed_pages
+    :rtype: int
+    """
+    if memtotal <= MEMTOTAL_MIN_KB:
+        return WMARK_DEFAULT
+    else:
+        wmark = int(MAX_PAGES / managed_pages)
+        if wmark > 1000:
```
niedbalski, line 123: make this a constant.

charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,126 @@
...
+                if "MemTotal" in line:
+                    memtotal = int(re.search(r"\d+", line).group())
+                    break
+            else:
+                raise Exception("Could not find MemTotal")
+    except (Exception, OSError) as e:
+        log("Failed to parse /proc/meminfo in calculating watermark_scale_factor: {}".format(e), ERROR)
+        raise e
+
+    return memtotal
```
niedbalski, line 73: Can you use https://github.com/juju/charm-helpers/blob/b5725ac546372e7d4004d15095f79cdd5e7da687/charmhelpers/core/host.py#L1116 instead? Having 2 mechanisms for calculating total memory might be harder to follow.

niedbalski (changes requested):
LGTM, minor changes required.

charmhelpers/contrib/sysctl/watermark_scale_factor.py
brettmilford: I'm not adverse to it, although it does appear to return value in a different unit.

charmhelpers/contrib/sysctl/watermark_scale_factor.py
brettmilford: just declare wmark in all caps?

niedbalski: yes please

charmhelpers/contrib/sysctl/watermark_scale_factor.py
brettmilford: done now

brettmilford:
@niedbalski @ajkavanagh please take another look, I've added the changes requested. TY

charmhelpers/contrib/sysctl/watermark_scale_factor.py
```diff
@@ -0,0 +1,104 @@
...
+    :param managed_pages: Number of managed pages
+    :type managed_pages: int
+    :returns: normal_managed_pages
+    :rtype: int
+    """
+    if memtotal <= MEMTOTAL_MIN_BYTES:
+        return WMARK_DEFAULT
+    else:
+        WMARK = int(MAX_PAGES / managed_pages)
+        if WMARK > 1000:
```
niedbalski, line 101: replace 1000 with WMARK_MAX

niedbalski (changes requested):
Minor change requested.

niedbalski (approved):

PR #613, opened 2021-06-22
Set WSGI env with lang=C.UTF-8

thedac:
The default apache2 environment, and therefore the default WSGI
environment, sets LANG=C. For languages beyond the scope of ASCII
this can create problems.

Explicitly set the lang and locale to C.UTF-8 which will support all
UTF-8 language sets.

Related-Bug: #1933109


wolsen (approved):
LGTM

PR #617, opened 2021-07-08
Add a base openstack code name as a source for add_source()

ajkavanagh:
This is a 'smart' option, where the openstack-origin=ussuri, for
example, passing 'ussuri' to the add_source() function.  This, on focal,
will be the equivalent of 'distro' and on bionic, will be
'cloud:bionic-ussuri'.  On groovy, victoria is distro, etc.

Also support a 'ussuri/proposed' to allow proposed to be carried between
distro versions.  This is to simplify the logic for setting openstack
versions so that the version wanted is the name rather than 'distro'.

### DIFF
--- a/charmhelpers/fetch/ubuntu.py
+++ b/charmhelpers/fetch/ubuntu.py
@@ -625,7 +625,7 @@ def add_source(source, key=None, fail_invalid=False):
     '<openstack-version>': translate to cloud:<release> based on the current
       distro version (i.e. for 'ussuri' this will either be 'bionic-ussuri' or
       'distro'.
-    '<openstack-version/proposed': as above, but for proposed.
+    '<openstack-version>/proposed': as above, but for proposed.
 
     Otherwise the source is not recognised and this is logged to the juju log.
     However, no error is raised, unless sys_error_on_exit is True.
--- a/charmhelpers/fetch/ubuntu.py
+++ b/charmhelpers/fetch/ubuntu.py
@@ -810,7 +810,6 @@ def _add_bare_openstack(openstack_release):
     :type openstack_release: str
     :raises: SourceConfigError
     """
-    # So it's distro, which means it's a no-op so end here.
     # TODO(ajkavanagh) - surely this means we should be removing cloud archives
     # if they exist?
     __add_bare_helper(openstack_release, "{}-{}", lambda: None)
--- a/charmhelpers/fetch/ubuntu.py
+++ b/charmhelpers/fetch/ubuntu.py
@@ -211,6 +211,54 @@ CLOUD_ARCHIVE_POCKETS = {
 }
 
 
+OPENSTACK_RELEASES = (
+    'diablo',
+    'essex',
+    'folsom',
+    'grizzly',
+    'havana',
+    'icehouse',
+    'juno',
+    'kilo',
+    'liberty',
+    'mitaka',
+    'newton',
+    'ocata',
+    'pike',
+    'queens',
+    'rocky',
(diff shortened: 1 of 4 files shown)


charmhelpers/fetch/ubuntu.py
```diff
@@ -574,6 +622,10 @@ def add_source(source, key=None, fail_invalid=False):
       with be used.  If staging is NOT used then the cloud archive [3] will be
       added, and the 'ubuntu-cloud-keyring' package will be added for the
       current distro.
+    '<openstack-version>': translate to cloud:<release> based on the current
+      distro version (i.e. for 'ussuri' this will either be 'bionic-ussuri' or
+      'distro'.
+    '<openstack-version/proposed': as above, but for proposed.
```
ChrisMacNaughton, line 628: Should this say: `'<openstack-version>/proposed'`?

ajkavanagh: > Should this say: `'<openstack-version>/proposed'`?

Ack, yes, it should.  Good catch, thank you.

charmhelpers/fetch/ubuntu.py
```diff
@@ -738,6 +799,74 @@ def _verify_is_ubuntu_rel(release, os_release):
...
+
+    The spec given is, say, 'ussuri', but this could apply cloud:bionic-ussuri
+    or 'distro' depending on whether the ubuntu release is bionic or focal.
+
+    :param openstack_release: the OpenStack codename to determine the release
+        for.
+    :type openstack_release: str
+    :raises: SourceConfigError
+    """
+    # So it's distro, which means it's a no-op so end here.
```
AurelienLourot, line 813: Should this comment be moved somewhere else? It's not necessarily "distro" at this point. I suspect this line was moved by mistake from somewhere else?

tests/fetch/test_fetch_ubuntu.py
```diff
@@ -730,6 +730,98 @@ def test_add_distro_source(self, log):
...
+        mock_get_distrib_codename.return_value = 'bionic'
+        fetch.add_source('ussuri')
+        mock_add_cloud_pocket.assert_called_once_with("bionic-ussuri")
+
+    @patch('charmhelpers.fetch.ubuntu._add_cloud_pocket')
+    @patch('charmhelpers.fetch.ubuntu.get_distrib_codename')
+    def test_add_bare_openstack_impossible_version(
+            self, mock_get_distrib_codename, mock_add_cloud_pocket):
+        mock_get_distrib_codename.return_value = 'xenial'
+        try:
```
AurelienLourot, line 754: nitpick, you just reinvented this wheel:

```
with self.assertRaises(fetch.SourceConfigError):
    fetch.add_source('ussuri')
mock_add_cloud_pocket.assert_not_called()
```

AurelienLourot (commented):
LGTM in general, thanks! Just two nitpicks

charmhelpers/fetch/ubuntu.py
ajkavanagh: Ack; left in by mistake.

tests/fetch/test_fetch_ubuntu.py
ajkavanagh: If you look at the rest of the file, it's the pattern that already exists.  Also, assertRaises() isn't actually available in this class for the bizarre reason that it uses testtools where it doesn't exist.  One of those oddities of the library.

AurelienLourot (approved):

PR #618, opened 2021-07-12
Drop Amulet support

AurelienLourot:
Amulet has been replaced by Zaza.

https://launchpad.net/bugs/1828424


AurelienLourot:
Actually there is urgency to remove this code as it's making use of `distro-info` which makes the new pip resolver clash, see #621

ChrisMacNaughton:
Could we just drop the OpenStack related paths?

AurelienLourot:
`distro_info` is used in `get_ubuntu_releases()`, which isn't OpenStack-related. Or what do you mean?

ChrisMacNaughton:
Is ‘get_ubuntu_releases’ really only used for amulet? That’s quite surprising to me

AurelienLourot:
Yes. Please grep the repo recursively for `get_ubuntu_releases` and you'll see what I'm talking about.

ajkavanagh:
> Actually there is urgency to remove this code as it's making use of `distro-info` which makes the new pip resolver clash, see #621

No I totally agree; let's get rid it of it.

fnordahl:
+1, I was about to merge this but I did not find that the PR removes the offending `distro-info` package from test-requirements.txt?

It is worth noting that while the `distro-info` package has value in the distribution, it is absolutely useless as an independent Python wheel. I guess the best way to consume it across a virtualenv boundary would be to execute the `distro-info` binary.

fnordahl (dismissed):
LGTM, thanks!

fnordahl (approved):
LGTM, thanks!

PR #621, opened 2021-07-19
Confirm wrong data-port format raises ValueError

gabrielcocenza:
[EDIT]

This PR attempts to add tests that confirm the behavior of raise ValueError when wrong `data-port` config is passed from users. The handler of this error should be implemented in the respective charm that uses the charm-helpers.

The bug was filled  on charm-ovn-chassis  located at this [link](https://bugs.launchpad.net/charm-ovn-chassis/+bug/1919481)
    
Closes-Bug: #1919481
    

~~The expected format for the bridges key is space-delimited list of key-value pairs.
Ex: "br-internet:00:00:5e:00:00:42 br-provider:enp3s0f0". If user sets a wrong format,
the config-changed hook fails.~~
    
~~In order to fix this, a regex was set to check if is in the correct format including
empty strings. If the format is wrong, it sets the juju status to "blocked" and
raises an exception.~~

~~The regex ` ^((\S+:\S+)( \S+:\S+)*|^$)$ ` means that the string should start and
end with content inside that can be empty or that have one or more non-space
character followed by ":" and one or more non-space character. In case of more
than one key-value, the expression searches for exactly one space and then
one or more non-space character followed by ":" and one or more non-space
character. The expression accepts any number of pairs of key-value.~~

~~The following examples of key-value pairs are accepted:~~

~~1. "<str>:<str>" Ex: "br-provider:enp3s0f0"
2.  "<str>:<str> <str>:<str>" Ex: "br-provider:enp3s0f0 br-internet:00:00:5e:00:00:42"
3.  "" (empty string)~~

~~The following examples of key-value pairs are not accepted:~~

~~1. "<str>"
2. "<str>:"
3. "<str>:<str>  <str>:<str>" (Double spaces between key-values)
4. "<str>:<str> <str>:"~~


charmhelpers/contrib/openstack/context.py
```diff
@@ -2798,6 +2798,18 @@ def __init__(self, bridges_key=None, bonds_key=None, enable_dpdk_key=None,
...
+        # check config_bridges pattern "<str>:<str>","<str>:<str> <str>:<str>" or ""
+        config_bridges_pattern = re.match(
+            r'^((\S+:\S+)( \S+:\S+)*|^$)$',
+            config_bridges
+        )
+        if not config_bridges_pattern:
+            msg = ('Wrong format. '
+                   'Expected format is space-delimited list of key-value pairs.'
+                   'Ex: "br-internet:00:00:5e:00:00:42 br-provider:enp3s0f0"')
+            status_set('blocked', msg)
```
fnordahl, line 2810: The call to `status_set` here would be rather pointless, as you have no idea what the charm implementation might do after the raise. If you want all of the charms to use the same message I guess you could add that through a custom exception or something like that?

charmhelpers/contrib/openstack/context.py
```diff
@@ -2798,6 +2798,18 @@ def __init__(self, bridges_key=None, bonds_key=None, enable_dpdk_key=None,
                 self._mac_pci_address_map[bond_mac.mac] = pci_address
 
         config_bridges = config(bridges_key) or ''
+        # check config_bridges pattern "<str>:<str>","<str>:<str> <str>:<str>" or ""
+        config_bridges_pattern = re.match(
+            r'^((\S+:\S+)( \S+:\S+)*|^$)$',
```
fnordahl, line 2803: I'm not too happy about the use of regular expressions in this context. Reading the original bug, Pythons text parsing functions used below would raise an ValueError already, and looking at an charm implementation consuming this change:

https://github.com/openstack-charmers/charm-layer-ovn/pull/50/commits/1d2a5a61c6289c3831e893aed3044e8d20be9c42#diff-727089e91d7a493a8b32c81042f625eed48f5b67ca485f26469a85d078eaa353R785-R793

You are adding exception handling to the charm implementation anyway. What do you think about just omitting the regex?

fnordahl (changes requested):
Thank you for your work on this. As pointed out in-line I don't think the regex is necessary here. If you want all consuming charms to use the same message create a custom exception with the message. And keep the tests!

There are multiple consumers of this code, I see you have staged a review for `ovn-chassis` with this code, have you done the same for the `neutron-openvswitch` charm?

charmhelpers/contrib/openstack/context.py
gabrielcocenza: My idea behind the regular expression it was to be more restrictive to what we consider as an bad input, but looking again the `config_bridges.split()` does an amazing job of removing white spaces between the key-values, been more fault tolerant with the user input.

The only "real gain" that I can see now using regex would be in a situation where the user doesn't give the `ifname_or_mac` like `br-internet:` that would return an empty string to `ifname_or_mac`, probably adding an unmapped interface which I think it's not a big deal.

Thanks for your view. I do agree that omitting the regex is a good idea.

gabrielcocenza:
I think you are totally right of getting out the regex and keeping the tests. Maybe just showing that we expect to raise a `ValueError` when `data-port` is wrong is enough.

>There are multiple consumers of this code, I see you have staged a review for ovn-chassis with this code, have you done the same for the neutron-openvswitch charm?

I didn't. After you saying this I checked what would happen when passed wrong parameters of `data-port` to the `neutron-openvswitch ` and the short answer is it depends. If the config `enable-dpdk` is false, which is the standard, it seems that it just ignores the `data-port` config. 

But if `enable-dpdk` is `true`, it breaks when it tries to use the function `install_packages()` before trying to use the `charm-helpers`. The logs can be found [here](https://pastebin.canonical.com/p/nDwX4SGHdm/), but I think this issue is not part of this PR and I would be glad to talk/help about this later.

What do you think?

fnordahl:
Now that this change has been reduced to valuable test additions we can drop any test requirement for neutron-openvswitch.

Had we made any changes to a context it used a test run, manual or automatic that confirms it still working would be mandatory, even if it is for a non-default configuration option. And to be clear, neutron-openvswitch does indeed use the `data-port` configuration option, but it might use different contexts to parse it depending on other config for historic reasons.

The PR description should probably be updated to reflect what is actually proposed now, and what are your thoughts on the test failures?

gabrielcocenza:
I've updated the name and description of this PR. Regarding the test failures reading the [logs](https://pastebin.canonical.com/p/ngvKWWtgkD/) it seems that is facing issues to install the dependency `distro-info` which has metada issues:
>inconsistent version: filename has '0.10', but metadata has '0.0.0' 

One workaround that I found is to change `install_command` to `pip install {opts} {packages} --use-deprecated=legacy-resolver` in the `tox.ini` file. I didn't understand why keeps installing pip versions above 20.3. Should I include this fix in this PR?

The weird part is that it was running normally in my machine, when I deleted the .tox folder it started showing the same error as here.

fnordahl:
I'd like @AurelienLourot to comment on the proposed fix to unblock the gate as I know we've battled the `distro-info` thing elsewhere recently. Over in [zaza-openstack-tests](https://github.com/openstack-charmers/zaza-openstack-tests/pull/608) we simly remove it, perhaps that could be the fix here to?

fnordahl (approved):

AurelienLourot:
Yes in another repo we are removing `distro-info` as it was unused: https://github.com/openstack-charmers/zaza-openstack-tests/pull/608

However here it is used in the [amulet module](https://github.com/juju/charm-helpers/blob/master/charmhelpers/contrib/amulet/utils.py#L27) . This module is obsolete though and I'm actually suggesting to remove it in https://github.com/juju/charm-helpers/pull/618

fnordahl:
@gabrielcocenza the tests should pass if you rebase this PR

fnordahl:
Ah, thanks for the rebase @AurelienLourot

PR #622, opened 2021-07-28
Allow local testing on Pop!_OS

cjohnston1158:


ajkavanagh:
The tests are failing due to the missing 'distro-info' package which we are seeing everywhere.  We mostly deal with it by pinning it out for the versions that don't exist, but it's probably not that useful on PyPi anyway.

ajkavanagh (approved):
LGTM

ajkavanagh (approved):

PR #623, opened 2021-08-02
Base DPDK socket mem on sockets not NUMA nodes

gnuoy:
OVSDPDKDeviceContext().socket_memory() is currently using the
number of NUMA nodes to generate the memory option for
socket-mem/dpdk-socket-mem *1. However, this option should
be based on the number of sockets *2 & *3. Fwiw, looking at the
charms, socket_memory() only seems to be used to set
dpdk-socket-mem *4 *5. See bug report for more details *6

*1 https://github.com/juju/charm-helpers/blob/master/charmhelpers/contrib/openstack/context.py#L2587
*2 https://doc.dpdk.org/guides/linux_gsg/linux_eal_parameters.html#id3
*3 https://docs.openvswitch.org/en/latest/intro/install/dpdk/
*4 https://github.com/openstack-charmers/charm-layer-ovn/blob/master/lib/charms/ovn_charm.py#L664
*5 https://github.com/openstack/charm-neutron-openvswitch/blob/master/hooks/neutron_ovs_utils.py#L569
*6 https://bugs.launchpad.net/charm-ovn-chassis/+bug/1938557

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -25,7 +25,10 @@ import socket
 import time
 
 from base64 import b64decode
-from subprocess import check_call, CalledProcessError
+from subprocess import (
+    check_call,
+    check_output,
+    CalledProcessError)
 
 import six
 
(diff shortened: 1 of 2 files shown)


gnuoy:
Tested here: https://paste.ubuntu.com/p/sTKNHYvmyF/

javacruft:
I dug into this a bit and concur that this should have always been sockets rather than NUMA node driven.

The majority of deployments where this is in use these will match anyway, so there will be no change to the configuration already applied.

Where deployments exists with multiple NUMA ndoes per socket, they are either not using DPDK or broken due to the associated bug this resolves so I think this is OK to land.

LGTM

javacruft (approved):

PR #624, opened 2021-08-02
Add missing Xena bits

AurelienLourot:
Forgotten in #619


coreycb (changes requested):
I think we also need an update to OPENSTACK_CODENAMES. At least it was updated for wallaby.

AurelienLourot:
@coreycb done already in mentioned #619 ?

coreycb (approved):

PR #626, opened 2021-08-10
Base DPDK socket mem on sockets not NUMA nodes

fnordahl:
OVSDPDKDeviceContext().socket_memory() is currently using the
number of NUMA nodes to generate the memory option for
socket-mem/dpdk-socket-mem *1. However, this option should
be based on the number of sockets *2 & *3. Fwiw, looking at the
charms, socket_memory() only seems to be used to set
dpdk-socket-mem *4 *5. See bug report for more details *6

Closes-Bug: #1938557

*1 https://github.com/juju/charm-helpers/blob/master/charmhelpers/contrib/openstack/context.py#L2587
*2 https://doc.dpdk.org/guides/linux_gsg/linux_eal_parameters.html#id3
*3 https://docs.openvswitch.org/en/latest/intro/install/dpdk/
*4 https://github.com/openstack-charmers/charm-layer-ovn/blob/master/lib/charms/ovn_charm.py#L664
*5 https://github.com/openstack/charm-neutron-openvswitch/blob/master/hooks/neutron_ovs_utils.py#L569
*6 https://bugs.launchpad.net/charm-ovn-chassis/+bug/1938557

(cherry picked from commit 56e2adc84ebff8ba136fb117940c584439b06e34)

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -25,7 +25,10 @@ import socket
 import time
 
 from base64 import b64decode
-from subprocess import check_call, CalledProcessError
+from subprocess import (
+    check_call,
+    check_output,
+    CalledProcessError)
 
 import six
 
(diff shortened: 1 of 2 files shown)


ajkavanagh (approved):
LGTM

Issue #629, opened 2021-08-16
HAProxy exporting metrics


r-f-g:
We could add support for exporting HAProxy statistics to prometheus and then creating a Grafana dashboard.
To export metrics, we could use the `prometheus-exporter`, which has been available since HAProxy version 2.0 [1].

---
[1]: https://www.haproxy.com/blog/haproxy-exposes-a-prometheus-metrics-endpoint/

PR #630, opened 2021-08-16
HAProxy prometheus-exporter

r-f-g:
- add prometheus-exporter to haproxy.cfg template
- add host and port configuration to HAProxyContext
- add a template for the HAProxy dashboard in Grafana
- add function to render dashboard

fixes: #629, [#lp1941762]

---
[#lp1941762]: https://bugs.launchpad.net/charm-helpers/+bug/1941762


charmhelpers/contrib/openstack/ha/utils.py
```diff
@@ -56,6 +57,8 @@
     get_hacluster_config
 )
 
+from charmhelpers.contrib.templating import jinja
```
auria, line 60: This failed when I tested the change because that module is not included in charm-keystone.

charmhelpers/contrib/openstack/templates/grafana-haproxy-dashboard.json.j2
```diff
@@ -0,0 +1,1157 @@
+{
```
auria, line 1: Please have a look to [this other dashboard](https://gist.github.com/auria/ee63b3d988a10273e9cc156c26a58fb6) that I have created based on yours. I have added a "Summary" section with a bit of everything, and instant metrics... and then other sections "Frontends", "Backends" (instead of "Counters"), and the "Responses" and "Servers" that could be extended with more metrics.

I think the title should always be "[juju] HAProxy <application>" (where <application> may be "keystone" or others). The version should be kept on the `version` attribute at the end of the json file, which gets increased.

Other dashboards imported in Grafana via Juju relations have names like "[juju] Telegraf Host Metrics" and such, that's why we should keep "[juju] ..." for consistency. Besides, we want to dashboards to get updated in a transparent way, so having different dashboard titles like "haproxy 1", "haproxy 2", etc. makes explicit something that a user doesn't need to know?

auria (commented):
Overall, changes looks good. I have added some comments inline with some suggestions.

charmhelpers/contrib/openstack/context.py
```diff
@@ -955,10 +957,19 @@ def __call__(self):
...
 
+        # NOTE(rgildein): configure prometheus exporter for haproxy > 2.0.0
+        #                 New bind will be created and a prometheus-exporter
+        #                 will be used for path /metrics. At the same time,
+        #                 prometheus-exporter avoids using auth.
+        haproxy_version = get_installed_version("haproxy")
+        if haproxy_version and haproxy_version.ver_str >= LooseVersion("2.0.0"):
+            ctxt["stats_exporter_host"] = get_relation_ip("prometheus")
+            ctxt["stats_exporter_port"] = \
+                config("haproxy-prometheus-stats-port") or 8404
```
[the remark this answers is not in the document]
r-f-g: I thought about using the `is_relation_made` function, but it occurred to me that charmhelpers should not deal with relations. Thus, it is always exported but only on a private network. Do you think that it's problem to bind it on private network?

charmhelpers/contrib/openstack/ha/utils.py
```diff
@@ -346,3 +351,22 @@ def update_hacluster_vip(service, relation_data):
...
+    """
+    template_dir = os.path.join(os.path.dirname(__file__), "../templates/")
+    datasource = "{} - Juju generated source".format(prometheus_app_name)
+    return jinja.render(GRAFANA_DASHBOARD,
+                        {"datasource": datasource,
+                         "prometheus_app_name": prometheus_app_name,
+                         "dashboard_version": GRAFANA_DASHBOARD_VERSION},
+                        template_dir=template_dir,
+                        jinja_env_args={"variable_start_string": "<< ",
+                                        "variable_end_string": " >>"})
```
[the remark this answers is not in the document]
r-f-g: Yes, it's possible to use `jinja.render` function directly, but I think that dashboard template should be part of charmhelpers. Or what `render` function do you mean?

Good point, I'll add comment to `GRAFANA_DASHBOARD_VERSION` constant.

charmhelpers/contrib/openstack/ha/utils.py
r-f-g: Thanks, I'll add it into `charm-helpers-hooks.yaml` file. (I copied it manually)

charmhelpers/contrib/openstack/templates/grafana-haproxy-dashboard.json.j2
r-f-g: Thanks. You are right, the name should not contain "[juju-keystone]" but only "[juju]". 
As for the version, it must be in the name, as the version in json is automatically increased. e.g. `version=1` and the `grafana-relation-joined` hook is trigger one more time -> `version=2`. This will cause another charm to overwrite the current dashboard to an older version.

r-f-g: I went through [charm-grafana](https://git.launchpad.net/charm-grafana), and I found that function [1] is used to create dashboard title. I can't find a way to prevent the name from being in the form "[juju-keystone] ...".

---
[1]: https://git.launchpad.net/charm-grafana/tree/src/lib/charms/layer/grafana.py#n182

charmhelpers/contrib/openstack/ha/utils.py
ajkavanagh: You could make this an optional import by adding it into the `render_grafana_dashboard(...)` function.  That way it won't break on (classic) charms that don't use the functionality, and charms that call `render_grafana_dashboard(...)` (if they are classic) would need their `charm-helpers-hooks.yaml` file updated anyway.

r-f-g: What do you mean by "optional"? What do you mean by "optional"? Move the import into function or used something like this?
```python
if is_charm_classsic(): 
  from ... import jinja 
esle:
  ...
```

ajkavanagh: Essentially:

```python
def render_grafana_dashboard(...):
    from ... import jinja
    ...
```

i.e. the module is only attempted to be imported if the render_grafana_dashboard() function is actually used.  This means that if a charm doesn't call the function it doesn't need the module synced (as a classic charm).

r-f-g: Ok, that's how I understood it at the beginning, but I wasn't sure.

auria (approved):
LGTM

ajkavanagh (dismissed):
LGTM

charmhelpers/contrib/openstack/context.py
```diff
@@ -955,10 +957,21 @@ def __call__(self):
...
+        #                 New bind will be created and a prometheus-exporter
+        #                 will be used for path /metrics. At the same time,
+        #                 prometheus-exporter avoids using auth.
+        haproxy_version = get_installed_version("haproxy")
+        if (haproxy_version and
+                haproxy_version.ver_str >= LooseVersion("2.0.0") and
+                is_relation_made("haproxy-exporter")):
+            ctxt["stats_exporter_host"] = get_relation_ip("haproxy-exporter")
+            ctxt["stats_exporter_port"] = \
+                config("haproxy-exporter-stats-port") or 8404
```
ajkavanagh, line 973: Just noticed this.  It would be good, here, to verify that the config value is a port and can be turned into an `int`.  e.g. something like

```python
    try:
        ctxt["stats_exporter_port"] = \
            int(config("haproxy-exporter-stats-port")) or 8404
        assert ctxt['stats_exporter_port'] != 0, 'ha-proxy-export-status-port must not be 0'
    except (AssertionError, ValueError, TypeError) as e:
         # do something with the error?  (e.g. raise another exception, or bail out with a log or something else?

```

r-f-g: Thanks, I add `haproxy_exporter_stats_port` property, where I convert the port to integer and if an error occurs I return the default value. I'm logging error messages.

ajkavanagh (approved):
LGTM.  Thanks for all you work on sorting this.

ajkavanagh:
Hey there @r-f-g - one final thought: is there a review that exercises the code or at least shows it doesn't break existing functional tests?  Thanks.

r-f-g:
> Hey there @r-f-g - one final thought: is there a review that exercises the code or at least shows it doesn't break existing functional tests? Thanks.

Hi @ajkavanagh, Hi Alex, I have temporarily added this to my [review](https://review.opendev.org/c/openstack/charm-keystone/+/804760).  We'll see if the tests pass.

ajkavanagh (changes requested):
I think this held due to the comment on https://review.opendev.org/c/openstack/charm-keystone/+/804760:

> This is blocked until specification [this review] is approved.

> [this review]: https://review.opendev.org/c/openstack/charm-specs/+/816909

r-f-g:
The [spec](https://review.opendev.org/c/openstack/charm-specs/+/816909) has been merged, so I think it is good to go. What do you think @ajkavanagh?

ajkavanagh (commented):
Just one minor comment.  Looks great.

charmhelpers/contrib/openstack/context.py
```diff
@@ -865,9 +868,11 @@ class HAProxyContext(OSContextGenerator):
     interfaces = ['cluster']
 
     def __init__(self, singlenode_mode=False,
-                 address_types=ADDRESS_TYPES):
+                 address_types=ADDRESS_TYPES,
```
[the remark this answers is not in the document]
r-f-g: This is good catch, I fixed it.

PR #633, opened 2021-08-17
Include internal endpoint information in IdentityServiceContext

wolsen:
Include the internal endpoint information in IdentityServiceContext.
The keystone service provides the internal information along the
relation but it is not included in the context. Make it available
for consuming charms.

Partial-Bug: #1896438

Signed-off-by: Billy Olsen <billy.olsen@gmail.com>


gnuoy (approved):
LGTM, thanks

PR #635, opened 2021-08-31
Add enable-fwaas to neutron api context

hemanthnakkina:
- Introduces enable-fwaas to neutron api context
- Updates l3_extension_plugins based on new context variable enable-fwaas

fixes: #lp1934129


dosaboy (approved):
lgtm, thanks!

PR #636, opened 2021-08-31
Fix call to Popen.communicate

lmlg:
The call is wrongly passing a string when only bytes-like are allowed.


charmhelpers/contrib/storage/linux/lvm.py
```diff
@@ -60,7 +60,7 @@ def remove_lvm_physical_volume(block_device):
     '''
     p = Popen(['pvremove', '-ff', block_device],
               stdin=PIPE)
-    p.communicate(input='y\n')
+    p.communicate(input=b'y\n')
```
ajkavanagh, line 63: Whilst I'm generally in agreement here, I'm a little worried this might be incompatible across versions of python3.  e.g. the original change was 8 years ago, so it's worked up till today.  I'm going to nack this from landing until we've verified that it works across py3.6 -> py3.9 (?py 3.10?).

ajkavanagh (changes requested):
Holding until it's been verified; the unit tests passing don't necessarily mean that it'll actually work in deployment.  Thanks.

charmhelpers/contrib/storage/linux/lvm.py
wolsen: pvremove supports the `-y` or `--yes` flag from [xenial](https://manpages.ubuntu.com/manpages/xenial/en/man8/pvremove.8.html) through [impish](https://manpages.ubuntu.com/manpages/impish/en/man8/pvremove.8.html). That might be a better change for this.

lmlg: Yes, that does sound better, and it's similar to what `create_logical_volume` does via `lvcreate`. What do you think, @ajkavanagh ?

freyes: I'm +1 on passing `--yes` and replacing the use of `Popen` with `subprocess.check_call`.

freyes (changes requested):
@lmlg , do you have some CI job where this patch is being exercised?

lmlg:
> @lmlg , do you have some CI job where this patch is being exercised?

No, but I've been using it as part of running the functional tests locally for the cinder-lvm charm.

charmhelpers/contrib/storage/linux/lvm.py
```diff
@@ -58,9 +58,7 @@ def remove_lvm_physical_volume(block_device):
 
     :param block_device: str: Full path of block device to scrub.
     '''
-    p = Popen(['pvremove', '-ff', block_device],
-              stdin=PIPE)
-    p.communicate(input=b'y\n')
+    check_call(['pvremove', '-ff', '--yes', block_device])
```
freyes, line 61: This change broke the unit tests, because the unit test mocks `subprocess.Popen`

freyes (changes requested):

charmhelpers/contrib/storage/linux/lvm.py
lmlg: Hadn't seen that. I just fixed it.

freyes:
Rebased on top of master.

freyes (approved):
LGTM. Changes have been addressed

ajkavanagh (approved):
LGTM

PR #638, opened 2021-09-03
Fix failing unit tests

javacruft:
Resolves test failures due to landed commit which lacked the
required updates to the associated unit tests.


AurelienLourot:
FYI this is fixing a test regression introduced by #615

AurelienLourot (approved):

PR #639, opened 2021-09-07
Adding new neutron functions

gabrielcocenza:
Those new functions can be used in some charms, e.g. [charm-neutron-gateway](https://review.opendev.org/c/openstack/charm-neutron-gateway/+/782896/9) and [charm-neutron-openvswitch](https://bugs.launchpad.net/charm-neutron-openvswitch/+bug/1921144).

As soon this PR is merged, those functions will be removed from charm-neutron-gateway and imported from charm-helpers.


ajkavanagh (changes requested):
Following on from my previous comment re adding dependencies to charm-helpers, I've made the suggestion of using the lazy import pattern to only import the required modules in the function that uses them.  This means that other users of the library don't need to have those modules installed, and it pushes the charms that do need the functionality to independently require them.  Thanks.

dosaboy:
Hi Gabriel, thank you for this patch. Would it be possible for you to write a commit message that properly explains the patch and its intended use? You also have multiple commits in this patchset, I assume you plan to squash them into one?

gabrielcocenza:
Hi Edward. I've rebased, squashed my commits into one and updated my commit message.
Thanks.

PR #644, opened 2021-10-11
ceph: make ceph-common version checks optional

javacruft:
Its possible that the ceph module will be used on units where
ceph-common is not installed in order to create requests to
send to the Ceph Broker (hosted by the ceph-mon charm).

Avoid any hard failures when checking the version and assume
and older version that nautilus is actually installed.

### DIFF
--- a/tests/contrib/openstack/test_deferred_events.py
+++ b/tests/contrib/openstack/test_deferred_events.py
@@ -158,7 +158,7 @@ class DeferredCharmServiceEventsTestCase(tests.utils.BaseTestCase):
         get_event_record_file.return_value = test_file
         deferred_events.save_event(self.exp_event_a)
         with open(test_file, 'r') as f:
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -341,7 +340,8 @@ def _post_create(self):
         Do not add calls for a specific pool type here, those should go into
         one of the pool specific classes.
         """
-        if self.nautilus_or_later:
+        nautilus_or_later = cmp_pkgrevno('ceph-common', '14.2.0') >= 0
```
gnuoy, line 343: Would it make more sense to have `nautilus_or_later` a property of the BasePool class, then you don't have to repeat the `cmp_pkgrevno` code ?

gnuoy (approved):

PR #648, opened 2021-10-28
Escape special chars in nrpe command arguments

mkalcok:
Properly escaping arguments for NRPE checks will prevent possible command injection. It will also remove burden of properly escaping such characters on users part. This need to properly escape special characters is also often not apparent to the user. For example, when currently [configuring queue_thresholds](https://github.com/openstack/charm-rabbitmq-server/blob/bfcc4693b94e671f2287c140ce87190f30afec0b/config.yaml#L114) for rabbitmq-server, if user wants to use wildcards, they must be double-escaped like this:
```
juju config rabbitmq-server queue_threshold="[['\\*', '\\*', 100, 200]]"
```

If charmhelpers would automatically take care of escaping these characters when creating NRPE check commands, the above could be simplified to:
```
juju config rabbitmq-server queue_thresholds="[['*', '*', 100, 200]]"
```

### DIFF
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -174,7 +174,8 @@ define service {{
             if os.path.exists(os.path.join(path, parts[0])):
                 command = os.path.join(path, parts[0])
                 if len(parts) > 1:
-                    command += " " + " ".join(parts[1:])
+                    safe_args = [shlex.quote(arg) for arg in parts[1:]]
+                    command += " " + " ".join(safe_args)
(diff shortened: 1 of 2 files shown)


ajkavanagh:
FYI: I put this up to remove py27 from the lib: #651 -- if/when that lands then this can just be rebased to pick up the changes and it should then check okay.

mkalcok:
@ajkavanagh UTs for this change now pass and it's ready for review.

ajkavanagh (approved):
LGTM

mkalcok:
thanks

PR #653, opened 2021-11-15
Avoid using a private attribute from MySQLdb

facundofc:
Bionic ships with mysqlclient version 1.3 and the MySQLdb._exceptions attribute was introduced on 1.4.0 (renamed from MySQLdb._mysql_exceptions, see commit 8ad1525c on mysqlclient's repo).

Even if that wouldn't have changed, it's not good practice to use private attributes of external packages.

Fixes launchpad issue [#1905586](https://bugs.launchpad.net/charm-helpers/+bug/1905586).

### DIFF
--- a/charmhelpers/contrib/database/mysql.py
+++ b/charmhelpers/contrib/database/mysql.py
@@ -729,7 +729,7 @@ class MySQL8Helper(MySQLHelper):
                 remote_ip=remote_ip,
                 password=password)
             )
-        except MySQLdb._exceptions.OperationalError:
+        except MySQLdb.OperationalError:
             log("DB user {} already exists.".format(db_user),
                 "WARNING")
         finally:


AurelienLourot (commented):
LGTM, but can you please open a charm review (charm-percona-cluster?) that will exercise this code, to demonstrate that things still work inside a charm? Thanks!

facundofc:
Thanks Aurelien. The funny thing is that the percona-cluster actually does not exercise this code, because it uses the `MySQLHelper` class from `charmhelpers` (instead of `MySQL8Helper` which this PR is touching). I am not sure why that bug was raised while working on the percona-cluster, I have already commented about this.

I will be opening a charm review soon but for the `mysql-innodb-cluster` charm, which will exercise this piece.

ajkavanagh:
> Thanks Aurelien. The funny thing is that the percona-cluster actually does not exercise this code, because it uses the `MySQLHelper` class from `charmhelpers` (instead of `MySQL8Helper` which this PR is touching). I am not sure why that bug was raised while working on the percona-cluster, I have already commented about this.

@facundofc percona-cluster is pre-mysql8 which is why this code wouldn't be used.; I think @AurelienLourot meant mysql-innodb-cluster.  Please could you add the link to the mysql-innodb-cluster review when it is ready?  Thanks.

facundofc:
[Here](https://review.opendev.org/c/openstack/charm-mysql-innodb-cluster/+/818990)!

facundofc:
Both upstream zuul and zosci passed tests on the PR I cited above.

AurelienLourot (approved):

PR #654, opened 2021-11-16
Add sriov-enabled information to contex

sombrafam:
This will be used to determine when to add a missing option
for SRIOV deployments in change I44fbf5991d2606161b1bcbc064d4bc68a2fbbb5b.

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -3212,6 +3212,9 @@ class SRIOVContext(OSContextGenerator):
         sriov_numvfs = charm_config.get(numvfs_key) or ''
         sriov_device_mappings = charm_config.get(device_mappings_key) or ''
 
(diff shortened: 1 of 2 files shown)


hemanthnakkina:
@sombrafam I see this change is related to https://review.opendev.org/c/openstack/charm-neutron-openvswitch/+/813882.

Just wondering if this change is required at all or not given neutron-openvswitch charm is not supported for releases < xenial.

AurelienLourot (changes requested):
This doesn't seem necessary as this information is being added to the context already here? https://github.com/openstack/charm-neutron-openvswitch/blob/master/hooks/neutron_ovs_context.py#L240

AurelienLourot:
I understand: this was needed in early patchsets of https://review.opendev.org/c/openstack/charm-neutron-openvswitch/+/813882 but we decided to go a different path meanwhile. Closing.

Issue #459, opened 2020-04-21
"functions" is no longer the new name for actions


timClicks:
Actions are now actions again. We should un-deprecate `action_get` and friends. 

https://github.com/juju/charm-helpers/blob/2521285b900e5412c08e83671f63e588ed894b17/charmhelpers/core/hookenv.py#L987-L994

fnordahl:
Fixed by #655

PR #655, opened 2021-11-23
undeprecate action-* functions (#459)

odufourc:
From LP #1926317 also, the action functions should not be set as
deprecated anymore. 
It also set instead the function-get/set/fail as deprecated.
It fixes misleading and unnecessary warning logs.


fnordahl (approved):
Great, thank you for helping us clean this one up!

Pending test results, this LGTM.

PR #659, opened 2021-11-26
Fix get_mon_map() for octopus and later

ajkavanagh:
The "ceph mon_status" command seems to have disappeared on octopus and
later, and is replaced by "ceph quorum_status".  This changes the
get_mon_map() function to detect the underlying ceph version and do the
right thing.

Related bug: https://bugs.launchpad.net/charm-ceph-mon/+bug/1951094


ajkavanagh:
Note, I manually validated this work using a focal-victoria -> focal-xena upgrade.

AurelienLourot (approved):

PR #661, opened 2021-12-06
Add log statement to gain visibility on resolved addresses.

freyes:
This change should allow the debugging of the bug LP: #1952414 which
seems to be related to situations where the endpoint is resolved to a
hostname and later in different calls to an IP address.

Related-Bug: #1952414


coreycb (approved):

coreycb (approved):

ajkavanagh (approved):
LGTM

PR #662, opened 2021-12-09
Support HAproxy https health checks

cornmetto:
This change adds the option of running HTTPS health checks. It is
proposed in the context of LP: #1946280 where the backend radosgw server
can be configured to run in https mode.
We disable certificate verification because we are only interested in
the health of the service.

Related ceph-radosgw change:
* https://review.opendev.org/c/openstack/charm-ceph-radosgw/+/817582

### DIFF
--- a/charmhelpers/contrib/openstack/templates/haproxy.cfg
+++ b/charmhelpers/contrib/openstack/templates/haproxy.cfg
@@ -82,7 +82,11 @@ backend {{ service }}_{{ frontend }}
     {% endif -%}
     {% endif -%}
     {% for unit, address in frontends[frontend]['backends'].items() -%}
+    {% if https -%}
+    server {{ unit }} {{ address }}:{{ ports[1] }} check check-ssl verify none
+    {% else -%}
     server {{ unit }} {{ address }}:{{ ports[1] }} check
+    {% endif -%}
     {% endfor %}
 {% endfor -%}
 {% endfor -%}


auria:
* This change uses the `{% if backend_options[service] -%}` scope to define specific checks (e.g. `/swift/healthcheck`).
* Certificate validation is out of scope of this check (a different nrpe check is used to track expiration)

LGTM.

charmhelpers/contrib/openstack/templates/haproxy.cfg
```diff
@@ -82,7 +82,11 @@ backend {{ service }}_{{ frontend }}
     {% endif -%}
     {% endif -%}
     {% for unit, address in frontends[frontend]['backends'].items() -%}
+    {% if https -%}
+    server {{ unit }} {{ address }}:{{ ports[1] }} check check-ssl verify none
```
wolsen, line 86: This is essentially adding a tls check on the existing tcp check. Would it be better to have an actual http check here?

wolsen: ah ignore me, I see that this is done in the radosgw charm via the backend options. Apologies for the noise

jtroup (approved):
This change looks good to me (and, with it's counterpart ceph-radosgw change, is live on an internal Production cloud).

freyes (approved):
LGTM.

ajkavanagh (approved):
Thanks for the patch.

PR #664, opened 2021-12-21
Use layer 7 checks when apache is used as reverse proxy

drencrom:
When charms are used with HTTPS enabled, an apache server is configured
as a reverse proxy between HAProxy and the charm. In this case the
standard layer 4 HAProxy checks do not work when the charm service is
down but apache is still running.
Now HAProxy is configured to use layer 7 checks in this case so it can
detect that the charm is down even if apache is running.


drencrom:
Duplicated

Issue #665, opened 2022-01-04
Missing openstack release name support in get_os_codename_install_source


wolsen:
Commit 5c377025 added support for specifying simply the OpenStack release name as an installation source, however the charmhelpers.contrib.openstack.utils.get_os_codename_install_source does not support the bare OpenStack release name which prevents the neutron-api charm from installing when configured as such.

ajkavanagh:
This should be pretty easy to fix; I'll rustle up something.

PR #667, opened 2022-01-05
 Ensure get_os_codename_install_source() for bug/665

ajkavanagh:
get_os_codename_install_source() needs to work with bare OpenStack names
e.g "ussuri", rather than "bionic-ussuri".  This patch updates it to
work with bare names.

Fixes: #665


wolsen (approved):
Heh, I had basically the same change :-)

LGTM

PR #668, opened 2022-01-06
Add do_action_package_upgrade function

coreycb:
The do_action_package_upgrade function is similar to
do_action_openstack_upgrade except that it performs package
upgrades only if openstack-origin hasn't changed.

The force_upgrade parameter for do_action_openstack_upgrade is
dropped in favor of the new function for package upgrades.

### DIFF
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -1414,45 +1414,75 @@ def incomplete_relation_data(configs, required_interfaces):
         for i in incomplete_relations}
 
 
-def do_action_openstack_upgrade(package, upgrade_callback, configs,
-                                force_upgrade=False):
+def do_action_openstack_upgrade(package, upgrade_callback, configs):
     """Perform action-managed OpenStack upgrade.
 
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/utils.py
```diff
@@ -1414,45 +1414,75 @@ def incomplete_relation_data(configs, required_interfaces):
...
+                                   'False, skipped upgrade'})
     else:
-        action_set({'outcome': 'no upgrade available.'})
+        action_set({'outcome': 'no upgrade available'})
+
+    return ret
+
+
+def do_action_package_upgrade(package, upgrade_callback, configs):
+    """Perform package upgrade within the current OpenStack release.
```
ajkavanagh, line 1458: Initially I was against the idea of duplicating all of the code, with strong DRY vibes ringing in my head.  But then, the more I thought about it, the more I suspected that this function may change over time (or the other one) and they are both simple enough to be kept in sync as required.  In this case, I think the duplication is the right call.

ajkavanagh (approved):
LGTM

PR #669, opened 2022-01-11
Utilities for configuring deferred restarts NRPE checks

cornmetto:
* Added new NRPE plugin 'check_deferred_restarts.py'. Charms can use
  copy_nrpe_checks() as needed to move it to the nagios plugin directory.
* Added {add,remove}_deferred_restarts_check utilities for adding and
  removing the NRPE checks.

### DIFF
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -28,6 +28,7 @@ import subprocess
 import yaml
 
 from charmhelpers.core.hookenv import (
+    application_name,
     config,
     hook_name,
     local_unit,
@@ -520,3 +521,29 @@ def remove_deprecated_check(nrpe, deprecated_services):
     for dep_svc in deprecated_services:
         log('Deprecated service: {}'.format(dep_svc))
         nrpe.remove_check(shortname=dep_svc)
+
+
+def add_deferred_restarts_check(nrpe):
+    unit_name = local_unit().replace('/', '-')
+    shortname = unit_name + '_deferred_restarts'
+    check_cmd = 'check_deferred_restarts.py --application {}'.format(
+        application_name())
+
+    log('Adding deferred restarts nrpe check: {}'.format(shortname))
+    nrpe.add_check(
+        shortname=shortname,
+        description='Check deferred service restarts {}'.format(unit_name),
+        check_cmd=check_cmd)
+
+
+def remove_deferred_restarts_check(nrpe):
+    unit_name = local_unit().replace('/', '-')
+    shortname = unit_name + '_deferred_restarts'
+    check_cmd = 'check_deferred_restarts.py --application {}'.format(
+        application_name())
+
+    log('Removing deferred restarts nrpe check: {}'.format(shortname))
+    nrpe.remove_check(
+        shortname=shortname,
+        description='Check deferred service restarts {}'.format(unit_name),
+        check_cmd=check_cmd)
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -524,6 +524,11 @@ def remove_deprecated_check(nrpe, deprecated_services):
 
 
 def add_deferred_restarts_check(nrpe):
+    """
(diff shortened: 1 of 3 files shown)


charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -520,3 +521,29 @@ def remove_deprecated_check(nrpe, deprecated_services):
     for dep_svc in deprecated_services:
         log('Deprecated service: {}'.format(dep_svc))
         nrpe.remove_check(shortname=dep_svc)
+
+
+def add_deferred_restarts_check(nrpe):
```
ajkavanagh, line 526: Please could we add docstring comments detailing what the function does, paramters/types, return value/type (if any), and any exceptions that the functions raises?

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -520,3 +521,29 @@ def remove_deprecated_check(nrpe, deprecated_services):
...
+        application_name())
+
+    log('Adding deferred restarts nrpe check: {}'.format(shortname))
+    nrpe.add_check(
+        shortname=shortname,
+        description='Check deferred service restarts {}'.format(unit_name),
+        check_cmd=check_cmd)
+
+
+def remove_deferred_restarts_check(nrpe):
```
ajkavanagh, line 539: As above; docstring please.

charmhelpers/contrib/openstack/files/check_deferred_restarts.py
```diff
@@ -0,0 +1,93 @@
...
+import sys
+import yaml
+
+
+DEFERRED_EVENTS_DIR = '/var/lib/policy-rc.d'
+
+
+def get_deferred_events():
+    """
+    Return a list of deferred events dicts from policy-rc.d files
```
ajkavanagh, line 20: I know it's a touch picky, but please could you put the header of the docstring on the same line as the triple quotes (""")

charmhelpers/contrib/openstack/files/check_deferred_restarts.py
```diff
@@ -0,0 +1,93 @@
...
+
+    services = set(get_deferred_restart_services(args.application))
+
+    if not len(services):
+        print('OK: No deferred service restarts.')
+        sys.exit(0)
+    else:
+        print(
+            'CRITICAL: Restarts are deferred for services: {}.'.format(', '.join(services)))
+        sys.exit(2)
```
ajkavanagh, line 93: You might want to put the whole runtime in a try: except and exit with a 1 if there is a runtime error.  (e.g. yaml.safe_load() fails, can't read a file for some reason).  This would probably be better than just crashing?

ajkavanagh (commented):
Nice patch; just a few thoughts/comments for your consideration.

charmhelpers/contrib/charmsupport/nrpe.py
cornmetto: This was an oversight. They are added now, thanks!

charmhelpers/contrib/charmsupport/nrpe.py
cornmetto: Added, thanks.

charmhelpers/contrib/openstack/files/check_deferred_restarts.py
cornmetto: Yeah error handling was a bit sketchy, thank you. I have improved that as well.

charmhelpers/contrib/openstack/files/check_deferred_restarts.py
cornmetto: Done. I didn't do the same in nrpe.py because the rest of the file has them on different lines. Let me know if you think I should change those as well.

ajkavanagh: No that's fine.  I'm generally in favor of just gradually improving code, but to make those changes as well would probably just confuse the purpose of this PR.  Thanks for make the changes!

ajkavanagh (approved):
LGTM; thanks for the patch.

PR #670, opened 2022-01-13
NRPE: add check: check_sriov_numvfs.py

SPFZ:
Check to verify the number of virtual functions (VFs) active on network interfaces. Based on /sys/class/net/<interface>/device/sriov_numvfs and /sys/class/net/<interface>/device/sriov_totalvfs.

This check is intended to be used in various networking charms like ovn-chassis,  therefore it is added here to avoid replicating the code. 

The goal of the check is to identify network interface configuration problems regarding virtual functions.

Documentation:
```
Check to verify the number of virtual functions (VFs) active on network
interfaces. Based on /sys/class/net/<interface>/device/sriov_numvfs and
/sys/class/net/<interface>/device/sriov_totalvfs.

    usage: check_sriov_numvfs.py [-h] sriov_numvfs [sriov_numvfs ...]

    Check sriov numvfs configuration

    positional arguments:
    sriov_numvfs  format: <interface>:<numvfs>

For each interfaces:numvfs pair given it verifies that:
    1. VFs are enabled
    2. VFs are configured are the same as specified in the check parameter
    3. VFs configured do not exceed the maximum available on the interface

Example: ./check_sriov_numvfs.py ens3f0:32 ens3f1:32 ens6f0:32 ens6f1:32
```

### DIFF
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -520,3 +520,15 @@ def remove_deprecated_check(nrpe, deprecated_services):
     for dep_svc in deprecated_services:
         log('Deprecated service: {}'.format(dep_svc))
         nrpe.remove_check(shortname=dep_svc)
+
+
+def add_sriov_numvfs_checks(nrpe, sriov_numvfs):
+    """
+    Add checks for openvswitch
+    :param NRPE nrpe: NRPE object to add check to
+    :param str sriov_numvfs: space separated list of <interface>:<numvfs>
+    """
+    nrpe.add_check(
+        shortname='sriov_numvfs',
+        description='Check SRIOV numvfs',
+        check_cmd='check_sriov_numvfs.py {}'.format(sriov_numvfs))
--- a/charmhelpers/contrib/network/files/__init__.py
+++ b/charmhelpers/contrib/network/files/__init__.py
@@ -1,0 +1,13 @@
+# Copyright 2014-2015 Canonical Limited.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
--- a/charmhelpers/contrib/network/files/check_sriov_numvfs.py
+++ b/charmhelpers/contrib/network/files/check_sriov_numvfs.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+"""
+Check to verify the number of virtual functions (VFs) active on network
+interfaces. Based on /sys/class/net/<interface>/device/sriov_numvfs and
+/sys/class/net/<interface>/device/sriov_totalvfs.
+
+    usage: check_sriov_numvfs.py [-h] sriov_numvfs [sriov_numvfs ...]
+
+    Check sriov numvfs configuration
+
+    positional arguments:
+    sriov_numvfs  format: <interface>:<numvfs>
+
+For each interfaces:numvfs pair given it verifies that:
+    1. VFs are enabled
+    2. VFs are configured are the same as specified in the check parameter
+    3. VFs configured do not exceed the maximum available on the interface
+
+Example: ./check_sriov_numvfs.py ens3f0:32 ens3f1:32 ens6f0:32 ens6f1:32
+"""
+
+import argparse
+import os.path
+import sys
+
+
+DEVICE_TEMPLATE = "/sys/class/net/{0}"
+SRIOV_NUMVFS_TEMPLATE = "/sys/class/net/{0}/device/sriov_numvfs"
+SRIOV_TOTALVFS_TEMPLATE = "/sys/class/net/{0}/device/sriov_totalvfs"
+
+
+def get_interface_setting(file):
+    """ Return the value content of a setting file as int """
+    with open(file) as f:
+        value = f.read()
+    return int(value)
+
+
+def check_interface_numvfs(iface, numvfs):
+    """ Check SRIOV numvfs config """
+    sriov_numvfs_path = SRIOV_NUMVFS_TEMPLATE.format(iface)
+    sriov_totalvfs_path = SRIOV_TOTALVFS_TEMPLATE.format(iface)
+    msg = []
+    if os.path.exists(DEVICE_TEMPLATE.format(iface)):
+        if (not os.path.exists(sriov_totalvfs_path) or
+                not os.path.exists(sriov_numvfs_path)):
+            msg.append("{}: VFs are disabled or not-available".format(iface))
+        else:
+            sriov_numvfs = get_interface_setting(sriov_numvfs_path)
+            sriov_totalvfs = get_interface_setting(sriov_totalvfs_path)
+            if numvfs != sriov_numvfs:
+                msg.append(
+                    "{}: Number of VFs on interface ({}) does not match check ({})"
+                    .format(iface, sriov_numvfs, numvfs)
+                )
+            if numvfs > sriov_totalvfs:
+                msg.append(
+                    "{}: Maximum number of VFs available on interface ({}) is lower than the check ({})"
+                    .format(iface, sriov_totalvfs, numvfs)
+                )
+    return msg
+
+
+def parse_sriov_numvfs(device_numvfs):
+    """ Parse parameters and check format """
+    msg = "parameter format must be '<interface>:<numvfs>', e.g. ens3f0:32, given; {}".format(device_numvfs)
+    assert device_numvfs != '', msg
+    parts = device_numvfs.split(':')
+    assert len(parts) == 2, msg
+    assert len(parts[0]) > 0, msg
+    assert int(parts[1]) > 0, msg
+    iface = str(device_numvfs.split(':')[0])
+    numvfs = int(device_numvfs.split(':')[1])
+    return (iface, numvfs)
+
+
+def parse_args():
+    """Parse command-line options."""
+    parser = argparse.ArgumentParser(description="Check sriov numvfs configuration")
+    parser.add_argument("sriov_numvfs", nargs='+', help="format: <interface>:<numvfs>")
+    args = parser.parse_args()
+    return args
+
+
+def main():
+    """Parse args and check the sriov numvfs."""
+    args = parse_args()
+    error_msg = []
+
+    for device_numvfs in args.sriov_numvfs:
+        iface, numvfs = parse_sriov_numvfs(device_numvfs)
+        error_msg += check_interface_numvfs(iface, numvfs)
+    if error_msg:
+        print("CRITICAL: {} problems detected\n".format(len(error_msg)) + "\n".join(error_msg))
+        sys.exit(2)
+
+    print("OK: sriov_numvfs set to " + ", ".join(args.sriov_numvfs))
+
+
+if __name__ == "__main__":
+    main()
--- a/tests/contrib/network/files/__init__.py
+++ b/tests/contrib/network/files/__init__.py
@@ -1,0 +1,13 @@
+# Copyright 2014-2015 Canonical Limited.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#  http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
--- a/tests/contrib/network/files/test_check_sriov_numvfs.py
+++ b/tests/contrib/network/files/test_check_sriov_numvfs.py
@@ -1,0 +1,82 @@
+import mock
+import os
+import tempfile
+import unittest
+
+from charmhelpers.contrib.network.files import check_sriov_numvfs
+
+__author__ = 'Stephan Pampel <stephan.pampel@canonical.com>'
+
+
+class TestCheckSriovNumfs(unittest.TestCase):
+
+    interface_folder = os.path.join(tempfile.gettempdir(), 'ens3f0')
+    sriov_numvfs_file = os.path.join(interface_folder, 'sriov_numvfs')
+    sriov_totalvfs_file = os.path.join(interface_folder, 'sriov_totalvfs')
+
+    def __init__(self, *args, **kwargs):
+        super(TestCheckSriovNumfs, self).__init__(*args, **kwargs)
+        if not os.path.exists(self.interface_folder):
+            os.mkdir(self.interface_folder)
+        with open(self.sriov_numvfs_file, "w") as f:
+            f.write("32")
+        with open(self.sriov_totalvfs_file, "w") as f:
+            f.write("64")
+
+    def __del__(self):
+        if os.path.exists(self.interface_folder):
+            for file in os.listdir(self.interface_folder):
+                os.remove(os.path.join(self.interface_folder, file))
+            os.removedirs(self.interface_folder)
+
+    def test_parameter_parsing(self):
+        """ check if the sriov_numvfs parameter is parsed correctly """
+        iface, numvfs = check_sriov_numvfs.parse_sriov_numvfs('ens3f0:32')
+        self.assertEqual(iface, 'ens3f0')
+        self.assertEqual(numvfs, 32)
+
+        for param in ['', ':', ':test', ':32', 'ens3f2', 'a:1:b']:
+            self.assertRaises(AssertionError, check_sriov_numvfs.parse_sriov_numvfs, param)
+
+        for param in ['ens3f2:', 'ens3f2:test']:
+            self.assertRaises(ValueError, check_sriov_numvfs.parse_sriov_numvfs, param)
+
+    def test_check_interface_numvfs_no_interface(self):
+        """ check should report nothing if the interface does not exists """
+        self.assertListEqual(
+            check_sriov_numvfs.check_interface_numvfs('no-interface', 0), []
+        )
+
+    @mock.patch('charmhelpers.contrib.network.files.check_sriov_numvfs.DEVICE_TEMPLATE', interface_folder)
+    def test_check_interface_numvfs_vfs_disabled(self):
+        """ check if virtual functions are disabled """
+        self.assertListEqual(
+            check_sriov_numvfs.check_interface_numvfs('ens3f0', 0),
+            ['ens3f0: VFs are disabled or not-available']
+        )
+
+    @mock.patch('charmhelpers.contrib.network.files.check_sriov_numvfs.DEVICE_TEMPLATE', interface_folder)
+    @mock.patch('charmhelpers.contrib.network.files.check_sriov_numvfs.SRIOV_NUMVFS_TEMPLATE', sriov_numvfs_file)
+    @mock.patch('charmhelpers.contrib.network.files.check_sriov_numvfs.SRIOV_TOTALVFS_TEMPLATE', sriov_totalvfs_file)
+    def test_check_interface_numvfs_vfs_enabled(self):
+        """ check if virtual functions are enabled """
+
+        # check numvfs correct
+        self.assertListEqual(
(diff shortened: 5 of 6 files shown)


charmhelpers/contrib/network/files/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2014-2015 Canonical Limited.
```
jtroup, line 1: Please either update this to 2014-2022, or remove the copyright entirely (it's an empty file)

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
```
jtroup, line 1: Please add a copyright header to this file :-)

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -520,3 +520,15 @@ def remove_deprecated_check(nrpe, deprecated_services):
     for dep_svc in deprecated_services:
         log('Deprecated service: {}'.format(dep_svc))
         nrpe.remove_check(shortname=dep_svc)
+
+
+def add_sriov_numvfs_checks(nrpe, sriov_numvfs):
+    """
+    Add checks for openvswitch
```
jtroup, line 527: Technically this isn't an openvswitch specific check.  Perhaps:

```Add a check for the number of VFs configured for SR-IOV interfaces```

File: charmhelpers/contrib/charmsupport/nrpe.py [hunk 4]
```diff
@@ -520,3 +520,15 @@ def remove_deprecated_check(nrpe, deprecated_services):
     for dep_svc in deprecated_services:
         log('Deprecated service: {}'.format(dep_svc))
         nrpe.remove_check(shortname=dep_svc)
+
+
+def add_sriov_numvfs_checks(nrpe, sriov_numvfs):
```
jtroup, line 525: Any reason this is `checks` rather than `check`?

charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -520,3 +520,15 @@ def remove_deprecated_check(nrpe, deprecated_services):
     for dep_svc in deprecated_services:
```
jtroup, line 520: Please update the copyright year in this file as you're updating it.

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,101 @@
...
+import sys
+
+
+DEVICE_TEMPLATE = "/sys/class/net/{0}"
+SRIOV_NUMVFS_TEMPLATE = "/sys/class/net/{0}/device/sriov_numvfs"
+SRIOV_TOTALVFS_TEMPLATE = "/sys/class/net/{0}/device/sriov_totalvfs"
+
+
+def get_interface_setting(file):
+    """ Return the value content of a setting file as int """
```
jtroup, line 33: Please remove leading and trailing whitespace?

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,101 @@
...
+    with open(file) as f:
```
jtroup, line 34: Do you want to guard against potential errors here (e.g. permission denied or the file not existing)?

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,101 @@
...
+        value = f.read()
+    return int(value)
+
+
+def check_interface_numvfs(iface, numvfs):
+    """ Check SRIOV numvfs config """
```
jtroup, line 40: Please remove leading and trailing whitespace.  Please also s/SRIOV/SR-IOV/

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,101 @@
...
+    sriov_numvfs_path = SRIOV_NUMVFS_TEMPLATE.format(iface)
+    sriov_totalvfs_path = SRIOV_TOTALVFS_TEMPLATE.format(iface)
+    msg = []
+    if os.path.exists(DEVICE_TEMPLATE.format(iface)):
```
jtroup, line 44: While this works, the structure feels a little unwieldly.  I would personally suggest refactoring it, so that rather than getting to 3-4 levels of indentation deep, you instead return early where possible.  e.g. if the device template path doesn't exist, just return msg at that point and then the rest of the function can be de-dented by a level?

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,101 @@
...
+        if (not os.path.exists(sriov_totalvfs_path) or
+                not os.path.exists(sriov_numvfs_path)):
+            msg.append("{}: VFs are disabled or not-available".format(iface))
+        else:
+            sriov_numvfs = get_interface_setting(sriov_numvfs_path)
+            sriov_totalvfs = get_interface_setting(sriov_totalvfs_path)
+            if numvfs != sriov_numvfs:
+                msg.append(
+                    "{}: Number of VFs on interface ({}) does not match check ({})"
```
jtroup, line 53: s/check/expected/ perhaps?

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,101 @@
...
+"""
+
+import argparse
+import os.path
+import sys
+
+
+DEVICE_TEMPLATE = "/sys/class/net/{0}"
+SRIOV_NUMVFS_TEMPLATE = "/sys/class/net/{0}/device/sriov_numvfs"
+SRIOV_TOTALVFS_TEMPLATE = "/sys/class/net/{0}/device/sriov_totalvfs"
```
jtroup, line 29: Hmm - why are we checking this one?  This is the total number of VFs that the PF supports?  Customers could have valid reasons for configuring less VFs than the number the PF supports, e.g. the less VFs you have, the more channels/queues you have per VF.

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,101 @@
...
+            msg.append("{}: VFs are disabled or not-available".format(iface))
+        else:
+            sriov_numvfs = get_interface_setting(sriov_numvfs_path)
+            sriov_totalvfs = get_interface_setting(sriov_totalvfs_path)
+            if numvfs != sriov_numvfs:
+                msg.append(
+                    "{}: Number of VFs on interface ({}) does not match check ({})"
+                    .format(iface, sriov_numvfs, numvfs)
+                )
+            if numvfs > sriov_totalvfs:
```
jtroup, line 56: See comment on or around line 29.

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,101 @@
...
+                msg.append(
+                    "{}: Maximum number of VFs available on interface ({}) is lower than the check ({})"
+                    .format(iface, sriov_totalvfs, numvfs)
+                )
+    return msg
+
+
+def parse_sriov_numvfs(device_numvfs):
+    """ Parse parameters and check format """
```
jtroup, line 65: Please strip leading & trailing whitespace.

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,101 @@
...
+    msg = "parameter format must be '<interface>:<numvfs>', e.g. ens3f0:32, given; {}".format(device_numvfs)
```
jtroup, line 66: s/parameter/Parameter/ please

tests/contrib/network/files/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2014-2015 Canonical Limited.
```
jtroup, line 1: As with other __init__.py comment

tests/contrib/network/files/test_check_sriov_numvfs.py
```diff
@@ -0,0 +1,82 @@
...
+            f.write("64")
+
+    def __del__(self):
+        if os.path.exists(self.interface_folder):
+            for file in os.listdir(self.interface_folder):
+                os.remove(os.path.join(self.interface_folder, file))
+            os.removedirs(self.interface_folder)
+
+    def test_parameter_parsing(self):
+        """ check if the sriov_numvfs parameter is parsed correctly """
```
jtroup, line 33: Whitespace + more sentence like please.  (s/check/Check/)

tests/contrib/network/files/test_check_sriov_numvfs.py
```diff
@@ -0,0 +1,82 @@
...
+        self.assertEqual(numvfs, 32)
+
+        for param in ['', ':', ':test', ':32', 'ens3f2', 'a:1:b']:
+            self.assertRaises(AssertionError, check_sriov_numvfs.parse_sriov_numvfs, param)
+
+        for param in ['ens3f2:', 'ens3f2:test']:
+            self.assertRaises(ValueError, check_sriov_numvfs.parse_sriov_numvfs, param)
+
+    def test_check_interface_numvfs_no_interface(self):
+        """ check should report nothing if the interface does not exists """
```
jtroup, line 45: Id.

tests/contrib/network/files/test_check_sriov_numvfs.py
```diff
@@ -0,0 +1,82 @@
...
+        self.assertListEqual(
+            check_sriov_numvfs.check_interface_numvfs('no-interface', 0), []
+        )
+
+    @mock.patch('charmhelpers.contrib.network.files.check_sriov_numvfs.DEVICE_TEMPLATE', interface_folder)
+    def test_check_interface_numvfs_vfs_disabled(self):
+        """ check if virtual functions are disabled """
```
jtroup, line 52: Id.

tests/contrib/network/files/test_check_sriov_numvfs.py
```diff
@@ -0,0 +1,82 @@
...
+        self.assertListEqual(
+            check_sriov_numvfs.check_interface_numvfs('ens3f0', 0),
+            ['ens3f0: VFs are disabled or not-available']
+        )
+
+    @mock.patch('charmhelpers.contrib.network.files.check_sriov_numvfs.DEVICE_TEMPLATE', interface_folder)
+    @mock.patch('charmhelpers.contrib.network.files.check_sriov_numvfs.SRIOV_NUMVFS_TEMPLATE', sriov_numvfs_file)
+    @mock.patch('charmhelpers.contrib.network.files.check_sriov_numvfs.SRIOV_TOTALVFS_TEMPLATE', sriov_totalvfs_file)
+    def test_check_interface_numvfs_vfs_enabled(self):
+        """ check if virtual functions are enabled """
```
jtroup, line 62: Id.

jtroup (commented):
Great start, thanks.

charmhelpers/contrib/network/files/check_sriov_numvfs.py
SPFZ: This is the number of VFs that the PF supports. 
We are checking this to make sure we are not trying to configure more VFs than the PF supports. This could happen if we have different NICs in a group of servers with different capabilities.
Currently, if the number of VFs set on the PF is higher than totalvfs, it just sets it to totalvfs (silent fail). 

If `numvfs > totalvfs` then `numvfs != sriov_numvfs` will also fail, however the help text will report a more accurate reason and therefore hopefully help to resolve it faster.

charmhelpers/contrib/network/files/check_sriov_numvfs.py
SPFZ: Improved, should be more straightforward now

charmhelpers/contrib/network/files/check_sriov_numvfs.py
SPFZ: changed

charmhelpers/contrib/network/files/check_sriov_numvfs.py
SPFZ: L66 checks if the files do exist. But I added an exception catching wrapper around the check now to be safe.

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,128 @@
...
+            error_msg += check_interface_numvfs(iface, numvfs)
+        if error_msg:
+            print("CRITICAL: {} problems detected\n".format(len(error_msg)) + "\n".join(error_msg))
+            sys.exit(2)
+    except:  # noqa: E722
+        print("{} raised unknown exception '{}'".format(__file__, sys.exc_info()[0]))
+        traceback.print_exc(file=sys.stdout)
+        sys.exit(3)
+
+    print("OK: sriov_numvfs set to " + ", ".join(args.sriov_numvfs))
```
auria, line 124: nitpick: Use `"... {}".format(...)` as in the print within the except scope (and other places) for readability.

auria (changes requested):
Overall, changes look good. I've shared a couple of suggestions. One changes what is considered "unknown" and what is "critical". The other is a readability suggestion.

charmhelpers/contrib/network/files/check_sriov_numvfs.py
SPFZ: done

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,128 @@
...
+    error_msg = []
+
+    try:
+        for device_numvfs in args.sriov_numvfs:
+            iface, numvfs = parse_sriov_numvfs(device_numvfs)
+            error_msg += check_interface_numvfs(iface, numvfs)
+        if error_msg:
+            print("CRITICAL: {} problems detected\n".format(len(error_msg)) + "\n".join(error_msg))
+            sys.exit(2)
+    except:  # noqa: E722
```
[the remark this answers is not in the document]
SPFZ: Thank you for the suggestions, I changed it to match your suggestion.

auria (dismissed):
lgtm, thank you.

charmhelpers/contrib/network/files/check_sriov_numvfs.py
wolsen: Can you elaborate a bit more on why this is a critical alert when the requested VFs cannot be provided by the card? It seems conceivable to me that 2 cards with different capabilities across different units have the same interface name but one can serve 128 VFs and another 64. Setting the value to 128 maxes out both available VFs but it is fewer than 128*2.

wolsen (changes requested):
Looks good at the high level, but a few comments inline.

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+class ArgsFormatError(Exception):
+    """This indicates argument format that is not supported."""
+
+    pass
+
+
+def get_interface_setting(file):
+    """Return the value content of a setting file as int"""
```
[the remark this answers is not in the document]
SPFZ: I added parameter and return docstrings to all functions now, please verify.

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,151 @@
...
+    parts = device_numvfs.split(":")
+    if (
+        len(parts) != 2 or      # exactly 2 parts
+        len(parts[0]) < 3 or    # interface name should be at least 3 chars
+        len(parts[1]) < 1 or    # numvfs should be at least 1 char
+        int(parts[1]) < 0       # numvfs should be a valid int > 0
+    ):
+        raise ArgsFormatError(msg)
+    iface = str(device_numvfs.split(":")[0])
+    numvfs = int(device_numvfs.split(":")[1])
```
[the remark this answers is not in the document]
SPFZ: Great suggestion! I changed it to use regex.

charmhelpers/contrib/network/files/check_sriov_numvfs.py
```diff
@@ -0,0 +1,151 @@
...
+    """Parse parameters and check format"""
+    msg = "Parameter format must be '<interface>:<numvfs>', e.g. ens3f0:32, given: '{}'".format(
+        device_numvfs
+    )
+    parts = device_numvfs.split(":")
+    if (
+        len(parts) != 2 or      # exactly 2 parts
+        len(parts[0]) < 3 or    # interface name should be at least 3 chars
+        len(parts[1]) < 1 or    # numvfs should be at least 1 char
+        int(parts[1]) < 0       # numvfs should be a valid int > 0
```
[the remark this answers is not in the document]
SPFZ: No particular reason. Due to the change to regex this is an ArgsFormatError now.

SPFZ:
@wolsen Thank you for the review! I improved the MR based on your comments.

Regarding the sriov_totalvfs: My assumption is that the number of virtual functions that was requested should be available for use.

If the requested numvfs > sriov_totalvfs then the alert will have two messages:
1. Number of VFs on interface does not match expected (because numvfs != sriov_numvfs L95)
2. Maximum number of VFs available on interface is lower than the expected (because numvfs > sriov_totalvfs L103)

The second message is just a hint for an operator that the reason might be the hardware or hardware configuration that does not support the requested number of VFs. This might be helpful if the cluster hardware is heterogeneous or a NIC has been replace with a different model that supports less VFs (we had a ase like this in the past).

wolsen:
> @wolsen Thank you for the review! I improved the MR based on your comments.
> 
> Regarding the sriov_totalvfs: My assumption is that the number of virtual functions that was requested should be available for use.
> 
> If the requested numvfs > sriov_totalvfs then the alert will have two messages:
> 
> 1. Number of VFs on interface does not match expected (because numvfs != sriov_numvfs L95)
> 2. Maximum number of VFs available on interface is lower than the expected (because numvfs > sriov_totalvfs L103)
> 
> The second message is just a hint for an operator that the reason might be the hardware or hardware configuration that does not support the requested number of VFs. This might be helpful if the cluster hardware is heterogeneous or a NIC has been replace with a different model that supports less VFs (we had a ase like this in the past). I can convert it to a Warning or remove it entirely if this behaviour is not desired but it might lead to a situation when VMs can not be deployed because there are no more VFs available.

To explore it further, for these given checks, what actions are available to the operator with to avoid getting spammed with critical alerts when the heterogeneous scenario occurs?

SPFZ:
> To explore it further, for these given checks, what actions are available to the operator with to avoid getting spammed with critical alerts when the heterogeneous scenario occurs?

If the hardware is heterogeneous by design then an operator would have to create separate applications with different settings. We already have cases like this where were we have different type of machines due to different NICs or customer needs for the machines (with/without SR-IOV). There we deploy separate applications due to differences in hardware related settings (e.g. nova-compute and ovn-chassis).

An alternative would be to make this behavior configurable - essentially disabling checks 1. and 2. from https://github.com/juju/charm-helpers/pull/670#issue-1101375559.

wolsen:
> > To explore it further, for these given checks, what actions are available to the operator with to avoid getting spammed with critical alerts when the heterogeneous scenario occurs?
> 
> If the hardware is heterogeneous by design then an operator would have to create separate applications with different settings. We already have cases like this where were we have different type of machines due to different NICs or customer needs for the machines (with/without SR-IOV). There we deploy separate applications due to differences in hardware related settings (e.g. nova-compute and ovn-chassis).
> 
> An alternative would be to make this behavior configurable - essentially disabling checks 1. and 2. from [#670 (comment)](https://github.com/juju/charm-helpers/pull/670#issue-1101375559).

I think if its known in advance, then the heterogeneous by design could be accounted for. However, in the case of an existing deployment having to consider this I think it places a burden on the operator in this scenario. One option is certainly to make this configurable and this is a worthwhile discussion as to whether it should be configurable and what the default value should be.

However I think there's value in the other checks that are part of this PR and I don't want to hold those up for the discussion on this particular check. I propose that you break out the case where numvfs expected > total_vfs_sriov into another pull request and we can have appropriate discourse separately.

Does that sound reasonable?

SPFZ:
>However I think there's value in the other checks that are part of this PR and I don't want to hold those up for the discussion on this particular check. I propose that you break out the case where numvfs expected > total_vfs_sriov into another pull request and we can have appropriate discourse separately.
> Does that sound reasonable?

Only the "VFs are enabled" check would remain.
> For each interfaces:numvfs pair given it verifies that:
    1. VFs are enabled
    2. VFs are configured are the same as specified in the check parameter
    3. VFs configured do not exceed the maximum available on the interface

Because the second check would still be CRITICAL.
E.g. card supports 64 VFs, check is set to 128 VFs then `numvfs=64 != sriov_numvfs=128`
 https://github.com/SPFZ/charm-helpers/blob/master/charmhelpers/contrib/network/files/check_sriov_numvfs.py#L95

If that is needed to get this PR to move forward then I can remove the checks 2 and 3.

PR #672, opened 2022-01-24
Move documentation for how to use this module to the docstring so it's rendered on RTD

mthaddon:
Currently the documentation on https://charm-helpers.readthedocs.io/en/latest/api/charmhelpers.contrib.charmsupport.html only renders as one line that's not very useful. Move the documentation from comments lower down in the file to the docstring.


erik78se:
Might also consider mentioning that this will not build for charms using setuptools > 58 which effectively will break all builds with charmcraft as of today.

So, basically, its broken with charmcraft.

https://github.com/canonical/charmcraft/issues/632#issuecomment-1019346554

PR #673, opened 2022-02-07
Add profile-name parameter to replicated pool

aqsamm:
This change adds a profile name parameter in the create pool function that
allows a replicated pool to be created with a CRUSH profile other than
the default 'replicated_rule'.

### DIFF
--- a/charmhelpers/contrib/storage/linux/ceph.py
+++ b/charmhelpers/contrib/storage/linux/ceph.py
@@ -614,7 +614,8 @@ class Pool(BasePool):
 
 class ReplicatedPool(BasePool):
     def __init__(self, service, name=None, pg_num=None, replicas=None,
-                 percent_data=None, app_name=None, op=None):
+                 percent_data=None, app_name=None, op=None,
+                 profile_name='replicated_rule'):
         """Initialize ReplicatedPool object.
 
         Pool information is either initialized from individual keyword
@@ -631,6 +632,8 @@ class ReplicatedPool(BasePool):
                          to this replicated pool.
         :type replicas: int
         :raises: KeyError
+        :param profile_name: Crush Profile to use
+        :type profile_name: Optional[str]
         """
         # NOTE: Do not perform initialization steps that require live data from
         # a running cluster here. The *Pool classes may be used for validation.
@@ -645,11 +648,20 @@ class ReplicatedPool(BasePool):
             # we will fail with KeyError if it is not provided.
             self.replicas = op['replicas']
             self.pg_num = op.get('pg_num')
+            self.profile_name = op.get('crush-profile', profile_name)
         else:
             self.replicas = replicas or 2
             self.pg_num = pg_num
+            self.profile_name = profile_name or 'replicated_rule'
 
     def _create(self):
+        # Validate if crush profile exists
+        if self.profile_name is None:
+            msg = ("Failed to discover crush profile named "
+                   "{}".format(self.profile_name))
+            log(msg, level=ERROR)
+            raise PoolCreationError(msg)
+
         # Do extra validation on pg_num with data from live cluster
         if self.pg_num:
             # Since the number of placement groups were specified, ensure
@@ -667,12 +679,12 @@ class ReplicatedPool(BasePool):
                 '--pg-num-min={}'.format(
                     min(AUTOSCALER_DEFAULT_PGS, self.pg_num)
                 ),
-                self.name, str(self.pg_num)
+                self.name, str(self.pg_num), self.profile_name
             ]
         else:
             cmd = [
                 'ceph', '--id', self.service, 'osd', 'pool', 'create',
-                self.name, str(self.pg_num)
+                self.name, str(self.pg_num), self.profile_name
             ]
         check_call(cmd)
 
@@ -691,7 +703,7 @@ class ErasurePool(BasePool):
     def __init__(self, service, name=None, erasure_code_profile=None,
                  percent_data=None, app_name=None, op=None,
                  allow_ec_overwrites=False):
-        """Initialize ReplicatedPool object.
+        """Initialize ErasurePool object.
 
         Pool information is either initialized from individual keyword
         arguments or from a individual CephBrokerRq operation Dict.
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -645,11 +647,20 @@ def __init__(self, service, name=None, pg_num=None, replicas=None,
             # we will fail with KeyError if it is not provided.
             self.replicas = op['replicas']
             self.pg_num = op.get('pg_num')
+            self.profile_name = op.get('crush-profile')
```
ajkavanagh, line 650: You may want to do this instead?

```python
self.profile_name = op.get('crush-profile', profile_name or 'replicated_rule')
```

i.e. always ensure that the profile name exists even if it's not in the `op`; otherwise 'older' users of the library might not have updated their 'ops' to use the `crush-profile` key?

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -1856,6 +1867,10 @@ def add_op_create_replicated_pool(self, name, replica_count=3, pg_num=None,
                        for pool.
         :type pg_num: int
         :raises: AssertionError if provided data is of invalid type/range
+        :param crush_profile: Name of crush profile to use. If not set the
+                              ceph-mon unit handling the broker request will
+                              set its default value.
+        :type crush_profile: str
```
ajkavanagh, line 1873: Either `:type crush_profile: Optional[str]` or require `crush_profile` in the function signature?

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -1865,6 +1880,7 @@ def add_op_create_replicated_pool(self, name, replica_count=3, pg_num=None,
             'name': name,
             'replicas': replica_count,
             'pg_num': pg_num,
+            'crush-profile': crush_profile
```
ajkavanagh, line 1883: Does there need to be a default here (or is this now covered by my other comment above?)

ajkavanagh (commented):
Please see inline comments.

charmhelpers/contrib/storage/linux/ceph.py
aqsamm: I think this one is resolved with the suggestion above

ajkavanagh (approved):
LGTM; someone more familiar with ceph may want to take a look too.  Thanks for the changes!

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -614,7 +614,7 @@ def create(self):
 
 class ReplicatedPool(BasePool):
     def __init__(self, service, name=None, pg_num=None, replicas=None,
-                 percent_data=None, app_name=None, op=None):
+                 percent_data=None, app_name=None, profile_name=None, op=None):
```
ChrisMacNaughton, line 617: I'd rather the new argument (`profile_name`) be added to the end of this list, and also to have the function signature include the default:

```
    def __init__(self, service, name=None, pg_num=None, replicas=None,
                 percent_data=None, app_name=None, op=None, profile_name='replicated_rule'):
```

Is there any strong reason why it was done the other way?

ChrisMacNaughton (commented):
Bit of a nitpick on the __init__ changes for replicated pools

charmhelpers/contrib/storage/linux/ceph.py
aqsamm: None, I thought this was already covered with line 656

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -645,11 +648,21 @@ def __init__(self, service, name=None, pg_num=None, replicas=None,
             # we will fail with KeyError if it is not provided.
             self.replicas = op['replicas']
             self.pg_num = op.get('pg_num')
+            self.profile_name = op.get('crush-profile', profile_name or
```
ChrisMacNaughton, line 651: With the default in the function above, you don't need the `or` bit

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -631,6 +632,8 @@ def __init__(self, service, name=None, pg_num=None, replicas=None,
                          to this replicated pool.
         :type replicas: int
         :raises: KeyError
+        :param profile_name: Crush Profile to use (default: 'replicated_rule')
```
ChrisMacNaughton, line 635: With the default in the function above, the function signature documents this so isn't needed in a comment

charmhelpers/contrib/storage/linux/ceph.py
ajkavanagh: > I'd rather the new argument (profile_name) be added to the end of this list, and also to have the function signature include the default:

With named params the order generally doesn't matter, and so I let that one through.

re: default.  Personally, I **much** prefer people to signal 'None' as a default parameter as the type can then be `Optional[str]` rather than `str` (which is a good signal that it is optional).  By setting the default if None is passed, it's really obvious in the code that the default is being used.  Finally, it stops (forever) the risk that someone will accidentally put an internally-mutable reference as a default leading to all sorts of shenanigans.

e.g.

```python
def some_function(arg=[]):
    arg.append("thing")
    print(arg)
```

:)

So, I'd like optional arguments to still be `None`.

ChrisMacNaughton: I'm very aware of mutable references as defaults; _however_, this is a `str` and, while is has a default, it isn't really optional after this change. It's entirely OK to leave it as the default but a `None`, in this case, isn't actually true. The final resting place of this argument looks like:

```python
cmd = [
    'ceph', '--id', self.service, 'osd', 'pool', 'create',
    self.name, str(self.pg_num), self.profile_name
]
```

Given that the `profile_name` isn't actually optional, I'd prefer if the type signature actually tells me what the default is rather than having to read the function's docstrings to figure out what the default profile name is.

ChrisMacNaughton (approved):
With the caveat that this library should be both refactored and, more importantly, removed from Charmhelpers, this change seems like a reasonable change in the short-term to accomplish the goal. Thanks!

PR #674, opened 2022-02-08
Remove tempita as it doesn't seem to be required anymore

ajkavanagh:
It is a dependency that has been hanging around since 2014.  Let's get rid of it as it breaks when 2to3 tools are missing.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,6 +10,5 @@ PyYAML
 Jinja2
 
 netaddr
-Tempita
 
 pbr!=2.1.0,>=2.0.0 # Apache-2.0


freyes:
:+1:  LGTM, Tempita hasn't been needed since 2014 when [this commit](https://github.com/juju/charm-helpers/commit/f68c655c915f5be432a9b7cb2bf51095aa8242d6) removed the module that was using it and since then tempita has been being carried around as dependency and everyone who changed the building scripts was including Tempita just because it was there.

ChrisMacNaughton (approved):

PR #676, opened 2022-02-23
Adding new function for enabling service

agileshaw:
* add a service_enable function in charmhelpers/core/host.py
* add a unit test for service_enable function in tests/core/test_host.py

Some charms, such as charm-glance (concerning bug/1960636), requires calling
an additional function during installation to enable services. This new function
can be used to address such a problem.

### DIFF
--- a/charmhelpers/core/host.py
+++ b/charmhelpers/core/host.py
@@ -114,6 +114,33 @@ def service_stop(service_name, **kwargs):
     return service('stop', service_name, **kwargs)
 
 
+def service_enable(service_name, **kwargs):
+    """Enable a system service.
+
+    The specified service name is managed via the system level init system.
+    Some init systems (e.g. upstart) require that additional arguments be
+    provided in order to directly control service instances whereas other init
+    systems allow for addressing instances of a service directly by name (e.g.
+    systemd).
+
+    The kwargs allow for the additional parameters to be passed to underlying
+    init systems for those systems which require/allow for them. For example,
+    the ceph-osd upstart script requires the id parameter to be passed along
+    in order to identify which running daemon should be restarted. The follow-
+    ing example restarts the ceph-osd service for instance id=4:
+
+    service_enable('ceph-osd', id=4)
+
+    :param service_name: the name of the service to enable
+    :param **kwargs: additional parameters to pass to the init system when
+                     managing services. These will be passed as key=value
+                     parameters to the init system's commandline. kwargs
+                     are ignored for init systems not allowing additional
+                     parameters via the commandline (systemd).
+    """
+    return service('enable', service_name)
+
+
 def service_restart(service_name, **kwargs):
     """Restart a system service.
 
@@ -134,7 +161,7 @@ def service_restart(service_name, **kwargs):
     :param service_name: the name of the service to restart
     :param **kwargs: additional parameters to pass to the init system when
                      managing services. These will be passed as key=value
-                     parameters to the  init system's commandline. kwargs
+                     parameters to the init system's commandline. kwargs
                      are ignored for init systems not allowing additional
                      parameters via the commandline (systemd).
     """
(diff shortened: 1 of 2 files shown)


charmhelpers/core/host.py
```diff
@@ -114,6 +114,33 @@ def service_stop(service_name, **kwargs):
...
+    ing example restarts the ceph-osd service for instance id=4:
+
+    service_enable('ceph-osd', id=4)
+
+    :param service_name: the name of the service to enable
+    :param **kwargs: additional parameters to pass to the init system when
+                     managing services. These will be passed as key=value
+                     parameters to the init system's commandline. kwargs
+                     are ignored for init systems not allowing additional
+                     parameters via the commandline (systemd).
```
ajkavanagh, line 139: This is misleading; the `**kwargs` aren't actually used in the function, but this docstring implies that they are.  If `**kwargs` are added as a compatibility, then this docstring should say that they are ignored.  In which case, why accept them at all?  I suspect that they shouldn't be included in the function signature at all?

ajkavanagh (changes requested):
Thanks for the patch. Please see inline comment, as I have a query about the function signature of the added function.  Thanks.

charmhelpers/core/host.py
agileshaw: Hello Alex. Thanks for the review. The `**kwargs` in the function is indeed solely for compatibility. Do you suggest to remove it completely or just updating the signature?

ajkavanagh: I think, as it's a new function, the `**kwargs` should just be dropped from the signature and the docstring; they don't do anything and it would be an error if a user of the function called it with a kwarg expecting it to do something with it.  Much better to fail hard, rather than silently pass but perhaps propagate a semantic error further into the run time of the code.

agileshaw: Sounds great. I have pushed a new commit addressing out discussion. I also noticed similar issues with service_restart function. I will open another PR afterwards.

coreycb:
Looks good and Alex's comments have been handled appropriately. Thank you!

ajkavanagh (dismissed):

tests/core/test_host.py
```diff
@@ -184,6 +184,14 @@ def test_stops_a_service(self, service):
         self.assertTrue(host.service_stop(service_name))
 
         service.assert_called_with('stop', service_name)
+
```
ajkavanagh, line 187: According to pep8 for py35 this line failed with:

```
tests/core/test_host.py:187:1: W293 blank line contains whitespace
```

agileshaw: Thanks for the review. I have just pushed another commit fixing this issue.

charmhelpers/core/host.py
```diff
@@ -114,6 +114,17 @@ def service_stop(service_name, **kwargs):
...
 
+def service_enable(service_name):
+    """Enable a system service.
+
+    The specified service name is managed via the system level init system.
+    Enabling a service allows it to run automatically on boot.
+
+    :param service_name: the name of the service to enable
+    """
+    return service('enable', service_name)
```
ajkavanagh, line 125: So, I've made a mistake and need to ask you to change direction on this please.  I hadn't looked sufficiently around the code to notice that I was steering you in the wrong direction, for which I apologise, particularly as I'm asking you to re-work the patch please.

Essentially, looking at the other functions (`service_restart`, etc) the `**kwargs` argument is passed to the `service(...)` function and *may* have effects there.

So the ask is to restore the `**kwargs` to the function signature *and* pass them to as `service('enable', service_name, **kargs)`.

This also means restoring the docstring (a cut and paste from one of the other functions), as well.

Thanks very much!

ajkavanagh (changes requested):

ajkavanagh (approved):
LGTM.  Thank you for the patch.

PR #677, opened 2022-03-09
Make ansible-playbook verbose

xavpaice:
The code that runs ansible-playbook doesn't give us much information when  a playbook fails.  This change adds a verbose switch to the CLI to let us get a lot more debugging information.


ajkavanagh (approved):
LGTM.  Thanks for the patch.

PR #678, opened 2022-03-11
collect stdout from ansible-playbook execution

xavpaice:
When Ansible playbooks fail, log the stdout so that we can read it and find out why the playbook failed.

### DIFF
--- a/charmhelpers/contrib/ansible/__init__.py
+++ b/charmhelpers/contrib/ansible/__init__.py
@@ -218,7 +218,8 @@ def apply_playbook(playbook, tags=None, extra_vars=None):
         subprocess.check_output(call, env=env)
     except subprocess.CalledProcessError as e:
         charmhelpers.core.hookenv.log("Ansible playbook failed with {} "
-                                      "Stdout: {}".format(e, e.output))
+                                      "Stdout: {}".format(e, e.output),
+                                      level="ERROR")
         raise subprocess.CalledProcessError(e)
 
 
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/ansible/__init__.py
```diff
@@ -214,7 +214,12 @@ def apply_playbook(playbook, tags=None, extra_vars=None):
         call.extend(['--tags', '{}'.format(tags)])
     if extra_vars:
         call.extend(['--extra-vars', json.dumps(extra_vars)])
-    subprocess.check_call(call, env=env)
+    try:
+        subprocess.check_output(call, env=env)
+    except subprocess.CalledProcessError as e:
+        charmhelpers.core.hookenv.log("Ansible playbook failed with {} "
+                                      "Stdout: {}".format(e, e.output))
+        raise subprocess.CalledProcessError(e)
```
ajkavanagh, line 222: A minor thought here: it might be useful to log to level ERROR instead of the default (INFO, I think?) so that Ansible playbook errors show more prominently in the log?

xavpaice: Thanks! Yes, since we're logging for the exception only, it makes sense for this to be at ERROR level.  I wonder if we should also log the stdout for successful runs at DEBUG level?  Not sure we want to flood the logs every time an update-status hook runs.

Anyway, I've made that one change.

ajkavanagh: > Thanks! Yes, since we're logging for the exception only, it makes sense for this to be at ERROR level. I wonder if we should also log the stdout for successful runs at DEBUG level? Not sure we want to flood the logs every time an update-status hook runs.

Yeah, I tend to agree that logging on success is a bit of an anti-pattern, and just makes finding problems harder.  My rule-of-thumb is to log things that changed materially, or things that failed.  However, DEBUG is very useful for debugging issues, and the default for models is ERROR IIRC, so it might be appropriate to log that it was successful at DEBUG, and use TRACE for the actual stdout?  However, that can be for another PR! :)

ajkavanagh (approved):
LGTM.  Thanks for the patch.

PR #687, opened 2022-03-23
Add service_type to keystone-authtoken section

dosaboy:
This also moves the service types map here from the keystone
charm so that it can be used for all charms.

Related-Bug: #1965967


gnuoy (approved):
LGTM, thanks

PR #688, opened 2022-03-29
Make inability to determine Openstack codename version non-fatal

canonical-xavd:
This change allows `error_out` callers to specify a function signalling an error, appropriate to it's context.

Related change: https://review.opendev.org/c/openstack/charm-keystone/+/835087
Related issue: [LP#1965966](https://bugs.launchpad.net/charm-keystone/+bug/1965966/)

### DIFF
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -400,13 +400,16 @@ def get_os_codename_version(vers):
         error_out(e)
 
 
-def get_os_version_codename(codename, version_map=OPENSTACK_CODENAMES):
+def get_os_version_codename(codename, version_map=OPENSTACK_CODENAMES,
+                            raise_exception=False):
     '''Determine OpenStack version number from codename.'''
     for k, v in version_map.items():
         if v == codename:
             return k
     e = 'Could not derive OpenStack version for '\
         'codename: %s' % codename
+    if raise_exception:
+        raise ValueError(str(e))
     error_out(e)
 
 
--- a/tests/contrib/openstack/test_openstack_utils.py
+++ b/tests/contrib/openstack/test_openstack_utils.py
@@ -280,6 +280,13 @@ class OpenStackHelpersTestCase(TestCase):
         expected_err = 'Could not derive OpenStack version for codename: foo'
         mocked_error.assert_called_with(expected_err)
 
+        try:
+            openstack.get_os_version_codename('foo', raise_exception=True)
+            raise Exception("Failed call should have raised ValueError")
+        except ValueError as e:
+            self.assertEquals(e.args[0],
+                              "Could not derive OpenStack version for codename: foo")
+
     def test_os_version_swift_from_codename(self):
         """Test mapping a swift codename to numerical version"""
         self.assertEquals(openstack.get_os_version_codename_swift('liberty'),


charmhelpers/contrib/openstack/utils.py
```diff
@@ -407,7 +411,7 @@ def get_os_version_codename(codename, version_map=OPENSTACK_CODENAMES):
             return k
     e = 'Could not derive OpenStack version for '\
         'codename: %s' % codename
-    error_out(e)
+    error_out(e, raiser=ValueError)
```
ajkavanagh, line 414: This seems like a lot of work to go to?  couldn't you just do:

```python
raise ValueError(str(e))
```

instead if you wanted a change in behaviour here?

ajkavanagh (changes requested):
I'm currently pushing back on this, as the name of the function `error_out` is pretty unambiguous that it's going to quit.  I think this change blurs the meaning using an optional param which isn't very explicit.

charmhelpers/contrib/openstack/utils.py
canonical-xavd: I think it stemmed from an abundance of caution in context of the specific case I'm proposing this change for to not break other charms, but now that it has spilled over to the common codebase, this is indeed more sensible to do.

charmhelpers/contrib/openstack/utils.py
```diff
@@ -407,7 +407,7 @@ def get_os_version_codename(codename, version_map=OPENSTACK_CODENAMES):
             return k
     e = 'Could not derive OpenStack version for '\
         'codename: %s' % codename
-    error_out(e)
+    raise ValueError(str(e))
```
ajkavanagh, line 410: Sorry, it's been a while since I got back to this.  This is a still a change in behaviour.  Previously, this function killed the charm hook dead (using `error_out()`).  The proposal is to change it to a `ValueError`.  Now all the charms that assumed that it would die at that point, now get an exception that they haven't got code to handle.

What are you trying to achieve please?  Do you have some sample code that uses this function and wants to 'handle' this exception when `get_os_codename_version()` is called?

The usual way of handling this is to do something like this:

```python
def get_os_version_codename(
        codename,
        version_map=OPENSTACK_CODENAMES,
        raise_exception=False):
    '''Determine OpenStack version number from codename.'''
    for k, v in version_map.items():
        if v == codename:
             return k
    e = 'Could not derive OpenStack version for '\
        'codename: %s' % codename
    if raise_exception:
        raise ValueError(str(e))
    error_out(e)


# use it, but with an exception
try:
    codename = get_os_version_codename(codename, raise_exception=True)
except ValueError:
    # do something with the fact no version maps to the codename

This way the old behaviour is preserved (for charms expecting it), and the new behaviour can be used by the charms wanting the exception rather than exiting the hook.
```

canonical-xavd: The contribution this stems from aims to handle Openstack release codename inference, when it's not possible to deduce it from apt repository specification while using a custom repository, per linked related change to the Keystone charm and linked Launchpad issue, which also serve as code samples demonstrating intent.

Based on the previous proposed change, your current proposition gets half of the way there, as `error_out` would be parameterized to no longer exclusively fatally `sys.exit()`, but not allow for costumization of this behavior where it may be contextually appropriate, while preserving currently expected behavior (per lambda provided as default function argument value in the original changeset revision).

ajkavanagh: I'm obviously not explaining my point well.  The problem with the code in the linked example is this:

```python
def get_os_version_codename(codename, version_map=OPENSTACK_CODENAMES):
    '''Determine OpenStack version number from codename.'''
    for k, v in six.iteritems(version_map):
        if v == codename:
            return k
    e = 'Could not derive OpenStack version for '\
        'codename: %s' % codename
    error_out(e, raiser=ValueError)
```

The last part *changes the behaviour of* `get_os_version_codename()` which other charms rely on.  This means that `get_os_version_codename()` now raises a `ValueError`, and doesn't do `sys.exit()` because `error_out()` is also changed.  These are identical in the `get_os_version_codename()` function as changed above.

```python
error_out(e, raiser=ValueError)
# or
raise ValueError(str(e))
```
Which makes it worse as it now redefines what error_out does, which is no longer "erroring out of the hook", but "raise a ValueError instead, if told".  Semantically, these things are different and *will* lead to confusion in the future.  It matters what things are named and that this matches what the thing does.

That's the problem.  Please don't change the default behaviour of `get_os_version_codename()`.

ajkavanagh (approved):
LGTM - thanks for the updates.

PR #691, opened 2022-04-01
Validate X509 certificate inputs

canonical-xavd:
This change attempts to validate `ssl_*` configuration options and relations for valid X509 certificates, as well as verify their trust chains.

Related change: https://review.opendev.org/c/openstack/charm-openstack-dashboard/+/836797
Related issue: [LP#1772674](https://bugs.launchpad.net/charm-openstack-dashboard/+bug/1772674)

### DIFF
--- a/charmhelpers/contrib/hahelpers/apache.py
+++ b/charmhelpers/contrib/hahelpers/apache.py
@@ -24,6 +24,8 @@
 
 import os
 
+from base64 import b64decode
+
 from charmhelpers.core import host
 from charmhelpers.core.hookenv import (
     config as config_get,
@@ -31,7 +33,12 @@ from charmhelpers.core.hookenv import (
     relation_ids,
     related_units as relation_list,
     log,
+    ERROR,
     INFO,
+)
+from charmhelpers.contrib.openstack.cert_utils import (
+    x509_get_pubkey,
+    x509_validate_cert,
 )
 
 # This file contains the CA cert from the charms ssl_ca configuration
@@ -61,6 +68,11 @@ def get_cert(cn=None):
                 if not key:
                     key = relation_get(ssl_key_attr,
                                        rid=r_id, unit=unit)
+
+    # this likely fails too quietly, raise?
+    if not x509_validate_cert(b64decode(cert), ssl_key=b64decode(key)):
+        return (None, None)
+
     return (cert, key)
 
 
@@ -75,6 +87,11 @@ def get_ca_cert():
                 if ca_cert is None:
                     ca_cert = relation_get('ca_cert',
                                            rid=r_id, unit=unit)
+
+    # this likely fails too quietly, raise?
+    if not x509_get_pubkey(b64decode(ca_cert)):
+        return None
+
     return ca_cert
 
 
--- a/charmhelpers/contrib/openstack/cert_utils.py
+++ b/charmhelpers/contrib/openstack/cert_utils.py
@@ -16,7 +16,11 @@
 
 import os
 import json
+import subprocess
+import tempfile
 from base64 import b64decode
+
+import six
 
 from charmhelpers.contrib.network.ip import (
     get_hostname,
@@ -34,6 +38,7 @@ from charmhelpers.core.hookenv import (
     log,
     WARNING,
     INFO,
+    ERROR,
 )
 from charmhelpers.contrib.openstack.ip import (
     resolve_address,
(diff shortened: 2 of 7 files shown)


charmhelpers/contrib/hahelpers/apache.py
```diff
@@ -88,3 +105,76 @@ def retrieve_ca_cert(cert_file):
 
 def install_ca_cert(ca_cert):
     host.install_ca_cert(ca_cert, CONFIG_CA_CERT_FILE)
+
+
+def _x509_normalize_input(input_):
```
ChrisMacNaughton, line 110: why name this argument `input_`?

charmhelpers/contrib/hahelpers/apache.py
```diff
@@ -61,6 +68,11 @@ def get_cert(cn=None):
                 if not key:
                     key = relation_get(ssl_key_attr,
                                        rid=r_id, unit=unit)
+
+    # this likely fails too quietly
```
ChrisMacNaughton, line 72: Should it have logging, then?

ChrisMacNaughton (commented):
It seems like most of the things in hahelpers/apache.py should be in cert_utils.py?

charmhelpers/contrib/hahelpers/apache.py
canonical-xavd: What's meant by "too quietly" applies to code's logic - there could be a case that false-value these returns could be simply used in conditions deteremining whether to configure TLS at all, which poses a huge red flag for the existing codebase.

All functions called by `x509_validate_cert` to validate parity and trust chains already have arguably sufficient logging about why things may have been going awry.

charmhelpers/contrib/hahelpers/apache.py
canonical-xavd: Primarily because the function is meant to normalize whatever value it receives into a bytestring - predominantly they're supposed to be `unicode/str` and `bytes`, but the last case simply returns a value for any other type that's going to fail downstream from this call.

charmhelpers/contrib/openstack/cert_utils.py
```diff
@@ -441,3 +448,79 @@ def get_bundle_for_cn(cn, relation_name=None):
...
+    cmd = ['openssl', 'verify']
+
+    with tempfile.TemporaryFile() as cert_fd, tempfile.NamedTemporaryFile() as ca_fd:
+        cert_fd.write(ssl_cert); cert_fd.seek(0)  # noqa: E702
+        if ssl_ca:
+            ca_fd.write(ssl_ca); ca_fd.seek(0)  # noqa: E702
+            cmd.extend(['-CAfile', ca_fd.name])
+
+        try:
+            subprocess.check_output(cmd, stdin=cert_fd)
```
nobuto-m, line 471: If I'm not mistaken, this always fails when `ssl_cert` consists of a leaf cert + intermediate CA(s). `openstack verify` only reads the first certificate of the chain in the file and ignores the rest.

If we take this openssl CLI approach, you need to split the content of `ssl_ca` into a leaf + the rest of the chain and do:
`openstack verify -CAfile "<ssl_ca>" -untrusted "<the rest of ssl_cert>" "<the first cert in ssl_cert"`.

`openstack verify -CAfile "<ssl_ca>" -untrusted "<ssl_cert>" "<ssl_cert>"` technically works to be simpler even though it's with a duplicate cert in the chain and it doesn't exactly match what Apache2 is doing.

nobuto-m (changes requested):
I left a comment on the point where the check always fails with intermediate CA(s).

In the meantime, SSL/TLS validation is always hard to implement properly and in the right way. So we have a slightly different approach in #718 (we didn't notice this PR before working on it, sorry!) by leveraging `asn1crypto` python library. The downside is to pull another dependency in charm-helpes but the upside is less effort in the charm-helpers itself and can give some user-friendly message on errors. You can see some example output in:
https://gist.github.com/nobuto-m/e7b2a7ab9f7206254b4dbe2f5aa48edd

charmhelpers/contrib/openstack/cert_utils.py
canonical-xavd: I've actually been aware of intermediate certificates posing a minor issue with verification using `openssl`, as you've outlined, which primarily requires in-place splitting of PEM chains and well-considered test cases to properly exemplify check sanity. I just haven't had a sufficient time block to update the proposal to address this.

If there is a way to fix this, I would consider splitting away the leaf as a proper solution.

PR #692, opened 2022-04-05
Ease KeepAliveTimeout in line with keystoneauth1.session.Session

nobuto-m:
Apache2's default value for KeepAliveTimeout is 5 seconds, which is okay
for general web-page serving use cases. However, sessions and connection
pools created by keystoneauth1.session.Session can be terminated
unnecessarily during multiple API calls in a session due to the short
KeepAliveTimeout.

Let's ease KeepAliveTimeout to 75 seconds, which is fairly standard for
API services behind a reverse proxy since it's the default value of
nginx.

Closes-Bug: #1947010
(cherry picked from commit 9a5bf40790080362cf340692b8992f31a6750cda)


wolsen (approved):
LGTM

PR #700, opened 2022-04-26
ArchiveUrlFetchHandler.download(...) needs to be able to access proxy variables

addyess:


javacruft:
Generally I think this looks good - Do we also need to plug noproxy configuration in as well?

addyess:
> Generally I think this looks good - Do we also need to plug noproxy configuration in as well?

I'll have to come up with another solution for noproxy.  I looked at the implementation of `ProxyHandler` and it seems that if you provide a value of `proxies={'no': no_proxies}` that it's not respected in the `proxy_open` method installed in `urllib.request`.  ONLY the os.environ variables are respected

See [for yourself](https://github.com/python/cpython/blob/a03a09e068435f47d02649dda93988dc44ffaaf1/Lib/urllib/request.py#L812)


```python
    def proxy_open(self, req, proxy, type):
        orig_type = req.type
        proxy_type, user, password, hostport = _parse_proxy(proxy)
        if proxy_type is None:
            proxy_type = orig_type

        if req.host and proxy_bypass(req.host):
            return None
```

you'd like that `proxy_bypass` method to use `self.proxies` like this:

```python
        if req.host and proxy_bypass(req.host, self.proxies):
            return None
```

but it doesn't. 

The only other solution i see currently is to alter my implementation here to temporarily insert the `JUJU_CHARM*` proxy variables into the `os.environ`

PR #703, opened 2022-04-29
Modify git tests so they continue to work with git < 2.28

ajkavanagh:
git 2.28 introduced the `--initial-branch` option to
`git init` which allowed the initial branch to be other than
master.  This was used in the GitUrlFetchHandlerTest class.
Unfortunately, this breaks when testing on a distro that doesn't
have git < 2.28 installed.  Modify the tests so that this is called
out explicitly.  When git pre-2.28 is gone, this patch can be
reverted.

### DIFF
--- a/tests/fetch/test_giturl.py
+++ b/tests/fetch/test_giturl.py
@@ -1,4 +1,5 @@
 import os
+from pkg_resources import parse_version
 import shutil
 import subprocess
 import tempfile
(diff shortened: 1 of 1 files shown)


tests/fetch/test_giturl.py
```diff
@@ -36,6 +37,22 @@ def setUp(self):
             "garbage",
         )
         self.fh = giturl.GitUrlFetchHandler()
+        self._git_version_2_28_plus = None
```
addyess, line 40: sigh -- this is why `@cached_property` should have been a thing long before `python3.8`

ajkavanagh: Indeed.  C-h will bump major version nums as each py3.x goes EOL so that eventually we can use the newer features.

addyess (approved):
You got my +1

wolsen (approved):
LGTM

PR #705, opened 2022-05-09
Fix `openstack_upgrade_available` failing on unconventional source

canonical-xavd:
An unfortunate follow-up to #688 after digging up [a relevant comment](https://review.opendev.org/c/openstack/charm-keystone/+/835087/1/charmhelpers/contrib/openstack/utils.py#407) clarifying reasons being the previous PR.

Fixes `openstack_upgrade_available` calls unexpectedly `sys.exit(1)`ing, while expecting to be thrown an exception, whenever the package source specification does not contain an Openstack release codename.

Related change: https://review.opendev.org/c/openstack/charm-keystone/+/835087
Related: LP#1965966

### DIFF
--- a/charmhelpers/contrib/openstack/utils.py
+++ b/charmhelpers/contrib/openstack/utils.py
@@ -386,9 +386,9 @@ def get_os_codename_install_source(src):
                 return v
 
 
-def get_os_version_install_source(src):
+def get_os_version_install_source(src, raise_exception=False):
     codename = get_os_codename_install_source(src)
-    return get_os_version_codename(codename)
+    return get_os_version_codename(codename, raise_exception=raise_exception)
 
 
 def get_os_codename_version(vers):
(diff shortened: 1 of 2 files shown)


UtkarshBhatthere (approved):
The changes look fine to me.

charmhelpers/contrib/openstack/utils.py
```diff
@@ -849,7 +849,8 @@ def openstack_upgrade_available(package):
         avail_vers = get_os_version_codename_swift(codename)
     else:
         try:
-            avail_vers = get_os_version_install_source(src)
+            avail_vers = get_os_version_install_source(src,
+                                                       raise_exception=True)
```
ajkavanagh, line 853: I've spent a while staring at this, as it was going around my head, and I then realised that this is changing the nature of `openstack_upgrade_availble()` from the existing behaviour of erroring out (using `error_out()` in `get_os_version_codename()`) to a new behaviour of raising an exception (`ValueError`).

To fix this, this  function (`openstack_upgrade_avialble()`) should *also* take an optional `raise_exception=False` parameter, and whatever calls it should pass in `raise_exception=True` to get the new behaviour.  Otherwise, a whole load of charms may get different behaviour from what they are expecting currently.

ajkavanagh (changes requested):
Unfortunately, this can't be merged as it changes the behaviour of `openstack_upgrade_behaviour()`. Please see inline comment.  Note, also that the doctring for the function also needs to be changed to indicate the new functionality available.  Thanks.

PR #711, opened 2022-05-26
Remove check_haproxy if haproxy service is monitored by pacemaker

gabrielcocenza:
check_haproxy is redundant if haproxy is managed by pacemaker.
check_crm nrpe alert is better suited because it avoids false
positives alerts when units are paused.

See [LP#1880601](https://bugs.launchpad.net/charm-hacluster/+bug/1880601) for more details.

### DIFF
--- a/charmhelpers/contrib/charmsupport/nrpe.py
+++ b/charmhelpers/contrib/charmsupport/nrpe.py
@@ -19,6 +19,7 @@
 
 import glob
 import grp
+import json
 import os
 import pwd
 import re
@@ -30,6 +31,7 @@ import yaml
 from charmhelpers.core.hookenv import (
     application_name,
     config,
+    ERROR,
     hook_name,
     local_unit,
     log,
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/charmsupport/nrpe.py
```diff
@@ -416,6 +417,15 @@ def add_init_service_checks(nrpe, services, unit_name, immediate_check=True):
     :param str unit_name: Unit name to use in check description
     :param bool immediate_check: For sysv init, run the service check immediately
     """
+    # check_haproxy is redundant in the presence of check_crm. See LP Bug#1880601 for details.
+    # just remove check_haproxy if haproxy is added as a lsb resource in hacluster.
+    for rid in relation_ids("ha"):
+        ha_resources = relation_get("json_resources", rid=rid, unit=local_unit())
+        if ha_resources:
+            if "lsb:haproxy" in json.loads(ha_resources).values():
```
ajkavanagh, line 425: `json.loads(...)` may raise an error; would be good to put this in a `try: except:` block to catch the error and do a nice log message if there is a problem.  e.g. should the hook crash if this fails?

ajkavanagh (changes requested):

charmhelpers/contrib/charmsupport/nrpe.py
gabrielcocenza: > should the hook crash if this fails?

I think it should crash because something is very wrong if the ha relation doesn't set a string that can be converted into a dict.
 
I've added the try except block and added another unit test to raise the error.

ajkavanagh (approved):
LGTM; thanks for the changes.

PR #713, opened 2022-06-03
docs/api: add package charmhelpers.contrib.database

mert-kirpici:
Add the templates for mysql module as well.

Signed-off-by: Mert Kirpici <mert.kirpici@canonical.com>


ajkavanagh (approved):
LGTM - thanks for the document update; these are much appreciated as they are so often overlooked.  :)

PR #715, opened 2022-06-08
Rotate the request_id on CephBrokerReq with operation changes

### LINKED ISSUE #714: Ceph broker requests can end up cached despite changing
Due to how the broker requests handle their `request-id`s  (https://github.com/juju/charm-helpers/blob/master/charmhelpers/contrib/storage/linux/ceph.py#L1705) combined with Ceph bailing on requests where a request-id has "finished" means that the mon charm may skip new operations that have been added to a broker request. The request-id could, instead, be a hash of the ops on the request which would cause the request-id to change when ops does.

ChrisMacNaughton:
When the operations requested change, the request_id must
change as well, kor ceph-mon will not process the newly
added operations.

Closes #714

### DIFF
--- a/charmhelpers/contrib/storage/linux/ceph.py
+++ b/charmhelpers/contrib/storage/linux/ceph.py
@@ -28,7 +28,6 @@ import os
 import shutil
 import json
 import time
-import uuid
 
 from subprocess import (
     check_call,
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -1695,14 +1698,16 @@ def __init__(self, api_version=1, request_id=None, raw_request_data=None):
...
+            self.request_id = request_data['request-id']
         else:
             self.api_version = api_version
             if request_id:
                 self.request_id = request_id
             else:
-                self.request_id = str(uuid.uuid1())
+                # The below hash is the result of running
+                # `hashlib.sha1('[]'.encode()).hexdigest()`
+                self.request_id = '97d170e1550eee4afc0af065b78cda302a97674c'
```
lmlg, line 1710: Personally, I'd move the value to a top-level declaration. Makes it easier to change and mock, IMO.

charmhelpers/contrib/storage/linux/ceph.py
```diff
@@ -1713,6 +1718,7 @@ def add_op(self, op):
         """
         if op not in self.ops:
             self.ops.append(op)
+            self.request_id = hashlib.sha1(json.dumps(self.ops, sort_keys=True).encode()).hexdigest()
```
lmlg, line 1721: Much like my other comment, I think it's worthy moving this to a function, since it's used more than once, and it makes testing it easier.

lmlg (changes requested):

lmlg (approved):

PR #717, opened 2022-06-14
Changes to the default inode size

FrancescoDeSimone:
The inode size default value is in contradiction with the current upstream advice.
This patch removes the default value and accept the filesystem defaults for mkfs.xfs inode sizes.

Related Bug #1879423

### DIFF
--- a/charmhelpers/contrib/storage/linux/utils.py
+++ b/charmhelpers/contrib/storage/linux/utils.py
@@ -15,6 +15,10 @@
 import os
 import re
 from stat import S_ISBLK
+from charmhelpers.core.hookenv import (
+    log,
+    WARNING
+)
 
 from subprocess import (
     CalledProcessError,
@@ -118,12 +122,16 @@ def mkfs_xfs(device, force=False, inode_size=1024):
     :ptype device: tr
     :param force: Force operation
     :ptype: force: boolean
-    :param inode_size: XFS inode size in bytes
+    :param inode_size: XFS inode size in bytes; if set to 0 or None,
+        the value used will be the XFS system default
     :ptype inode_size: int"""
     cmd = ['mkfs.xfs']
     if force:
         cmd.append("-f")
-    if inode_size:
+    if inode_size >= 256 and inode_size <= 2048:
         cmd += ['-i', "size={}".format(inode_size)]
+    elif inode_size != 0:
+        log("Config value xfs-inode-size={} is invalid. Using system default.".format(inode_size), level=WARNING)
+
     cmd += [device]
     check_call(cmd)
--- a/charmhelpers/contrib/storage/linux/utils.py
+++ b/charmhelpers/contrib/storage/linux/utils.py
@@ -15,10 +15,6 @@
 import os
 import re
 from stat import S_ISBLK
-from charmhelpers.core.hookenv import (
-    log,
-    WARNING
-)
 
 from subprocess import (
     CalledProcessError,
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/storage/linux/utils.py
```diff
@@ -123,6 +123,7 @@ def mkfs_xfs(device, force=False, inode_size=1024):
     cmd = ['mkfs.xfs']
     if force:
         cmd.append("-f")
-
-    cmd += ['-i', "size={}".format(inode_size), device]
+    if inode_size != 0:
```
gnuoy, line 126: As suggested in the companion review you can just test for truthiness here `if inode_size:`

gnuoy (approved):

gnuoy:
Seems to be a py310 unit test issue. I don't think it's related to this change

charmhelpers/contrib/storage/linux/utils.py
```diff
@@ -123,6 +123,7 @@ def mkfs_xfs(device, force=False, inode_size=1024):
     cmd = ['mkfs.xfs']
     if force:
         cmd.append("-f")
-
-    cmd += ['-i', "size={}".format(inode_size), device]
+    if inode_size:
```
gnuoy, line 126: Can you amend the doc string above so it makes it clear inode_size can be set None to accept the xfs default setting ?

gnuoy (changes requested):

charmhelpers/contrib/storage/linux/utils.py
```diff
@@ -110,7 +110,7 @@ def is_device_mounted(device):
     return bool(re.search(r'MOUNTPOINT=".+"', out))
 
 
-def mkfs_xfs(device, force=False, inode_size=1024):
+def mkfs_xfs(device, force=False, inode_size=0):
```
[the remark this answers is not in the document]
sombrafam: So, this is not something that affects only swift. The OS recommended inode_size is recommended regardless of the charm. There are a few other charms that use this function (e.g.charm-nova-compute, kubernetes-master) but will also benefit from this fix. And they still could overwrite the default value (hardcoded) if they had any specific case, which I don't think is the case.

The suggestion to use 0, is that this is an `int` value in the swift-charm, I didn't notice that this function was shared. But, using, `mkfs_xfs(device, force=False, inode_size=None)` would make a better semantics. We could then change the swift code to adapt that.

charmhelpers/contrib/storage/linux/utils.py
sombrafam: @FrancescoDeSimone not sure if you saw my last comment. We should use `mkfs_xfs(device, force=False, inode_size=None)`

charmhelpers/contrib/storage/linux/utils.py
```diff
@@ -132,8 +128,6 @@ def mkfs_xfs(device, force=False, inode_size=None):
     if inode_size:
         if inode_size >= 256 and inode_size <= 2048:
             cmd += ['-i', "size={}".format(inode_size)]
-        elif inode_size != 0:
-            log("Config value xfs-inode-size={} is invalid. Using system default.".format(inode_size), level=WARNING)
```
sombrafam, line 131: Hey, I think I wasn't clear in my sentence. We need to remove the check, but leave the warn so the user knows that he is using a bad value:

```python
if inode_size >= 256 and inode_size <= 2048:
    cmd += ['-i', "size={}".format(inode_size)]
else:
    log("Config value xfs-inode-size={} is invalid. Using system default.".format(inode_size), level=WARNING)
```

gnuoy (approved):
Thank you for your work on this

PR #718, opened 2022-06-16
Validate certificate chain for apache2 ssl

swalladge:
https://bugs.launchpad.net/charm-helpers/+bug/1978902

Add certificate validation during the process to configure certificates for apache.  The certificates are the values provided to the well-known charm config keys (ssl_ca, ssl_cert, ssl_key).   This should catch issues with certificates earlier, provide more friendly error messages, and help with debugging certificate issues.

### DIFF
--- a/charmhelpers/contrib/hahelpers/apache.py
+++ b/charmhelpers/contrib/hahelpers/apache.py
@@ -35,9 +35,6 @@ from charmhelpers.core.hookenv import (
     related_units as relation_list,
     log,
     INFO,
-    ERROR,
-    status_set,
-    WORKLOAD_STATES,
 )
 
 # This file contains the CA cert from the charms ssl_ca configuration
@@ -72,12 +69,12 @@ def get_cert(cn=None):
 
 def validate_cert(cert, key, ca):
     """
-    cert (bytes): cert.crt file contents (PEM armored)
-    key (bytes): cert.key file contents
-    ca (Optional[bytes]): None or private ca contents
+    cert (bytes): PEM armored cert chain
+    key (bytes): private key for leaf cert
+    ca (Optional[bytes]): an optional PEM armored root CA chain
 
-    return True if ok
-    otherwise return False, set blocked status, and set useful error message
+    returns: None if all ok, otherwise return a string error message
+    rtype: Optional[str]
     """
 
     context = None
@@ -88,22 +85,14 @@ def validate_cert(cert, key, ca):
     try:
         certs = list(map(lambda x: x[2], pem.unarmor(cert, multiple=True)))
     except Exception as e:
-        log("Failed to read certificates: {}".format(e),
-            level=ERROR)
-        status_set(WORKLOAD_STATES.BLOCKED, str(e))
-        return False
+        return "[processing cert chain] {}".format(e)
 
     validator = CertificateValidator(certs[0], certs[1:], validation_context=context)
 
     try:
         validator.validate_usage({"digital_signature"}, {"server_auth"})
     except Exception as e:
-        log("Failed to validate certificates: {}".format(e),
-            level=ERROR)
-        status_set(WORKLOAD_STATES.BLOCKED, str(e))
-        return False
-
-    return True
+        return "[validating cert usage] {}".format(e)
 
 
 def get_ca_cert():
--- a/charmhelpers/contrib/hahelpers/apache.py
+++ b/charmhelpers/contrib/hahelpers/apache.py
@@ -67,10 +67,9 @@ def get_cert(cn=None):
     return (cert, key)
 
 
-def validate_cert(cert, key, ca):
+def validate_cert(cert, ca):
     """
     cert (bytes): PEM armored cert chain
-    key (bytes): private key for leaf cert
     ca (Optional[bytes]): an optional PEM armored root CA chain
 
     returns: None if all ok, otherwise return a string error message
--- a/charmhelpers/contrib/hahelpers/apache.py
+++ b/charmhelpers/contrib/hahelpers/apache.py
@@ -35,6 +35,7 @@ from charmhelpers.core.hookenv import (
     related_units as relation_list,
     log,
     INFO,
+    ERROR,
     status_set,
     WORKLOAD_STATES,
 )
@@ -87,6 +88,8 @@ def validate_cert(cert, key, ca):
     try:
         certs = list(map(lambda x: x[2], pem.unarmor(cert, multiple=True)))
     except Exception as e:
+        log("Failed to read certificates: {}".format(e),
+            level=ERROR)
         status_set(WORKLOAD_STATES.BLOCKED, str(e))
         return False
 
@@ -95,6 +98,8 @@ def validate_cert(cert, key, ca):
     try:
         validator.validate_usage({"digital_signature"}, {"server_auth"})
     except Exception as e:
+        log("Failed to validate certificates: {}".format(e),
+            level=ERROR)
         status_set(WORKLOAD_STATES.BLOCKED, str(e))
         return False
 
--- a/charmhelpers/contrib/hahelpers/apache.py
+++ b/charmhelpers/contrib/hahelpers/apache.py
@@ -24,6 +24,9 @@
 
 import os
 
+from asn1crypto import pem
+from certvalidator import CertificateValidator, ValidationContext
+
 from charmhelpers.core import host
 from charmhelpers.core.hookenv import (
     config as config_get,
@@ -32,6 +35,8 @@ from charmhelpers.core.hookenv import (
     related_units as relation_list,
     log,
     INFO,
+    status_set,
+    WORKLOAD_STATES,
 )
 
 # This file contains the CA cert from the charms ssl_ca configuration
@@ -62,6 +67,38 @@ def get_cert(cn=None):
                     key = relation_get(ssl_key_attr,
                                        rid=r_id, unit=unit)
     return (cert, key)
+
+
+def validate_cert(cert, key, ca):
+    """
+    cert (bytes): cert.crt file contents (PEM armored)
+    key (bytes): cert.key file contents
+    ca (Optional[bytes]): None or private ca contents
+
+    return True if ok
+    otherwise return False, set blocked status, and set useful error message
+    """
+
+    context = None
+    if ca:
+        context = ValidationContext(trust_roots=[ca])
+
+    # split certs so we can pass intermediate certs separately to the validator
+    try:
+        certs = list(map(lambda x: x[2], pem.unarmor(cert, multiple=True)))
+    except Exception as e:
+        status_set(WORKLOAD_STATES.BLOCKED, str(e))
+        return False
+
+    validator = CertificateValidator(certs[0], certs[1:], validation_context=context)
+
+    try:
+        validator.validate_usage({"digital_signature"}, {"server_auth"})
+    except Exception as e:
+        status_set(WORKLOAD_STATES.BLOCKED, str(e))
+        return False
+
+    return True
 
 
 def get_ca_cert():
--- a/charmhelpers/contrib/hahelpers/apache.py
+++ b/charmhelpers/contrib/hahelpers/apache.py
@@ -79,7 +79,11 @@ def validate_cert(cert, key, ca):
 
     context = None
     if ca:
-        context = ValidationContext(trust_roots=[ca])
+        try:
+            cas = list(map(lambda x: x[2], pem.unarmor(ca, multiple=True)))
+        except Exception as e:
+            return "[processing root ca chain] {}".format(e)
+        context = ValidationContext(extra_trust_roots=cas)
 
     # split certs so we can pass intermediate certs separately to the validator
     try:
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -89,6 +89,7 @@ from charmhelpers.contrib.hahelpers.apache import (
     get_cert,
     get_ca_cert,
     install_ca_cert,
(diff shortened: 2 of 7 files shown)


charmhelpers/contrib/hahelpers/apache.py
```diff
@@ -64,6 +70,36 @@ def get_cert(cn=None):
...
+    otherwise return False, set blocked status, and set useful error message
+    """
+
+    context = None
+    if ca:
+        context = ValidationContext(trust_roots=[ca])
+
+    # split certs so we can pass leaf and intermediate certs separately
+    # The leaf/server certificate _must_ be the first cert https://www.rfc-editor.org/rfc/rfc5246#page-48
+    certs = re.findall(r'-----BEGIN CERTIFICATE-----.*?-----END CERTIFICATE-----', cert.decode(), re.DOTALL)
```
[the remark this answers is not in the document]
nobuto-m: Yes, we can assume PEM format.
https://httpd.apache.org/docs/2.4/mod/mod_ssl.html#sslcertificatefile

Also, I found in the parent library that `multiple=True` can split the concatenated certs into an array in DER format.
https://github.com/wbond/asn1crypto/blob/master/docs/pem.md#unarmor
So we could use `certs[0]` as a leaf cert, then the rest `cert[1], [2], ...` can be used as an intermediate CA chain. We might be able to use DER as is since CertificateValidator states it can accept DER and PEM or we need to armor those back to PEM.

tests/contrib/hahelpers/test_apache_utils.py
```diff
@@ -131,3 +131,129 @@ def test_retrieve_ca_cert_no_file(self, _isfile):
...
+DTIyMDYxNTIzMzg0N1oXDTMyMDYxMjIzMzg0N1owODESMBAGA1UEChMJTG9jYWxo
+b3N0MSIwIAYDVQQDExlMb2NhbGhvc3QgSW50ZXJtZWRpYXRlIENBMFkwEwYHKoZI
+zj0CAQYIKoZIzj0DAQcDQgAEKpD0lYmyDLX+guoi2WQupkRpYydpkT0fihJ9Asqp
+bQ0j8SK6XOgG17wIIynR/PchZPBnTew8poi1TOV6Sf195KNmMGQwDgYDVR0PAQH/
+BAQDAgEGMBIGA1UdEwEB/wQIMAYBAf8CAQAwHQYDVR0OBBYEFLCsW22w0eU1gSNQ
+qq+Kec9VH/5QMB8GA1UdIwQYMBaAFL7pE7+4bPZwOYJeDceK/B2Ev4iSMAoGCCqG
+SM49BAMCA0kAMEYCIQC3NxTe3PxTiF0W1L39xrHvPvFCIHkI65ThQWM+upgNIAIh
+AJD2IG2GS7MCxD5EiG4Y3cMcPjye4qud4OduSbnA8UO4
+-----END CERTIFICATE-----
+'''.encode()
```
nobuto-m, line 162: It might be better to add a header to PEM as follows to be easier to review or revisit later on.

With `openssl x509 -subject -issuer`, this block can be:

```
subject=CN = localhost
issuer=O = Localhost, CN = Localhost Intermediate CA
-----BEGIN CERTIFICATE-----
MIICKTCCAdCgAwIBAgIRAJ94pSPwW8j47TxyEbuy11cwCgYIKoZIzj0EAwIwODES
MBAGA1UEChMJTG9jYWxob3N0MSIwIAYDVQQDExlMb2NhbGhvc3QgSW50ZXJtZWRp
YXRlIENBMB4XDTIyMDYxNTIzMzgwMVoXDTIyMDYxNjIzMzkwMVowFDESMBAGA1UE
AxMJbG9jYWxob3N0MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAExkWwyfOpgiAh
oClU2ETkZ+C2Q5FD81f4nk6IDRCZ0JEhXl6b5t1z02F0+aA3FZRkmlJBuGv9Oi+v
AJ8R2Yh1BaOB3jCB2zAOBgNVHQ8BAf8EBAMCB4AwHQYDVR0lBBYwFAYIKwYBBQUH
AwEGCCsGAQUFBwMCMB0GA1UdDgQWBBQXIcm+FtQbyWuRivPP9xTVgCIp6zAfBgNV
HSMEGDAWgBSwrFttsNHlNYEjUKqvinnPVR/+UDAUBgNVHREEDTALgglsb2NhbGhv
c3QwVAYMKwYBBAGCpGTGKEABBEQwQgIBAQQQdGVzdEBleGFtcGxlLmNvbQQrZFRH
VC14NUtycHpRN1M3VnIxSVpnMVh4cVRuRi1INml3M1lqMVZoS1JxWTAKBggqhkjO
PQQDAgNHADBEAiAyQVuIbUuYKcps8EolwLL7HVbwVMdK06rvOqEIcNnpCAIgdEFi
8XsMn9hJmDiunM8RAd0aDVa1cC+aizu5mOgrBbc=
-----END CERTIFICATE-----
subject=O = Localhost, CN = Localhost Intermediate CA
issuer=O = Localhost, CN = Localhost Root CA
-----BEGIN CERTIFICATE-----
MIIBzTCCAXKgAwIBAgIQcT2oOYh7TNlbbto/RdvZGTAKBggqhkjOPQQDAjAwMRIw
EAYDVQQKEwlMb2NhbGhvc3QxGjAYBgNVBAMTEUxvY2FsaG9zdCBSb290IENBMB4X
DTIyMDYxNTIzMzg0N1oXDTMyMDYxMjIzMzg0N1owODESMBAGA1UEChMJTG9jYWxo
b3N0MSIwIAYDVQQDExlMb2NhbGhvc3QgSW50ZXJtZWRpYXRlIENBMFkwEwYHKoZI
zj0CAQYIKoZIzj0DAQcDQgAEKpD0lYmyDLX+guoi2WQupkRpYydpkT0fihJ9Asqp
bQ0j8SK6XOgG17wIIynR/PchZPBnTew8poi1TOV6Sf195KNmMGQwDgYDVR0PAQH/
BAQDAgEGMBIGA1UdEwEB/wQIMAYBAf8CAQAwHQYDVR0OBBYEFLCsW22w0eU1gSNQ
qq+Kec9VH/5QMB8GA1UdIwQYMBaAFL7pE7+4bPZwOYJeDceK/B2Ev4iSMAoGCCqG
SM49BAMCA0kAMEYCIQC3NxTe3PxTiF0W1L39xrHvPvFCIHkI65ThQWM+upgNIAIh
AJD2IG2GS7MCxD5EiG4Y3cMcPjye4qud4OduSbnA8UO4
-----END CERTIFICATE-----
```

charmhelpers/contrib/hahelpers/apache.py
```diff
@@ -64,6 +70,36 @@ def get_cert(cn=None):
...
+    certs = list(map(lambda x: x.encode(), certs))
+
+    validator = CertificateValidator(certs[0], certs[1:], validation_context=context)
+
+    try:
+        validator.validate_usage({"digital_signature"}, {"server_auth"})
+    except Exception as e:
+        status_set(WORKLOAD_STATES.BLOCKED, str(e))
```
nobuto-m, line 97: Probably this status_set has only a temporary effect. Most of the OpenStack charms call assess_status() or equivalent at the end of the hook execution and that will override any status set during the hook execution.

How to set the status appropriately might be a next step, but logging an error or warning is definitely one step forward than the current status.

charmhelpers/contrib/hahelpers/apache.py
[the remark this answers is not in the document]
nobuto-m: Well, the second example in `Basic Path Validation` section is exactly for it :-)
https://github.com/wbond/certvalidator/blob/master/docs/usage.md#basic-path-validation

charmhelpers/contrib/openstack/context.py
```diff
@@ -1032,6 +1033,13 @@ def configure_cert(self, cn=None):
         mkdir(path=ssl_dir)
         cert, key = get_cert(cn)
         if cert and key:
+            cert = b64decode(cert)
+            key = b64decode(key)
+            ca = config('ssl_ca')
+            ca = b64decode(ca) if ca else None
+            if not validate_cert(cert, key, ca):
+                return
```
nobuto-m, line 1041: Returning here before configuring certs might be too drastic. There might be corner cases where the underlying library doesn't work as expected so we might be able to proceed by giving a warning for the time being.

charmhelpers/contrib/hahelpers/apache.py
```diff
@@ -64,6 +70,42 @@ def get_cert(cn=None):
...
+    key (bytes): cert.key file contents
+    ca (Optional[bytes]): None or private ca contents
+
+    return True if ok
+    otherwise return False, set blocked status, and set useful error message
+    """
+
+    context = None
+    if ca:
+        context = ValidationContext(trust_roots=[ca])
```
nobuto-m, line 85: `ca` may be a bundle of CAs. So this example would help:
https://github.com/wbond/certvalidator/blob/master/docs/usage.md#custom-trust-rootsca-certs

Also, there may be a corner case where `ssl_ca` is used for unrelated services but `ssl_cert` can be validated via the system CA chain. So I would prefer using `extra_trust_roots` instead of `trust_roots`, which excludes the system's chain.

charmhelpers/contrib/hahelpers/apache.py
nobuto-m: For example, charm-keystone has `assess_status()` here at the end of a hook execution:
https://github.com/openstack/charm-keystone/blob/9b2f7e545dd682e90d5403a8cb8145309c42d5cf/hooks/keystone_hooks.py#L933
and eventually calls:
https://github.com/juju/charm-helpers/blob/8b22c53d86cb1ba745106e12ac713cff6810dd1e/charmhelpers/contrib/openstack/utils.py#L1003-L1010

nobuto-m (approved):

ajkavanagh:
@swalladge unfortunately, this PR now has conflicts with the master branch; please could you take a look, and I'll see what I can do about getting it merged.  Thanks.

ajkavanagh:
Does this also resolvee related issue: [LP#1772674](https://bugs.launchpad.net/charm-openstack-dashboard/+bug/1772674)?  Also, let's raise a review on a charm on gerrit that uses this code (either make sync on a classic charm, or a rebuild on a reactive one).  Also, if possible a linked zaza-openstack-tests new test would be great.

nobuto-m:
The question here is how we can install and import an external library.

> ModuleNotFoundError: No module named 'asn1crypto'

For other cases (e.g. netaddr), the charm-helper is installing APT packages inline.
https://github.com/juju/charm-helpers/blob/b2c089c39a0274e9d1326f776017f01af6a3d968/charmhelpers/contrib/network/ip.py#L48

But it's not the case here. Is there any way to overcome this situation?

nobuto-m:
> I'm slightly against merging this as it depends on two new libs that are in the category of "almost-but-not-quite-maintained". e.g. https://snyk.io/advisor/python/certvalidator

The page says "No known security issues" and it's just a thin wrapper for another library so "not actively maintained" status is totally expected because there is nothing to update really.

Instead, this is more appropriate metrics to be used:
https://snyk.io/advisor/python/asn1crypto
> security No known security issues
> popularity Influential project
> maintenance Sustainable
> community Sustainable

and this:
https://snyk.io/advisor/python/oscrypto

> It might be better to wait for the cryptography module to finish their work on certificate verification:[pyca/cryptography#8873](https://github.com/pyca/cryptography/pull/8873) - the issue here is that it would only be available in the most recent version, and charm-helpers has to go back to Python 3.8 for focal support.
> 
> Alternatively, the chain-of-fools answer on SO seems to provide some code that only relies on OpenSSL, of which the openssl binary is already used in charm-helpers: https://stackoverflow.com/a/69341797 - but this might require charms to have PyOpenSSL as a module.

I think we are in the loop. There is no suitable library in the Ubuntu archive as of right now. And using a bare openssl is tricky and requires the same amount of effort to rewrite the `asn1crypto` and `certvalidator` since properly handling certifications is hard. There was an effort using openssl and there were some amount of code:
https://github.com/juju/charm-helpers/pull/691/files
But it wasn't enough to handle certs charms are handling on a daily basis.

ajkavanagh:
> I think we are in the loop. There is no suitable library in the Ubuntu archive as of right now. And using a bare openssl is tricky and requires the same amount of effort to rewrite the asn1crypto and certvalidator since properly handling certifications is hard. There was an effort using openssl and there were some amount of code:
https://github.com/juju/charm-helpers/pull/691/files
But it wasn't enough to handle certs charms are handling on a daily basis.

It is an awkward situation.  The problem is two-fold;

1. Adding yet more dependencies to charm-helpers which then affects all charms that use it.  I suggested a workaround using local imports or the try-except method and then checking in the function.  It's not great, but it does prevent the lib from breaking charms.
2. classic charms (i.e. those that don't use a requirements.txt/pip/module installer) use distro packages for *all* of their modules, and thus, any dependencies have to be available as a package.  That effectively rules out `certvalidator` unless it can be packages for universe, but then getting it backported ...

So `asn1crypto` is probably okay.  However, there is a large amount of code in `certvalidator` and it isn't packaged.  We could vendor-in the module into charm-helpers, but that also has the problem of keeping it synced - maybe using a submodule; but that also feels 'wrong'.

Perhaps we should look at which classic charms need the functionality (any that use apache2 + TLS = all the OpenStack API classic charms at least), and perhaps have a vendoring-in at the `make sync` stage in the charm (note the `make sync` comes from charm-helpers at `tools/charm_helpers_sync/charm_helpers_sync.py`.

swalladge:
Conversation moved to https://github.com/juju/charm-helpers/pull/803

PR #728, opened 2022-07-22
Suppress YAMLLoadWarning

chanchiwai-ray:
This patch suppress the `YAMLLoadWarning`  when using `yaml.load()` without specifying `Loader`. By default, `yaml.load()` uses `FullLoader`.

### DIFF
--- a/charmhelpers/contrib/templating/contexts.py
+++ b/charmhelpers/contrib/templating/contexts.py
@@ -118,7 +118,8 @@ def juju_state_to_yaml(yaml_path, namespace_separator=':',
 
     if os.path.exists(yaml_path):
         with open(yaml_path, "r") as existing_vars_file:
-            existing_vars = yaml.load(existing_vars_file.read())
+            existing_vars = yaml.load(existing_vars_file.read(),
+                                      Loader=yaml.FullLoader)
     else:
         with open(yaml_path, "w+"):
             pass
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/templating/contexts.py
```diff
@@ -118,7 +118,8 @@ def juju_state_to_yaml(yaml_path, namespace_separator=':',
 
     if os.path.exists(yaml_path):
         with open(yaml_path, "r") as existing_vars_file:
-            existing_vars = yaml.load(existing_vars_file.read())
+            existing_vars = yaml.load(existing_vars_file.read(),
+                                      Loader=yaml.FullLoader)
```
ajkavanagh, line 122: We should probably use `yaml.safe_load()` as we don't actually need the full loader capabilities.

tools/charm_helpers_sync/charm_helpers_sync.py
```diff
@@ -34,7 +34,7 @@ def parse_config(conf_file):
     if not os.path.isfile(conf_file):
         logging.error('Invalid config file: %s.' % conf_file)
         return False
-    return yaml.load(open(conf_file).read())
+    return yaml.load(open(conf_file).read(), Loader=yaml.FullLoader)
```
ajkavanagh, line 37: Again, I suggest we use `safe_load()` as we don't actually need full loader.

ajkavanagh (changes requested):
Hi; thanks very much for the patch to resolve deprecation warnings.  I think this is an opportunity to tighten up the security and just use `safe_load` instead of the full loader as charms generally don't need to execute code when doing templating or reading config files.

tools/charm_helpers_sync/charm_helpers_sync.py
chanchiwai-ray: Noted, thanks for the suggestion.

ajkavanagh (approved):
LGTM; thanks very much for the changes.

PR #729, opened 2022-08-18
Don't run 'apt install' when list of packages is empty.

freyes:
A common pattern in charms is to call `apt_install()` passing a filtered
list of packages to avoid reinstallation of packages, so in normal
circumstances a charm may end up calling apt_install() with an empty
list producing the execution of "apt-get install" which will produce at
the very least a read of the list of packages database and build the
dependency tree, this is something that for IO constrained environments
impacts the performance of the running services. Also for environments
where the package database is broken (e.g. a post install script failed)
this would add more entropy and potentially making things more difficult
for the operator to fix.

$ sudo apt-get install
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
0 upgraded, 0 newly installed, 0 to remove and 63 not upgraded.

### DIFF
--- a/charmhelpers/contrib/hahelpers/cluster.py
+++ b/charmhelpers/contrib/hahelpers/cluster.py
@@ -324,7 +324,7 @@ def valid_hacluster_config():
     '''
     vip = config_get('vip')
     dns = config_get('dns-ha')
-    if not(bool(vip) ^ bool(dns)):
+    if not (bool(vip) ^ bool(dns)):
         msg = ('HA: Either vip or dns-ha must be set but not both in order to '
                'use high availability')
         status_set('blocked', msg)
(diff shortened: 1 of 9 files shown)


ajkavanagh:
I guess one thought I have around this is which charms might be relying on the 'odd' behaviour to work.  Thus, I'm in favour of landing on master, but would be very reticent to do a backport of this functionality.

ajkavanagh (approved):
LGTM; would like a second opinion regarding not backporting it as well.

freyes:
On Tue, 2022-08-23 at 02:20 -0700, Alex Kavanagh wrote:
> I guess one thought I have around this is which charms might be relying on the 'odd' behaviour to
> work. Thus, I'm in favour of landing on master, but would be very reticent to do a backport of
> this functionality.

I completely agree with you, I will just let this change organically get into the charms.

Thanks for the review, Alex ;-)

freyes:
@ajkavanagh , the PR got `closed`, hopefully just by mistake :-D

ajkavanagh:
> @ajkavanagh , the PR got `closed`, hopefully just by mistake :-D

Definitely "pushing the wrong button" problem here!  Re-opened and merged.  thanks for the patches!

PR #730, opened 2022-08-30
openstack: identity-service forwards compat

javacruft:
The new Keystone K8S operator exclusively uses the Application
data bag to provide endpoint information and credentials to
consuming services.

Check for the presence of data in the App data bag and use this
in preference to per-unit data as presented by the Keystone
machine charm.

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -61,6 +61,7 @@ from charmhelpers.core.hookenv import (
     network_get_primary_address,
     WARNING,
     service_name,
+    remote_service_name,
 )
 
 from charmhelpers.core.sysctl import create as sysctl_create
@@ -348,6 +349,14 @@ def db_ssl(rdata, ctxt, ssl_dir):
 
 class IdentityServiceContext(OSContextGenerator):
 
+    _forward_compat_remaps = {
+        'admin_user': 'admin-user-name',
+        'service_username': 'service-user-name',
+        'service_tenant': 'service-project-name',
+        'service_tenant_id': 'service-project-id',
+        'service_domain': 'service-domain-name',
+    }
+
     def __init__(self,
                  service=None,
                  service_user=None,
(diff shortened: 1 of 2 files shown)


javacruft:
WIP - please don't land yet!

tests/contrib/openstack/test_os_contexts.py
```diff
@@ -50,10 +50,16 @@ class FakeRelation(object):
         passwd = self.relation_get('password', rid='mysql:0', unit='mysql/0')
     '''
 
-    def __init__(self, relation_data):
+    def __init__(self, relation_data={}, app_data={}):
```
ajkavanagh, line 53: This is problematic as dictionaries have internal mutability, and are only initialised on the first run of the function; therefore, the function will memoize the `relation_data` and `app_data` between calls.  It's better to do:

```python
def __init__(self, relation_data=None, app_data=None):
    if relation_data is None:
        relation_data = {}
    if app_data is None:
        app_data = {}
    ...
```

charmhelpers/contrib/openstack/context.py
```diff
@@ -444,7 +464,27 @@ def __call__(self):
...
+                    # in key names - remap for compat with
+                    # existing relation data keys
+                    for key, value in _adata.items():
+                        if key == 'api-version':
+                            rdata[key.replace('-', '_')] = value.strip('v')
+                        else:
+                            rdata[key.replace('-', '_')] = value
+                    # Re-map some keys for backwards compatibility
+                    for target, source in self._forward_compat_remaps.items():
+                        rdata[target] = _adata.get(source)
```
ajkavanagh, line 483: Is it worth wrapping this with a `try: except KeyError:` in case `_adata` doesn't contain a `source`?  i.e. is it always guaranteed, or do we want to raise the KeyError here if it doesn't?

charmhelpers/contrib/openstack/context.py
```diff
@@ -478,6 +518,17 @@ def __call__(self):
...
                         'service_domain_id': rdata.get('service_domain_id')})
 
+                # NOTE:
+                # keystone-k8s operator presents full URLS
+                # for all three endpoints - public and internal are
+                # externally addressable for machine based charm
+                if 'public_auth_url' in rdata:
+                    ctxt.update({
+                        'public_auth_url': rdata.get('public_auth_url'),
+                        'internal_auth_url': rdata.get('internal_auth_url'),
```
ajkavanagh, line 528: Is it worth checking for `internal_auth_url` separately and adding it; e.g. this would KeyError if not available?

ajkavanagh (commented):
Just a minor comment and a couple of questions regarding what should happen if parts of the data are missing.

javacruft:
Thanks for the review @ajkavanagh  - I've pushed a fresh set of commits to resolve two of the changes you request.

charmhelpers/contrib/openstack/context.py
ajkavanagh: There's me being over-cautious.  I forgot that `dictionary.get()` supplies a `None`; python can be so inconsistent, as `.get_attribute` and friends all blow up!

ajkavanagh (commented):
LGTM

ajkavanagh (approved):

Issue #743, opened 2022-10-26
Unexpected escape of `fetch.ubuntu._run_with_retries` without successfully finishing the command


addyess:
[`fetch.ubuntu._run_with_retries`](https://github.com/juju/charm-helpers/blob/f086257853889220a042aa451e528eb18e10cf7a/charmhelpers/fetch/ubuntu.py#L907) seems to escape raising a failure in the event that the `subprocess.check_call` yields an exit code which is a failed exit code, but not in the list of `retry_exitcode`

this gives this method a false success when in reality it should fail immediately (no retrying)

i think the library method should raise when the exit value of `subprocess.check_call` is an error not on the rety list

TLDR:
if its exit value is 0 -- :tada:
if the exit value is in the `retry_results` -- retry max_retries like expected
if the exit value is not in the `retry_results` -- :fire:

PR #744, opened 2022-10-26
Issue#743: fetch.ubuntu._run_with_retries raises on unexpected exitcode

addyess:
this method gives a false success when in reality it should fail immediately (no retrying)

the method should raise when the exit value of `subprocess.check_call` is an error not on the rety list

TLDR:
if its exit value is 0 -- :tada:
if the exit value is in the `retry_results` -- retry max_retries like expected
if the exit value is not in the `retry_results` -- :fire:

Addresses https://github.com/juju/charm-helpers/issues/743

### DIFF
--- a/tests/contrib/openstack/test_deferred_events.py
+++ b/tests/contrib/openstack/test_deferred_events.py
@@ -304,10 +304,13 @@ class DeferredCharmServiceEventsTestCase(tests.utils.BaseTestCase):
     def test_check_restart_timestamps(self, get_service_start_time, log,
                                       clear_deferred_restarts,
                                       get_deferred_restarts):
+        request_time = '2021-02-02 10:19:55'
+        timestamp = datetime.datetime.strptime(
+            'Tue ' + request_time + ' UTC',
+            '%a %Y-%m-%d %H:%M:%S %Z').timestamp()
         deferred_restarts = [
-            # 'Tue 2021-02-02 10:19:55 UTC'
             deferred_events.ServiceEvent(
-                timestamp=1612261195.0,
+                timestamp=timestamp,
                 service='svcA',
                 reason='ReasonA',
                 action='restart')]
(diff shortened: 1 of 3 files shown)


kwmonroe (dismissed):
LGTM

addyess:
@kwmonroe do you like green tests?

Separate PR or can they be merged together

tests/contrib/openstack/test_deferred_events.py
```diff
@@ -304,10 +304,13 @@ def test_get_service_start_time(self, check_output):
     def test_check_restart_timestamps(self, get_service_start_time, log,
                                       clear_deferred_restarts,
                                       get_deferred_restarts):
+        request_time = '2021-02-02 10:19:55'
+        timestamp = datetime.datetime.strptime(
+            'Tue ' + request_time + ' UTC',
+            '%a %Y-%m-%d %H:%M:%S %Z').timestamp()
```
addyess, line 310: the timestamp is seconds since epoch -- but it's timezone specific.  This change was so this test would run successfully on all machines regardless of the locale's timezone. (ie -- it failed on my machine but was okay in other places where there was no TZ set)

tests/core/test_services.py
```diff
@@ -46,47 +46,56 @@ def test_register_preserves_order(self):
     @mock.patch.object(services.ServiceManager, 'reconfigure_services')
     @mock.patch.object(services.ServiceManager, 'stop_services')
     @mock.patch.object(hookenv, 'hook_name')
-    @mock.patch.object(hookenv, 'config')
-    def test_manage_stop(self, config, hook_name, stop_services, reconfigure_services):
+    @mock.patch.object(hookenv, '_run_atstart')
+    @mock.patch.object(hookenv, 'config', mock.Mock())
+    def test_manage_stop(self, _run_atstart, hook_name, stop_services, reconfigure_services):
```
addyess, line 51: adjust most of the `test_manage_*` tests such that they don't ACTUALLY run `_run_atstart` -- but use a mock instead

tests/core/test_services.py
```diff
@@ -46,47 +46,56 @@ def test_register_preserves_order(self):
...
+    def test_manage_other(self, _run_atstart, hook_name, stop_services, reconfigure_services, provide_data):
         manager = services.ServiceManager()
         hook_name.return_value = 'config-changed'
         manager.manage()
+        _run_atstart.assert_called_once_with()
         assert not stop_services.called
         reconfigure_services.assert_called_once_with()
         provide_data.assert_called_once_with()
 
+    @mock.patch.object(hookenv, '_atstart', [])
```
addyess, line 74: except for the `test_manage_calls_atstart` which guarantees `hookenv._atstart` is empty before the test runs

kwmonroe (approved):
LGTM

PR #745, opened 2022-10-27
use hvac unwrap api in 1.0.0

addyess:
Use the unwrap api when `hvac==1.0.0`

See https://hvac.readthedocs.io/en/stable/usage/auth_methods/token.html
> Wrapping/unwrapping a token:


kwmonroe (approved):
:cake:

PR #746, opened 2022-11-03
IndentityServiceContext app data fixes

ajkavanagh:
The IdentityServiceContext was recently updated to add the application data bag.  Unfortunately, they keystone charm uses *both* the regular relation data primarily, but the updates for endpoints are on both the application database and the regular relation data.  The IdentityServiceContext used the app data bag exclusively if there were any keys available, which leads to it ignoring the existing relation data entirely; this results in 'identity relation not ready' with missing data items.

This patch changes the functionality of the IdentityServiceContext so that it preferentially selects keys from the application data bag *unless* they are None, in which case it tries for those keys on the existing relation data.  e.g. it will stitch the two relations together with the app data bag taking priority.  This allows the (e.g.) nova-cloud-controller charm to correctly access the identity relation data and form a complete context.

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -485,10 +485,6 @@ class IdentityServiceContext(OSContextGenerator):
                     # Re-map some keys for backwards compatibility
                     for target, source in self._forward_compat_remaps.items():
                         adata[target] = _adata.get(source)
-                # else:
-                    # # No app data bag presented - fallback
-                    # # to legacy unit based relation data
-                    # rdata = relation_get(rid=rid, unit=unit)
                 # Now preferentially get data from the app data bag, but if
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/openstack/context.py
```diff
@@ -469,66 +469,82 @@ def __call__(self):
...
-                else:
-                    # No app data bag presented - fallback
-                    # to legacy unit based relation data
-                    rdata = relation_get(rid=rid, unit=unit)
-                serv_host = rdata.get('service_host')
+                        adata[target] = _adata.get(source)
+                # else:
+                    # # No app data bag presented - fallback
+                    # # to legacy unit based relation data
+                    # rdata = relation_get(rid=rid, unit=unit)
```
javacruft, line 491: L488-491 commented code?

javacruft (approved):

PR #755, opened 2023-02-28
Add RemoteRestartContext

gnuoy:
Add RemoteRestartContext to process remote restart triggers down a relation.


ajkavanagh:
Hi @gnuoy ; code looks good.  Is there a gerrit review or similar that it is being exercised against? Thanks!

PR #756, opened 2023-03-01
Remove test requirement python-distutils-extra

gnuoy:
Remove test requirement python-distutils-extra as it no longer seems to be needed and is breaking python3.5 tests.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,6 +1,5 @@
 # Test-only dependencies are unpinned.
 #
-git+https://git.launchpad.net/ubuntu/+source/python-distutils-extra
 pip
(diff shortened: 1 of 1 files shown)


gnuoy:
Example of failure https://github.com/juju/charm-helpers/actions/runs/4292353217/jobs/7478677223

wolsen:
Ah yes, this doesn't seem to be necessary anymore. Its not exactly clear to me *why* it was needed to begin with (it was introduced in a commit from 2019 that introduced tox - commit aa785c40). However, it is no longer necessary.

wolsen (approved):
LGTM

PR #759, opened 2023-03-13
Conditionally change the local DB file name

lmlg:
This PR tests whether the ops library is present in the system to decide the final file name for the local key/value database in order to avoid conflicts between that library and charmhelpers.

### DIFF
--- a/charmhelpers/core/unitdata.py
+++ b/charmhelpers/core/unitdata.py
@@ -173,12 +173,24 @@ class Storage(object):
     """
     def __init__(self, path=None):
         self.db_path = path
+        # The following is a hack to work around a bug in which both the
+        # ops framework libraries and charmhelpers end up using the same
+        # DB path, which leads to conflicts.
+        # See: https://bugs.launchpad.net/charm-ceph-mon/+bug/2005137
+        db_suffix = '.unit-state.db'
+        try:
+            import ops
+            db_suffix = '.unit-state2.db'
+            ops = None   # Don't hold an unneeded reference.
+        except:
+            pass
+
         if path is None:
             if 'UNIT_STATE_DB' in os.environ:
                 self.db_path = os.environ['UNIT_STATE_DB']
             else:
                 self.db_path = os.path.join(
-                    os.environ.get('CHARM_DIR', ''), '.unit-state.db')
+                    os.environ.get('CHARM_DIR', ''), db_suffix)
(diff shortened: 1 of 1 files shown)


charmhelpers/core/unitdata.py
```diff
@@ -173,13 +173,25 @@ class Storage(object):
     """
     def __init__(self, path=None, keep_revisions=False):
         self.db_path = path
+        # The following is a hack to work around a bug in which both the
+        # ops framework libraries and charmhelpers end up using the same
+        # DB path, which leads to conflicts.
+        # See: https://bugs.launchpad.net/charm-ceph-mon/+bug/2005137
+        db_suffix = '.unit-state.db'
```
ajkavanagh, line 180: It's not a 'suffix'; it's a hidden file.  e.g. `db_suffix` isn't quite right; `db_file` would be more accurate.

charmhelpers/core/unitdata.py
```diff
@@ -173,13 +173,25 @@ class Storage(object):
...
+        try:
+            import ops
+            db_suffix = '.unit-state2.db'
+            del ops   # Don't hold an unneeded reference.
+        except:
+            pass
```
ajkavanagh, line 186: This makes charm-helpers dependent on ops, which is entirely the wrong way around.

ajkavanagh (changes requested):
So I'm generally against this change due to opportunities for regressions and 'papering' over the bug.  Essentially charm-helpers 'owns' the file `.unit-state.db` as it predates the ops framework, and thus there are lots of charms *already* using that file.  If the ops framework needs an SQLite file, then it should have its own filename and not clobber the charm-helper's one (used in classic and reactive charms).  This is a hack, and it shouldn't be added to the library.

However, the Storage class initialise function *has* a path variable to allow multiple `Storage` objects to co-exist, so the ops framework could use.  Or it could set the `UNIT_STATE_DB` file.  Or the charm application code could do either of these actions, as the charm application code is bringing in charm-helpers.  If the ops framework is bring in charm-helpers then it shouldn't clobber the file itself.  I hope that makes sense.

charmhelpers/core/unitdata.py
lmlg: I disagree that it creates a dependency. It merely checks that a library can be imported; charm-helpers doesn't require `ops` to be installed for it to work.

PR #760, opened 2023-03-16
Ensure deferred events limited to current charm

coreycb:
The deferred_events() function was returning all deferred events on the machine, regardless of whether they were deferred for the current charm or not. This changes limits the returned events to those that correspond to the current charm.

Closes-Bug: #2011927


wolsen (changes requested):
The code change looks fine to me, though it could do with a unit test to cover this case.

cc @gnuoy for a review as well

gnuoy (approved):

gnuoy (approved):

PR #765, opened 2023-03-22
Refactor mysql password helper for setting passwords

ajkavanagh:
This modifies the existing method `set_mysql_password` and splits it into the original and a 2nd part called
`set_mysql_password_using_current_connection`.  This allows a root connection to be set up and then the password for any user to to be updated, so that a mysql instance charm can update the password.

It also splits the latter function into 3 parts:

 * `_update_password`
 * `_check_user_can_connect` - only used if hosts are not supplied
 * `_update_leader_settings` - the copy of the password.

The function is updated to allow 'hosts' to be supplied so that the password can be updated for users from different hosts.  This is mostly for MySQL8 which uses the 'user'@'host' form to further lockdown where authorisations can be performed from.

### DIFF
--- a/charmhelpers/contrib/database/mysql.py
+++ b/charmhelpers/contrib/database/mysql.py
@@ -342,9 +342,9 @@ class MySQLHelper(object):
         # changes to root-password were not supported) the user changed the
         # password, so leader-get is more reliable source than
         # config.previous('root-password').
-        rel_username = None if username == 'root' else username
         if not current_password:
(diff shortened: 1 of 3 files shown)


freyes (approved):
LGTM, thanks for the work on this patch and refactoring the method to avoid having one that's hard to understand.

freyes (approved):

gnuoy (approved):

PR #766, opened 2023-03-22
Handle stopstart in get_deferred_restarts()

coreycb:
This handles situations like the following when stopstart=True has been used for restart_on_change():

$ cat /var/lib/policy-rc.d/charm-ovn-chassis-<uuid>.deferred action: stop
policy_requestor_name: ovn-chassis
policy_requestor_type: charm
reason: Package update
service: ovs-vswitchd.service
timestamp: 1679507800

Closes-Bug: #2012553


coreycb:
Looking at the unit test it seems that action='stop' is not expected to be returned from get_deferred_restarts() so I may be going down the wrong path here. @gnuoy  , thoughts?

gnuoy (changes requested):

PR #767, opened 2023-03-29
Restart deferred stopped events in restart_services_action()

coreycb:
This change updates restart_services_action() to restart deferred 'stop' events, in addition to the already existing restart of 'restart' events.

Some services get stopped, and not restarted, when dpkg-reconfigure is called. For example: ovsdb-server, ovs-vswitchd, and ovs-record-hostname from the openvswitch package in kinetic+ are defined to behave this way. This results in a deferred event such as:

$ cat /var/lib/policy-rc.d/charm-ovn-chassis-<uuid>.deferred action: stop
policy_requestor_name: ovn-chassis
policy_requestor_type: charm
reason: Package update
service: ovs-vswitchd.service
timestamp: 1679507800

Closes-Bug: #2012553

### DIFF
--- a/charmhelpers/contrib/openstack/deferred_events.py
+++ b/charmhelpers/contrib/openstack/deferred_events.py
@@ -245,6 +245,15 @@ def get_deferred_restarts():
     return [e for e in get_deferred_events() if e.action == 'restart']
 
 
+def get_deferred_stops():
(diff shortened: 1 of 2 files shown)


coreycb:
based on discussion, this will make more sense as code that is specific to the ovn-chassis charm

PR #769, opened 2023-04-19

gnuoy:

https://bugs.launchpad.net/charm-keystone/+bug/2015103

### DIFF
--- a/tests/contrib/hahelpers/test_cluster_utils.py
+++ b/tests/contrib/hahelpers/test_cluster_utils.py
@@ -231,8 +231,11 @@ class ClusterUtilsTests(TestCase):
         ]
         self.assertTrue(cluster_utils.https())
 
-    def test_https_cert_key_in_identity_relation(self):
+    @patch('charmhelpers.contrib.openstack.cert_utils')
+    def test_https_cert_key_in_identity_relation(self, cert_utils):
         '''It determines https is available if cert in identity-service'''
+        cert_utils.get_certificate_request.return_value = False
+        cert_utils.get_requests_for_local_unit.return_value = {}
         self.config_get.return_value = False
         self.relation_ids.return_value = 'identity-service:0'
         self.relation_list.return_value = 'keystone/0'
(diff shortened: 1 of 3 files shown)


coreycb (commented):
LGTM. Could you add the bug reference to the commit message? 2015103

coreycb (dismissed):

wolsen (dismissed):

ajkavanagh (changes requested):
Unfortunately, it looks like a few unit tests are breaking due to the change; it looks like the tests need some additional mocks as the method-under-test eventually calls out to juju primatives.

e.g.

```
                                                                                                                                                                                                                              
======================================================================                                                                                                                                                        
ERROR: tests.contrib.hahelpers.test_cluster_utils.ClusterUtilsTests.test_https_cert_key_incomplete_identity_relation                                                                                                          
----------------------------------------------------------------------                                                                                                                                                        
testtools.testresult.real._StringException: Traceback (most recent call last):                                                                                                                                                
  File "/home/ubuntu/git/github.com/juju/charm-helpers/charmhelpers/core/hookenv.py", line 1180, in inner_translate_exc2                                                                                                      
    return f(*args, **kwargs)                                                                                                                                                                                                 
  File "/home/ubuntu/git/github.com/juju/charm-helpers/charmhelpers/core/hookenv.py", line 1374, in network_get_primary_address                                                                                               
    response = subprocess.check_output(                                                                                                                                                                                       
  File "/usr/lib/python3.8/subprocess.py", line 415, in check_output                                                                                                                                                          
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,                                                                                                                                                          
  File "/usr/lib/python3.8/subprocess.py", line 493, in run                                                                                                                                                                   
    with Popen(*popenargs, **kwargs) as process:                                                                                                                                                                              
  File "/usr/lib/python3.8/subprocess.py", line 858, in __init__                                                                                                                                                              
    self._execute_child(args, executable, preexec_fn, close_fds,                                                                                                                                                              
  File "/usr/lib/python3.8/subprocess.py", line 1704, in _execute_child                                                                                                                                                       
    raise child_exception_type(errno_num, err_msg, err_filename)                                                                                                                                                              
FileNotFoundError: [Errno 2] No such file or directory: 'network-get' 
```

Unfortunately, the github actions are currently broken due to: https://github.com/juju/charm-helpers/issues/774 so you'll need to run them manually (on say focal), until we get this fixed.

tests/contrib/hahelpers/test_cluster_utils.py
```diff
@@ -244,8 +247,27 @@ def test_https_cert_key_in_identity_relation(self):
...
+        self.config_get.return_value = False
+        self.relation_ids.return_value = 'identity-service:0'
+        self.relation_list.return_value = 'keystone/0'
+        self.relation_get.side_effect = [
+            'yes',  # relation_get('https_keystone')
+            'cert',  # relation_get('ssl_cert')
+            'key',  # relation_get('ssl_key')
+            'ca_cert',  # relation_get('ca_cert')
+        ]
+        self.assertTrue(cluster_utils.https())
```
ajkavanagh, line 264: I think this should be `assertFalse` as the request for local unit is `{}` and that it will return `False` from `https()`.

ajkavanagh (changes requested):
Please see inline test comment; I think a boolean assert should be flipped.

ajkavanagh:
I've manually verified that the tox pep8 and py3 targets run on focal and jammy (actually py310 due to nose2 requirement).  This is in lieu of the fix to the CI runners.

PR #771, opened 2023-04-19
Add -S option when calling add-apt-repository for a line starting with deb

### LINKED ISSUE #770: fetch.add_source fails for private ppa on jammy
This is related to [lp:2017014](https://bugs.launchpad.net/ubuntu/+source/software-properties/+bug/2017014) (and maybe https://bugs.launchpad.net/ubuntu/+source/software-properties/+bug/1991553)

When adding a private ppa via install_sources from https://github.com/stub42/layer-apt, which ends up calling charmhelpers.fetch.add_source it fails with:
```
  File "/usr/bin/add-apt-repository", line 364, in <module>
    sys.exit(0 if addaptrepo.main() else 1)
  File "/usr/bin/add-apt-repository", line 357, in main
    shortcut.add()
  File "/usr/lib/python3/dist-packages/softwareproperties/shortcuthandler.py", line 222, in add
    self.add_key()
  File "/usr/lib/python3/dist-packages/softwareproperties/shortcuthandler.py", line 398, in add_key
    if not all((self.trustedparts_file, self.trustedparts_content)):
  File "/usr/lib/python3/dist-packages/softwareproperties/ppa.py", line 141, in trustedparts_content
    key = self.lpppa.getSigningKeyData()
  File "/usr/lib/python3/dist-packages/lazr/restfulclient/resource.py", line 592, in __call__
    response, content = self.root._browser._request(
  File "/usr/lib/python3/dist-packages/lazr/restfulclient/_browser.py", line 429, in _request
    raise error
lazr.restfulclient.errors.Unauthorized: HTTP Error 401: Unauthorized
Response headers:
---
-content-encoding: gzip
content-length: 78
content-security-policy: frame-ancestors 'self';
content-type: text/plain;charset=utf-8
date: Wed, 19 Apr 2023 13:36:06 GMT
server: gunicorn
status: 401
strict-transport-security: max-age=15552000
vary: Accept,Accept-Encoding
x-content-type-options: nosniff
x-frame-options: SAMEORIGIN
x-launchpad-revision: 473868c1cc6b58a9bc722c23840374c93a7a274b
x-lazr-notifications: []
x-powered-by: Zope (www.zope.org), Python (www.python.org)
x-request-id: 7744610e-481a-4489-ad5b-e9288c8b7588
x-vcs-revision: 473868c1cc6b58a9bc722c23840374c93a7a274b
x-xss-protection: 1; mode=block
---
```
This is because of changes to add-apt-repository, which now treats evething as a PPA and tries to get information from launchpad via API calls. if the ppa is private, such attemps will fail and here we are now :-)

verterok:
This is only for jammy or newer, and it just bypass any custom ppa logic in add-apt-repository by passing `-S` if the source line starts with deb.
fixes #770

### DIFF
--- a/charmhelpers/fetch/ubuntu.py
+++ b/charmhelpers/fetch/ubuntu.py
@@ -769,13 +769,13 @@ def _add_apt_repository(spec):
     if '{series}' in spec:
         spec = spec.replace('{series}', series)
     cmd = ['add-apt-repository', '--yes']
-    if (spec.strip().startswith('deb')
-            and CompareHostReleases(series) >= 'jammy'):
+    if (spec.strip().startswith('deb') and
+            CompareHostReleases(series) >= 'jammy'):
         # if starts with deb, just add the source.list as per lp:2017014, only
         # in jammy or newer
         cmd.append('-S')
     cmd.append(spec)
-    _run_with_retries(['add-apt-repository', '--yes', spec],
+    _run_with_retries(cmd,
                       cmd_env=env_proxy_settings(['https', 'http', 'no_proxy'])
                       )
 


verterok:
Take this as an idea on a possible fix, I'm far from an expert on charmhelper internals and this also requires some tests.
We are using this in our code to work around the issue.

charmhelpers/fetch/ubuntu.py
```diff
@@ -760,9 +765,16 @@ def _add_apt_repository(spec):
...
+    cmd = ['add-apt-repository', '--yes']
+    if (spec.strip().startswith('deb')
+            and CompareHostReleases(series) >= 'jammy'):
+        # if starts with deb, just add the source.list as per lp:2017014, only
+        # in jammy or newer
+        cmd.append('-S')
+    cmd.append(spec)
     _run_with_retries(['add-apt-repository', '--yes', spec],
                       cmd_env=env_proxy_settings(['https', 'http', 'no_proxy'])
                       )
```
ajkavanagh, line 780: This code doesn't actually use `cmd` in `_run_with_retries(...)`?  I'm guessing this isn't the actual code that is being used?

If 'deb' needs specially handling at jammy+ then the `_mapping` further up the code could probably pick it out separately from `http:`, `https:`, and `ppa:` to make it obvious that just `deb` is being handled specifically.

ajkavanagh (changes requested):
Thanks very much for the code suggestion.  I don't then the suggested patch actually works, and it definitely needs unit tests to verify that it does what is expected, please.

charmhelpers/fetch/ubuntu.py
verterok: indeed, pushed things in a rush to provide something, will fix and push it
AFAIK, only deb for private ppa but haven't tested other cases

PR #773, opened 2023-04-26
Don't filter the proxy settings being passed to curl

vorlonofportland:
env_proxy_settings() DTRT by default with no arguments.  Passing 'https' as an argument just causes it to FAIL to honor no_proxy in the environment which it should.


ajkavanagh:
@vorlonofportland please could you add a reason for why this PR is needed.  e.g. is there a bug that it is fixing?  Thanks.

PR #775, opened 2023-05-06
Update method for checking endpoint protocol [antelope]

freyes:


ajkavanagh:
Needs another review for a backport.  Perhaps @javacruft or @coreycb please?

coreycb (approved):

PR #776, opened 2023-05-11
Fix dpkg-query usage

swalladge:
The previous use of dpkg-query was dependent on system locale, and would break if the locale wasn't a variant of English. For example, if the locale was fr_FR.UTF-8,
the heading would return 'nom' instead of 'name',
and the method would return an empty result.

This fix ensures that the output is consistent across locales. The new output is also easier to parse in the code, thus simplifying the method.

Closes-Bug: https://bugs.launchpad.net/charm-helpers/+bug/2019161


ajkavanagh:
Hi @swalladge please could you link this to a gerrit review that tests this change please.

ajkavanagh (approved):
LGTM

PR #777, opened 2023-05-11
Update Makefile to be consistent with CI (and fix CI)

swalladge:
Also remove python3.5 from the CI build,
because that is no longer supported in charm-helpers.


ajkavanagh (approved):
LGTM

PR #778, opened 2023-05-26
[stable/zed] Update method for checking endpoint protocol (#769) (#775)

coreycb:

Co-authored-by: Liam Young <liam.young@canonical.com>
(cherry picked from commit ed01437357921a95f2dedcff1ec50ae68ef46a53)

### DIFF
--- a/charmhelpers/contrib/hahelpers/cluster.py
+++ b/charmhelpers/contrib/hahelpers/cluster.py
@@ -221,6 +221,13 @@ def https():
         return True
     if config_get('ssl_cert') and config_get('ssl_key'):
         return True
+    # Local import to avoid ciruclar dependency.
+    import charmhelpers.contrib.openstack.cert_utils as cert_utils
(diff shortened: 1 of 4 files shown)


freyes (commented):
This backport LGTM, although I'm not sure why CI jobs were run here.

freyes (approved):

ajkavanagh:
It would be good to link to some reviews that use this functionality on stable/zed charms, please.

freyes (changes requested):
We should include in this same PR this other patch - https://github.com/juju/charm-helpers/commit/58097330bd15cfd7b539481c55718c6cee8ca124 .

Also this change should be rebased, so the CI can run on this patchset.

freyes (commented):
LGTM. We just need to review/merge the missing fix in stable/antelope first - https://github.com/juju/charm-helpers/pull/799

PR #779, opened 2023-05-26
[stable/yoga] Update method for checking endpoint protocol (#769) (#775)

coreycb:

Co-authored-by: Liam Young <liam.young@canonical.com>
(cherry picked from commit ed01437357921a95f2dedcff1ec50ae68ef46a53)


ajkavanagh (approved):
LGTM; tested by gerrit review: https://review.opendev.org/c/openstack/charm-keystone/+/884686

One more vote for a stable change to land.

PR #798, opened 2023-06-06
Support legacy cert requests

dosaboy:
openstack.cert_utils needs to support new and legacy tls cert requests.

Closes-Bug: #2022980

### DIFF
--- a/charmhelpers/contrib/openstack/cert_utils.py
+++ b/charmhelpers/contrib/openstack/cert_utils.py
@@ -409,6 +409,9 @@ def get_requests_for_local_unit(relation_name=None):
     relation_name = relation_name or 'certificates'
     bundles = []
     for rid in relation_ids(relation_name):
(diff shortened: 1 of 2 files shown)


freyes (changes requested):
I tested this change with ironic-api and it did the trick.

```
$ sudo grep Prox /etc/apache2/sites-enabled/openstack_https_frontend.conf
    ProxyPass / http://localhost:6365/
    ProxyPassReverse / http://localhost:6365/
```

@dosaboy , can you update the commit message to identify where the issue was introduced, why this is needed, etc, also can you add a unit test to make sure we don't introduce regressions down later.

dosaboy:
commit updated and tests added

freyes (approved):
LGTM, thanks Ed

freyes (approved):

javacruft (approved):

PR #799, opened 2023-06-07
Support legacy cert requests [antelope]

freyes:
The tls-certificates interface supports two ways of requesting certs and openstack.cert_utils needs to support both when checking if certificates are ready to be used. Without this it will never declare certificates as ready if they were requested with the old/legacy method.

Closes-Bug: #2022980
(cherry picked from commit 58097330bd15cfd7b539481c55718c6cee8ca124)


coreycb (approved):

dosaboy:
lgtm

PR #768, opened 2023-04-14
Fix openstack release comparison wrap

coreycb:
Fix OpenStack release comparison wrap in os_requires_version() to ensure, for example, that 'antelope' is > 'zed'.

PR #800, opened 2023-06-09
Fix openstack release comparison wrap (#768)

freyes:
Fix OpenStack release comparison wrap in os_requires_version() to ensure, for example, that 'antelope' is > 'zed'.

(cherry picked from commit 6ae6d7adde8ce500d77bac6f08edff4a676dcc03)


dosaboy:
could you perhaps add a unit test to prove the A > Z check is now fixed?

PR #801, opened 2023-06-12
Allow users to disable wsgi socket rotation

zhhuabj:
The lp bug 1863232 introduced a new configuration option called WSGISocketRotation which allows users to disable wsgi socket rotation on the apache side. This patch will also allow setting this option on the charm side.

Partial-Bug: 2021550

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -1748,6 +1748,9 @@ class WSGIWorkerConfigContext(WorkerConfigContext):
 
     def __call__(self):
         total_processes = _calculate_workers()
+        enable_wsgi_rotation = config('wsgi-rotation')
+        if enable_wsgi_rotation is None:
+            enable_wsgi_rotation = True
         ctxt = {
             "service_name": self.service_name,
             "user": self.user,
@@ -1761,6 +1764,7 @@ class WSGIWorkerConfigContext(WorkerConfigContext):
             "public_processes": int(math.ceil(self.public_process_weight *
                                               total_processes)),
             "threads": 1,
+            "wsgi_rotation": enable_wsgi_rotation,
         }
         return ctxt
 
--- a/tests/contrib/openstack/test_os_contexts.py
+++ b/tests/contrib/openstack/test_os_contexts.py
@@ -3569,7 +3569,9 @@ class ContextTests(unittest.TestCase):
     @patch.object(context, '_calculate_workers')
     def test_wsgi_worker_config_context(self,
                                         _calculate_workers):
-        self.config.return_value = 2  # worker-multiplier=2
+        self.config.side_effect = fake_config({
+            'worker-multiplier': 2, 'non-defined-wsgi-rotation': True
+        })
         _calculate_workers.return_value = 8
         service_name = 'service-name'
         script = '/usr/bin/script'
@@ -3586,13 +3588,16 @@ class ContextTests(unittest.TestCase):
             "admin_processes": 2,
             "public_processes": 6,
             "threads": 1,
+            "wsgi_rotation": True,
         }
         self.assertEqual(expect, ctxt())
 
     @patch.object(context, '_calculate_workers')
     def test_wsgi_worker_config_context_user_and_group(self,
                                                        _calculate_workers):
-        self.config.return_value = 1
+        self.config.side_effect = fake_config({
+            'worker-multiplier': 1, 'wsgi-rotation': False
+        })
         _calculate_workers.return_value = 1
         service_name = 'service-name'
         script = '/usr/bin/script'
@@ -3613,6 +3618,7 @@ class ContextTests(unittest.TestCase):
             "admin_processes": 1,
             "public_processes": 1,
             "threads": 1,
+            "wsgi_rotation": False,
         }
         self.assertEqual(expect, ctxt())
 
--- a/charmhelpers/contrib/openstack/templates/wsgi-openstack-api.conf
+++ b/charmhelpers/contrib/openstack/templates/wsgi-openstack-api.conf
@@ -12,6 +12,12 @@ Listen {{ admin_port }}
 Listen {{ public_port }}
 {% endif -%}
 
+{% if wsgi_rotation -%}
+WSGISocketRotation On
+{% else -%}
+WSGISocketRotation Off
+{% endif -%}
+
 {% if port -%}
 <VirtualHost *:{{ port }}>
     WSGIDaemonProcess {{ service_name }} processes={{ processes }} threads={{ threads }} user={{ user }} group={{ group }} \


charmhelpers/contrib/openstack/context.py
```diff
@@ -1761,6 +1761,7 @@ def __call__(self):
             "public_processes": int(math.ceil(self.public_process_weight *
                                               total_processes)),
             "threads": 1,
+            "disable_wsgi_rotation": config('disable-wsgi-rotation'),
```
freyes, line 1764: IIRC running `config('non-defined-config')` will fail, so introducing this change will break charms that don't have this config option defined.

freyes (dismissed):
I was reading the upstream doc[0], I think we should make this option the default to `WSGISocketRotation Off`, because it would allow a smoother restart for operators, although we should document (or point out  to the upstream doc) what are the cons of it for security minded operators.

[0] https://modwsgi.readthedocs.io/en/master/release-notes/version-4.6.0.html#new-features

charmhelpers/contrib/openstack/context.py
zhhuabj: has changed the option disable-wsgi-rotation to wsgi-rotation

zhhuabj:
> I was reading the upstream doc[0], I think we should make this option the default to `WSGISocketRotation Off`, because it would allow a smoother restart for operators, although we should document (or point out to the upstream doc) what are the cons of it for security minded operators.
> 
> [0] https://modwsgi.readthedocs.io/en/master/release-notes/version-4.6.0.html#new-features

Hi freyes, thanks for the review. Now I have changed to use 'WSGISocketRotation Off' as the default vaule, and also change the patch in ncc side as well - https://review.opendev.org/c/openstack/charm-nova-cloud-controller/+/885836

and this is my self-test result, it works - https://paste.ubuntu.com/p/qMfK9mTpt4/

charmhelpers/contrib/openstack/context.py
zhhuabj: and for non-defined-config, I also add the following 3 lines

+        enable_wsgi_rotation = config('wsgi-rotation')                         
+        if enable_wsgi_rotation is None:                                       
+            enable_wsgi_rotation = False

zhhuabj:
> > Hi freyes, thanks for the review. Now I have changed to use 'WSGISocketRotation Off' as the default vaule, [..]
> 
> do you have the intention to backport this patch to stable releases?, because changing the default behavior would make it a non-candidate for backporting.

Yeah, the customer is using ussuri, we need to backport this patch to stable release. Just like what I did in the initial patch, I wanted to use WSGISocketRotation=True as the default value, because this can be consistent with the default values of Apache and will not introduce any regressions. Do you want me to change the default value to true?

ajkavanagh (changes requested):
Please link the gerrit review into the description.  Also, if this changes the default behaviour for charm-helpers then it can't be backported as a feature.

zhhuabj:
> Please link the gerrit review into the description. Also, if this changes the default behaviour for charm-helpers then it can't be backported as a feature.

ok, let's me change back to the original default values 'WSGISocketRotation=On', thanks for the review

zhhuabj:
> > because changing the default behavior would make it a non-candidate for backporting.
> 
> > if this changes the default behaviour for charm-helpers then it can't be backported as a feature.
> 
> I know I'm bike-shedding, but if (and only if) using the new value is a fix for a particular issue and if beneficial for everyone with a minimal disruption, why don't we update the value without a config option since the charms can be opinionated about deployments?
> 
> Not sure if it's a good idea to have another config option with not using it out of the box for a simple task to achieve a graceful restart. What's the implication and potential regression of using the new value?

hi @nobuto-m, do you mean removing the config option wsgi-rotation and hardcoding WSGISocketRotation as Off directly?

ajkavanagh:
> > because changing the default behavior would make it a non-candidate for backporting.
> 
> > if this changes the default behaviour for charm-helpers then it can't be backported as a feature.
> 
> I know I'm bike-shedding, but if (and only if) using the new value is a fix for a particular issue and if beneficial for everyone with a minimal disruption, why don't we update the value without a config option since the charms can be opinionated about deployments?

Generally, my statement was that "in general" we don't backport features that change default behaviour. Therefore, this would fail that test (if the default from existing 'true' changed) and thus couldn't be backported *without an exception*.  Good exceptions are solving security exploits, etc, as one would expect.  Adding a feature generally isn't.

> 
> Not sure if it's a good idea to have another config option with not using it out of the box for a simple task to achieve a graceful restart. What's the implication and potential regression of using the new value?

So the way it's being implemented is, if the charm doesn't have the config option, then it gets the default behaviour from charm-helpers, and thus from mod_wsgi.  If the charm implements the config option it the charm's `config.yaml`, then it may opt into new behaviour.  As a default, without any changes to the charm code, the charm will get a rotating of the socket.

This is probably a good compromise.

Now, *should* the charm have the config option?  I don't know the answer to that.  If 503 responses are an issue then maybe it is required, but I would have thought that retries in client API code would get around that issue (as the 503's in this case are due to service restarts which could/should only really happen during deployment or other operator actions such as config-changed on the application.

Maybe we should collect more data before adding the config item to the charms' `config.yaml`.  However, I'm not against adding the ability to charm-helpers at this time.

arif-ali:
> Now, _should_ the charm have the config option? I don't know the answer to that. If 503 responses are an issue then maybe it is required, but I would have thought that retries in client API code would get around that issue (as the 503's in this case are due to service restarts which could/should only really happen during deployment or other operator actions such as config-changed on the application.
> 
> Maybe we should collect more data before adding the config item to the charms' `config.yaml`. However, I'm not against adding the ability to charm-helpers at this time.

So, the 503 issue happens after the `invoke-rc.d apache2 reload` via logrotate and at that particular time is when the socket does get rotated, but the process still sees the old socket. The issue remains at least for 24 hours, until the potential next reload.

In this particular scenario we can get various API requests failing, and hence can cause issues in large deployments via heat as an example

so, unfortunately it's not via config-changed, deployment or operator actions

ajkavanagh:
@arif-ali 

> So, the 503 issue happens after the `invoke-rc.d apache2 reload` via logrotate and at that particular time is when the socket does get rotated, but the process still sees the old socket. The issue remains at least for 24 hours, until the potential next reload.

So that seems very curious.  I'm not disputing it happening, but I am curious as to why it persists for 24 hours (keep-alive holding the socket open that long with no responses??) and even more curious as to why the *next* socket rotation fixes it.  Any ideas?

arif-ali:
> @arif-ali
> 
> > So, the 503 issue happens after the `invoke-rc.d apache2 reload` via logrotate and at that particular time is when the socket does get rotated, but the process still sees the old socket. The issue remains at least for 24 hours, until the potential next reload.
> 
> So that seems very curious. I'm not disputing it happening, but I am curious as to why it persists for 24 hours (keep-alive holding the socket open that long with no responses??) and even more curious as to why the _next_ socket rotation fixes it. Any ideas?

Looking at the details from the logs we have, the issue does disappear at midnight when the next reload is done. We have not been able to ascertain the exact the reason, as it is extremely difficult to reproduce

dosaboy:
@arif-ali im testing this out for myself on focal and i see the following behaviour:

```
# ss -nope | grep apache
u_str  ESTAB   0       768                                           * 6265434                                               * 6265435                           users:(("apache2",pid=1483521,fd=25)) <--                                      
u_str  ESTAB   5       0        /var/run/apache2/wsgi.1469761.1.1.sock 6265435                                               * 6265434                           users:(("apache2",pid=1483517,fd=10)) --> ino:539 dev:0/26                     
tcp    ESTAB   0       0                                    10.5.2.230:43766                                        10.5.1.148:35357                             users:(("apache2",pid=1483518,fd=26)) timer:(keepalive,54sec,0) uid:64060 ino:6265386 sk:225 <->
tcp    ESTAB   0       0                                    10.5.2.230:51916                                        10.5.1.148:5000                              users:(("apache2",pid=1483518,fd=27)) timer:(keepalive,55sec,0) uid:64060 ino:6265391 sk:226 <->
tcp    ESTAB   0       0                                     127.0.0.1:40004                                         127.0.0.1:3306                              users:(("apache2",pid=1483518,fd=22)) timer:(keepalive,119min,0) uid:64060 ino:6265370 sk:228 <->
tcp    ESTAB   0       0                                     127.0.0.1:40018                                         127.0.0.1:3306                              users:(("apache2",pid=1483518,fd=24)) timer:(keepalive,119min,0) uid:64060 ino:6265378 sk:229 <->
tcp    ESTAB   0       0                                     127.0.0.1:40002                                         127.0.0.1:3306                              users:(("apache2",pid=1483518,fd=21)) timer:(keepalive,119min,0) uid:64060 ino:6265366 sk:22c <->
tcp    ESTAB   0       0                                    10.5.2.230:34292                                        10.5.1.148:35357                             users:(("apache2",pid=1483518,fd=28)) timer:(keepalive,56sec,0) uid:64060 ino:6265393 sk:22e <->
tcp    ESTAB   0       0                                    10.5.2.230:59040                                        10.5.2.179:9696                              users:(("apache2",pid=1483518,fd=29)) timer:(keepalive,56sec,0) uid:64060 ino:6265396 sk:230 <->
tcp    ESTAB   0       0                                     127.0.0.1:42840                                         127.0.0.1:8754                              users:(("apache2",pid=1483522,fd=25)) uid:33 ino:6265284 sk:221 <->            
tcp    ESTAB   0       0                                     127.0.0.1:40006                                         127.0.0.1:3306                              users:(("apache2",pid=1483518,fd=23)) timer:(keepalive,119min,0) uid:64060 ino:6265374 sk:233 <->
tcp    ESTAB   0       0                           [::ffff:10.5.2.230]:8764                                [::ffff:10.5.2.230]:48554                             users:(("apache2",pid=1483522,fd=24)) timer:(keepalive,119min,0) uid:33 ino:6265282 sk:222 <->
tcp    ESTAB   0       0                                         [::1]:32834                                             [::1]:11211                             users:(("apache2",pid=1483518,fd=25)) uid:64060 ino:6265384 sk:235 <->         
tcp    ESTAB   0       0                            [::ffff:127.0.0.1]:8754                                 [::ffff:127.0.0.1]:42840                             users:(("apache2",pid=1483521,fd=24)) timer:(keepalive,119min,0) uid:33 ino:6265285 sk:223 <->
# invoke-rc.d apache2 reload
# ss -nope | grep apache
# 
```
i.e. the restart causes all connections, including keepalive ones, to be closed. Am i doing something differently to you?

dosaboy:
Having re-read https://modwsgi.readthedocs.io/en/master/release-notes/version-4.6.0.html#new-features and given that the problem has proven very hard to reproduce, I think that at the very least we should default to the default apache behaviour i.e. ```WSGISocketRotation On```. The docs clearly say "Rotation should only be disabled where the Apache configuration for the mod_wsgi application stays constant over time." and it was done this way for good reason since if the apache config changes e.g. to address a security concern and rotation is disabled, a request on a pre-restart connection can be serviced by a post-restart worker which could introduce security concerns. Having this support in charmhelpers to allow the config to be optionally overriden by individual charms at some later point seems fair so as-is I think the patch makes sense but I think that we need to find a stronger case to add a new config option to charms themselves and also need to test more before we consider disabling by default in any of the Openstack charms.

tests/contrib/openstack/test_os_contexts.py
```diff
@@ -3586,13 +3588,16 @@ def test_wsgi_worker_config_context(self,
...
         }
         self.assertEqual(expect, ctxt())
 
     @patch.object(context, '_calculate_workers')
     def test_wsgi_worker_config_context_user_and_group(self,
                                                        _calculate_workers):
-        self.config.return_value = 1
+        self.config.side_effect = fake_config({
+            'worker-multiplier': 1, 'wsgi-rotation': True
+        })
```
ajkavanagh, line 3600: We could do with the opposite test here; the default is `True`/`On`, so the handy test would be to set it to `False` to get an `Off` in the config.

ajkavanagh (changes requested):
Just one minor change please.

tests/contrib/openstack/test_os_contexts.py
zhhuabj: has changed to False, thanks a lot

ajkavanagh (approved):

sombrafam:
What criteria are we and using to determine which charms will be opinionated? There are others (ceilometer, manila, etc) not covered here.

zhhuabj:
> 

Hi @sombrafam , I'm just searching for the keyword 'wsgi-openstack-api.conf' or 'WSGIWorkerConfigContext' to decide which charms will be opinionated.

$ grep -r 'wsgi-openstack-api.conf' ceilometer/
ceilometer/lib/ceilometer_utils.py:WSGI_CEILOMETER_API_CONF = '/etc/apache2/sites-enabled/wsgi-openstack-api.conf'
$ grep -r 'wsgi-openstack-api.conf' manila/

$ grep -r 'WSGIWorkerConfigContext' ceilometer
ceilometer/lib/ceilometer_utils.py:                [context.WSGIWorkerConfigContext(name="ceilometer",
ceilometer/charmhelpers/contrib/openstack/context.py:class WSGIWorkerConfigContext(WorkerConfigContext):
$ grep -r 'WSGIWorkerConfigContext' manila/
<empty>

Oh, ceilometer charm also needs to be optionated. Just I added it as well - https://review.opendev.org/c/openstack/charm-ceilometer/+/887793

thanks

PR #802, opened 2023-06-26
[stable/ussuri] Add support for OVN UCA pocket

mkalcok:
Backport of #738 to allow `neutron-api-plugin-ovn` to use OVN UCA pockets in focal-ussuri. Rest of the backporting effort:

-  #804 
-  #805 
-  #806 

 Original PR message below:

The OVN UCA pocket will be used as an overlay repository in addition to any regular UCA pocket and the charms are expected to expose configuration of it as a separate source configuration option.

Add pocket names and allow for writing any OVN UCA configuration to a separate sources.list.d file.

### DIFF
--- a/charmhelpers/fetch/ubuntu.py
+++ b/charmhelpers/fetch/ubuntu.py
@@ -224,6 +224,10 @@ CLOUD_ARCHIVE_POCKETS = {
     'yoga/proposed': 'focal-proposed/yoga',
     'focal-yoga/proposed': 'focal-proposed/yoga',
     'focal-proposed/yoga': 'focal-proposed/yoga',
+
+    # OVN
+    'focal-ovn-22.03': 'focal-updates/ovn-22.03',
+    'focal-ovn-22.03/proposed': 'focal-proposed/ovn-22.03',
 }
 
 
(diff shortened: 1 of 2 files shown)


fnordahl (dismissed):
LGTM, the stable backport to ussuri only makes sense to me, because this is the only place this is really needed and can be easily tested.

The reason for adding it now is that it was missed for this charm when we added the `ovn-source` config for the other charms.

wolsen:
Do we need this in any other branch besides the ussuri branch? E.g. wallaby?

mkalcok:
We discussed this wrt the change in the charm itself (`neutron-api-plugin-ovn`) and conclusion was that it will most likely be only in the ussuri.

fnordahl:
>> Do we need this in any other branch besides the ussuri branch? E.g. wallaby?
> We discussed this wrt the change in the charm itself (`neutron-api-plugin-ovn`) and conclusion was that it will most likely be only in the ussuri.

Oh, but Billy is making a good point here.

Looking at the UCA pockets and OVN versions in them:
ussuri none
victoria none
wallaby 20.12
xena 21.09

So we most likely do need the `ovn-source` opotion for u, v, w and x

mkalcok:
I created backport PRs to `victoria`, `wallaby` and `xena` (links in the PR description). This change is already present in `yoga` and later.

fnordahl (approved):
LGTM, thanks!

PR #808, opened 2023-06-28
openstack: add keystone audit middleware support

arif-ali:
This PR is in preparation to fulfil  the audit middleware specification as documented in

 https://specs.openstack.org/openstack/charm-specs/specs/2023.1/backlog/audit-middleware.html

First review for nova-cloud-controller charm

https://review.opendev.org/c/openstack/charm-nova-cloud-controller/+/887213

Once this is merged, we can then do the rest of the charms in [LP#1856555](https://bugs.launchpad.net/charm-keystone/+bug/1856555)

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -554,6 +554,9 @@ class IdentityServiceContext(OSContextGenerator):
                             ctxt, keystonemiddleware_os_release)
 
                 if self.context_complete(ctxt):
+
+                    ctxt['audit_middleware'] = config('audit-middleware') or False
+
                     # NOTE(jamespage) this is required for >= icehouse
                     # so a missing value just indicates keystone needs
                     # upgrading
@@ -599,6 +602,7 @@ class IdentityCredentialsContext(IdentityServiceContext):
                 ctxt.update({
                     'service_port': rdata.get('credentials_port'),
                     'service_host': credentials_host,
+                    'service': self.service,
                     'auth_host': auth_host,
                     'auth_port': rdata.get('auth_port'),
                     'admin_tenant_name': rdata.get('credentials_project'),
--- a/charmhelpers/contrib/openstack/templates/section-audit-middleware-notifications
+++ b/charmhelpers/contrib/openstack/templates/section-audit-middleware-notifications
@@ -1,0 +1,8 @@
+{% if audit_middleware and transport_url-%}
+[audit_middleware_notifications]
+driver = {{ oslo_messaging_driver }}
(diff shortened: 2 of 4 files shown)


charmhelpers/contrib/openstack/context.py
```diff
@@ -554,6 +554,10 @@ def _resolve(key):
                             ctxt, keystonemiddleware_os_release)
 
                 if self.context_complete(ctxt):
+
+                    ctxt['audit_middleware'] = \
+                        False if config('audit-middleware') is None else config('audit-middleware')
```
ajkavanagh, line 559: Minor suggestion, but his tends to be simpler:

```python
ctxt['audit_middleware] = config('audit-middleware') or False
```

ajkavanagh (commented):
Looks good, and thanks for including a good set of unit tests.

One minor style point that is not important, but may make the code a little easier to read.

Only thing needed prior to merging is a gerrit review that exercises the change so we can verify that the change is good.  Please could you link a review to the PR's description?  Thanks.

charmhelpers/contrib/openstack/context.py
arif-ali: updated

arif-ali:
> One minor style point that is not important, but may make the code a little easier to read.

Update the PR
 
> Only thing needed prior to merging is a gerrit review that exercises the change so we can verify that the change is good. Please could you link a review to the PR's description? Thanks.

I've added the first review and the LP bug in the description. The review still needs rebasing once this PR is merged so that I can remove the location off the charmhelpers sync.

There are several other charms that need doing too, so once this PR is merged we can crack on with the others

arif-ali:
just found an issue with this, while testing, so marking it as a draft

seems like that `self.service` could be `None`, hence the process will not work; so need another way to overcome that

wolsen:
To avoid the issues encountered by using an optional field, I think this should likely just be a new Context object with the explicit needs of enabling the Audit Middleware.

ajkavanagh:
@arif-ali if the additional items are pulled out into a new context, as @wolsen suggests, it would look something like:

```python
class KeystoneAuditMiddleware(OSContextGenerator):
    def __init__(self, service_name):
        self.service_name = service_name
    def __call__(self):
        ctxt = {
            'audit_middleware': config('audit-middleware) or False,
            'service_name': service_name,
        }
        return ctxt

```
And then it can just be used by including the new Context in the `resource_map` and then it could be used in the templates as is in the current patch.  This would also significantly simplify the testing.

arif-ali:
As Myles is now working on this, I'm closing this PR

PR #827, opened 2023-07-28
[stable/yoga] Ensure get_requests_for_local_unit doesn't fail on incomplete relation (#824)

ajkavanagh:
If the certificates relation has been set from a consuming charm for
certificates, but the provider hasn't supplied a certificate, and the
request is in the 'legacy' form, then the get_request_for_local_unit()
function in cert_utils fails on accessing the (non-existant) 'ca'
parameter in the databag.  This patch ensures that if the 'ca' key is
missing that the processing moves on to the next relation.

(cherry-picked from: c8f47207c1f1652ab2cbecaaad8a83b66f004e5e)
Closes LP#2028683


jadonn:
This change looks good to me, but will probably require someone else's sign off

PR #833, opened 2023-08-05
Rename wsgi-rotation to wsgi-socket-rotation

dosaboy:


dosaboy:
not sure why pep8 is failing here, it works fine when i run locally. i have rebase and updated.

zhhuabj:
LGTM - thanks for the changes

ajkavanagh:
@dosaboy thanks for the patch.  Please could you add to the description above why these changes are necessary and, also, as it is will cause a corresponding change in charms (the config var is changing), what charms it is affecting.  At the moment, this looks like a breaking change, but it's not possible to see *what* is breaking.  Thanks!

charmhelpers/contrib/openstack/context.py
```diff
@@ -1748,9 +1748,9 @@ def __init__(self, name=None, script=None, admin_script=None,
 
     def __call__(self):
         total_processes = _calculate_workers()
-        enable_wsgi_rotation = config('wsgi-rotation')
-        if enable_wsgi_rotation is None:
-            enable_wsgi_rotation = True
+        enable_wsgi_socket_rotation = config('wsgi-socket-rotation')
```
ajkavanagh, line 1751: This causes a potentially breaking change, albeit only on master branch charms.  Please could you link to the bug or gerrit reviews where this will be consumed?  Thanks.

ajkavanagh (changes requested):
Just nacking for extra details on the change.

dosaboy:
@ajkavanagh thanks for the review. I have addressed your points in the commit msg. This is not a breaking change since the patches to implement it in the charms are blocked on this landing first.

ajkavanagh (approved):

PR #838, opened 2023-08-07
pep8: Use isinstance()

freyes:
Rule E721 requires the use of isinstance() instead of "type() == SomeType".

https://www.flake8rules.com/rules/E721.html


freyes:
Example of the failure - https://github.com/juju/charm-helpers/actions/runs/5790562939/job/15693844816?pr=834

freyes:
abandoning in favor of https://github.com/juju/charm-helpers/pull/839

PR #839, opened 2023-08-08
Remove f-strings, fix pep8 error, and py311 to matrix

ajkavanagh:
- Remove the f-strings; this caused problems with py3.5 which the
  library sort-of still supports.
- Fix the newest pep8 error.
- Add py311 testing the tox; not quite ready for github in this PR.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -50,6 +50,11 @@ basepython = python3.10
 deps = -r{toxinidir}/test-requirements.txt
 commands = nose2 {posargs} --with-coverage -s tests/
 
+[testenv:py311]
+basepython = python3.11
+deps = -r{toxinidir}/test-requirements.txt
+commands = nose2 {posargs} --with-coverage -s tests/
+
 [testenv:pep8]
 basepython = python3
 deps = -r{toxinidir}/test-requirements.txt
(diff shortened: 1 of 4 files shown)


jadonn:
The commit looks good to me, but someone else will have to sign off on merging the change.

tox.ini
```diff
@@ -50,6 +50,11 @@ basepython = python3.10
 deps = -r{toxinidir}/test-requirements.txt
 commands = nose2 {posargs} --with-coverage -s tests/
 
+[testenv:py311]
```
freyes, line 53: add it to the gh workflow matrix, please - https://github.com/juju/charm-helpers/blob/master/.github/workflows/build.yml#L18

freyes (changes requested):

tox.ini
ajkavanagh: I deliberately didn't do that for this one; py3.11 fails, and I just wanted to make the lib work with py3.5 again.  It's deliberately a smaller PR; we can tackle the breaking py3.11 issues in another PR.

freyes: ack.

freyes (approved):

PR #841, opened 2023-08-31
[stable/antelope] Add name parameter to send_application_name

rodrigogansobarbieri:
This way, multiple nova-compute applications can
share the same ceph key, as nova can invoke
send_application_name('nova-compute').

(cherry picked from commit 6a19a0a7090ea552118a7d220a49ab4c66ed6fb9)

### DIFF
--- a/charmhelpers/contrib/storage/linux/ceph.py
+++ b/charmhelpers/contrib/storage/linux/ceph.py
@@ -159,15 +159,19 @@ def get_osd_settings(relation_name):
     return _order_dict_by_key(osd_settings)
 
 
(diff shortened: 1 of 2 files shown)


ajkavanagh (approved):
Note that it's either two cherry-picks or a change and a cherry-pick as there are two commits.  However, they do match more recent branch changes.

rodrigogansobarbieri:
rebased on top of stable/antelope so the PR is now just 1 commit

coreycb (approved):
LGTM, needs one more core reviewer to land

PR #862, opened 2023-10-23
Mock get_platform / add CHARMHELPERS_IN_UNITTEST

ajkavanagh:
Patch osplatform.get_platform() to return "ubuntu" if the environment
variable CHARMHELPERS_IN_UNITTEST is set to something truthy. This is to
enable unit testing in classic charms on NON ubuntu systems (e.g. Debian
bookworm) where the function is often called via a decorator that is
evaluated at module load time and thus is very, very tricky to mock out,
as it is often 'hit' during the test discovery phase.

### DIFF
--- a/charmhelpers/osplatform.py
+++ b/charmhelpers/osplatform.py
@@ -49,8 +49,8 @@ def _get_platform_from_fs():
     return content["NAME"]
 
 
-## If the unit-test mode is set, the platform is always "ubuntu"
+# If the unit-test mode is set, the platform is always "ubuntu"
 if not os.environ.get('CHARMHELPERS_IN_UNITTEST', False):
     get_platform = _get_platform
 else:
-    get_platform = lambda: "ubuntu"
+    get_platform = lambda: "ubuntu"  # noqa
--- a/charmhelpers/osplatform.py
+++ b/charmhelpers/osplatform.py
@@ -2,7 +2,7 @@ import platform
 import os
 
 
-def get_platform():
+def _get_platform():
     """Return the current OS platform.
 
     For example: if current os platform is Ubuntu then a string "ubuntu"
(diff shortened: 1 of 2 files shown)


charmhelpers/osplatform.py
```diff
@@ -47,3 +47,10 @@ def _get_platform_from_fs():
     for k, v in content.items():
         content[k] = v.strip('"')
     return content["NAME"]
+
+
+# If the unit-test mode is set, the platform is always "ubuntu"
+if not os.environ.get('CHARMHELPERS_IN_UNITTEST', False):
```
freyes, line 53: the docs suggest that setting this env variable to "false" (considering that 'true' is valid to opt in), the doc should say something like "anything that python evaluates to true"

freyes (changes requested):
Not pretty, but it gets the job done, and considering the scale at which we operate charm-helpers this approach is reasonable to me. I just left one comment inline.

ajkavanagh:
@samuelengineer yes, it's not detecting Debian bookworm due to it being "Debian" rather than "debian".  We should probably throw a  `.lower()` into the string to make detection case agnostic.  However, it's also fundamentally broken as Debian doesn't use lsb_release and that also fails.  Maybe we should just rip out Debian support (and not pretend it is Ubuntu) and either 'do it properly', or not at all.  I favour the latter as charm-helpers is deprecated (sort of) due to the ops framework.

Also, I'm going to close this as we actually found a much simpler way of detailing with the issue; mocking the functions out  in `unit_tests/__init__.py` and making them return a sensible default for the charm under test. This has the added benefit of further (and more properly) insulating the unit tests from the platform they are being run on.  See https://review.opendev.org/c/openstack/charm-nova-compute-nvidia-vgpu/+/899177/2/unit_tests/__init__.py as an example.

PR #876, opened 2024-01-18
Add caracal support

ajkavanagh:
This patch enables caracal in the various places needed for charms to be
able to detect and operate with either jammy-caracal or caracal on the
nobel distro.


charmhelpers/fetch/ubuntu.py
```diff
@@ -308,6 +317,7 @@
     ('kinetic', 'zed'),
     ('lunar', 'antelope'),
     ('mantic', 'bobcat'),
+    ('nobel', 'caracal'),
```
javacruft, line 320: nobel -> noble

javacruft (approved):

PR #880, opened 2024-03-06
Fix platform check for Debian 12

wolsen:
The osplatform.get_platform() check fails on Debian Bookworm because the output determined by the /etc/os-release file has "Debian GNU/Linux" and the get_platform() check only looks for lowercase debian.

This is causing tests to fail in the OpenStack Charms CI, as seen here - https://review.opendev.org/c/openstack/charm-keystone-openidc/+/908638

### DIFF
--- a/tests/test_osplatform.py
+++ b/tests/test_osplatform.py
@@ -1,0 +1,74 @@
+#
+# Copyright (C) 2024 Canonical Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+# implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+try:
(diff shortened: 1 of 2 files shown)


hemanthnakkina (approved):

tests/test_osplatform.py
```diff
@@ -0,0 +1,74 @@
...
+    import unittest.mock as mock
+except ImportError:
+    import mock
+
+from charmhelpers import osplatform
+
+
+class TestPlatform(unittest.TestCase):
+
+    @mock.patch.object(osplatform, "_get_current_platform")
```
fnordahl, line 28: nit: there is a `BaseTestCase` class in [tests/utils.py](https://github.com/juju/charm-helpers/blob/b78107dc750644b1d868ff4a61748086783e02bd/tests/utils.py#L39) that allows you to do these patches in-line as opposed to using the decorators.

The decorator approach quickly becomes hard to manage with all the positional arguments etc.

fnordahl (approved):
The approach LGTM, thanks for the extended test coverage, although I have a non-blocking nit on the unit tests.

tests/test_osplatform.py
ajkavanagh: I don't disagree with the nit, but I guess for expediency, let's merge this.  I think there's a larger body of work in re-working (or adding separate methods) `.patch(...)` and `.patch_object` to also/instead return the mock so that static code analysers in vscode and Neovim+LSP don't complain about the magic members that are added to the object as part of the patching functionality.

PR #882, opened 2024-03-22
Skip service_resume if already enabled

dosaboy:
Repeatedly calling service_resume for a service that is already enabled can lead to unintended consequences. This adds a check to only resume a serice if it is not enabled.

Related-Bug: #2058505


ajkavanagh (changes requested):
The unit test(s) should probably be updated/added to to test the new feature.

brianphaley (approved):
lgtm

arif-ali (approved):
just tested this through the ovn-chassis charm, and works as expected

fnordahl (approved):
LGTM, thanks!

ajkavanagh (approved):

PR #884, opened 2024-03-28
Skip service_resume if already enabled

brianphaley:
Repeatedly calling service_resume for a service that is already enabled can lead to unintended consequences. This adds a check to only resume a serice if it is not enabled.

Related-Bug: #2058505
(cherry picked from commit a334500)


fnordahl (approved):
LGTM, thanks!

PR #885, opened 2024-03-28
Skip service_enable if already enabled

brianphaley:
Repeatedly calling service_enable for a service that is already enabled can lead to unintended consequences. Some charms frequently call servie_resume which will call service('enable') and this adds a check to only do so of the service is not enabled.

Related-Bug: #2058505
(cherry picked from commit a3345005e597f098909b7c12dfee630a366ce0f6)


fnordahl (approved):
LGTM, thanks!

PR #886, opened 2024-03-28
Skip service_enable if already enabled

brianphaley:
Repeatedly calling service_enable for a service that is already enabled can lead to unintended consequences. Some charms frequently call servie_resume which will call service('enable') and this adds a check to only do so of the service is not enabled.

Related-Bug: #2058505
(cherry picked from commit a3345005e597f098909b7c12dfee630a366ce0f6)

### DIFF
--- a/charmhelpers/core/host.py
+++ b/charmhelpers/core/host.py
@@ -256,8 +256,11 @@ def service_resume(service_name, init_dir="/etc/init",
     upstart_file = os.path.join(init_dir, "{}.conf".format(service_name))
     sysv_file = os.path.join(initd_dir, service_name)
     if init_is_systemd(service_name=service_name):
-        service('unmask', service_name)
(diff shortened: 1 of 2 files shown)


brianphaley (commented):
From looking at the failure looks like we need this change:

0f7ca5396edd6a932aada59385b90bd95739b30c

I'll cherry-pick that in another pull request.

fnordahl:
Thank you for putting forward multiple options to solve this, I chose to merge #890 which would allow us to consider this PR in a uniform manner along with the other backports. Would you be able to rebase?

brianphaley:
Sure, I can rebase this one. I threw the squashed option because i wanted them both tested together, but github had other plans. The cherry-pick for 890 at least fixed the unit test failure locally.

fnordahl (approved):
LGTM, thanks!

PR #904, opened 2024-06-26
Use looseversion in place of distutils

freyes:
(cherry picked from commit 012b2d65e69d564a8741bae25e2b22b145f40d27)

### DIFF
--- a/charmhelpers/core/hookenv.py
+++ b/charmhelpers/core/hookenv.py
@@ -18,7 +18,10 @@
 #  Charm Helpers Developers <juju@lists.ubuntu.com>
 
 import copy
-from distutils.version import LooseVersion
+try:
+    from distutils.version import LooseVersion
+except ImportError:
+    from looseversion import LooseVersion
 from enum import Enum
 from functools import wraps
 from collections import namedtuple, UserDict
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,5 @@ Jinja2
 netaddr
 
 pbr!=2.1.0,>=2.0.0 # Apache-2.0
+
+looseversion;python_version >= '3.12'


freyes:
Blocker for s390x validation

requirements.txt
```diff
@@ -12,3 +12,5 @@ Jinja2
 netaddr
 
 pbr!=2.1.0,>=2.0.0 # Apache-2.0
+
+looseversion;python_version >= '3.12'
```
freyes, line 16: how does this work when being used in a classic charm?

charmhelpers/core/hookenv.py
```diff
@@ -18,7 +18,10 @@
 #  Charm Helpers Developers <juju@lists.ubuntu.com>
 
 import copy
-from distutils.version import LooseVersion
+try:
+    from distutils.version import LooseVersion
+except ImportError:
+    from looseversion import LooseVersion
```
freyes, line 24: shouldn't this be more like:

```python
try:
    from distutils.version import LooseVersion
except ImportError:
    try:
        # reactive charm path, where the venv already has this package
        from looseversion import LooseVersion
    except ImportError:
        # classic charm code path where python packages are installed via apt-get
        apt_install(['python3-looseversion'], fatal=True)
        from looseversion import LooseVersion
```

ajkavanagh: I really don't like this pattern where charmhelpers runs `apt_install`; I'd much rather the parent charm resolved the dependencies and installed python3-looseversion where necessary; (which might be all of them in the future).  It makes unit testing harder when `apt_install()` is at module load time.

requirements.txt
ajkavanagh: Based on your comment above, it probably breaks unless the parent charm ensures that the package is installed.

ajkavanagh (approved):

ajkavanagh:
Holding off merging whilst @freyes conversation is concluded; there may be further changes needed to the original commits.

PR #906, opened 2024-07-01
Allow custom shell when creating system users

Adithya-Rajendran:
Allow setting the shell when creating system users using adduser()
Changed defaults to `/usr/sbin/nologin` for system users kept `/bin/bash` as default shell for normal users.


hemanthnakkina (approved):
LGTM

freyes (approved):
LGTM.

PR #908, opened 2024-07-25
[stable/zed] Add MySQL password setter that uses existing connection

ajkavanagh:
This modifies the existing method `set_mysql_password` and splits it
into the original and a 2nd part called
`set_mysql_password_using_current_connection`.  This allows a root
connection to be set up and then the password for any user to to be
updated, so that a mysql instance charm can update the password.

(cherry picked from commit 61ecb2eb7387e9d7603b149447441e07b6dca920)


freyes (approved):
LGTM.

PR #919, opened 2025-03-20
Adding send-proxy-v2 to see actual client ip

xtrusia:
currently, log shows proxy ip but need actual client ip

This PR suggest to add required module and
configuration method to setup options for Apache2 templates
and options for haproxy templates


charmhelpers/contrib/openstack/templates/openstack_https_frontend
```diff
@@ -22,6 +22,9 @@ Listen {{ ext_port }}
     ProxyPassReverse / http://localhost:{{ int }}/
     ProxyPreserveHost on
     RequestHeader set X-Forwarded-Proto "https"
+    {% if haproxy_enable_proxyv2 -%}
+    RemoteIPProxyProtocol On
```
freyes, line 26: did you consider setting `RemoteIPHeader X-Forwarded-For` as well? .... otherwise how Apache2 knows what ip address to use?

https://httpd.apache.org/docs/2.4/mod/mod_remoteip.html#remoteipheader

freyes (changes requested):

charmhelpers/contrib/openstack/templates/openstack_https_frontend
xtrusia: Hello, yes proxyv2 protocol provides the same function and safety.
https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

but please let me know if you have any advice for the header.

Thanks.

freyes: ack.

freyes (changes requested):
you are putting this feature behind a config option , does this mean all charms will be updated to include a new option in their config.yaml? ... can't we just make this the default behavior?, is there any cons on making this the default?

xtrusia:
> you are putting this feature behind a config option , does this mean all charms will be updated to include a new option in their config.yaml? ... can't we just make this the default behavior?, is there any cons on making this the default?

You're right.
I actually thought the same at first, and it seemed fine to just make it the default behavior.

But I had a question during testing. For example, in the keystone charm, there are its own templates for apache2. e.g wsgi-openstack-api.conf and openstack_https_frontend.conf.

I tested this, and it turns out that if a charm includes its own template files, it uses those instead of the ones from charmhelpers.
So even if we make this the default behavior in charmhelpers, we'd still need to update each individual charm that overrides the templates.

Also, fixes would need to be applied across haproxy, apache templates, and modules all at once, since they're all interdependent.
So charmhelpers(sync with charm) and the keystone charm itself(templates update) should be updated together, otherwise things will break I think.

If we adopt conf option, we can split commit to sync with charmhelpers and change each charms.

But yes. making this the default also makes sense to me as well.

xtrusia:
@freyes If we can do both at once, no issue.
if haproxy.conf has send-proxy-v2 but apache2 conf doesn't have it (vice versa) it doesn't seem to work. so I mentioned it.

So, would you like me to change it to default instead of adding config option? I'll change the fix tomorrow.

charmhelpers/contrib/openstack/templates/haproxy.cfg
```diff
@@ -88,9 +88,9 @@ backend {{ service }}_{{ frontend }}
     {% endif -%}
     {% for unit, address in frontends[frontend]['backends'].items() -%}
     {% if https -%}
-    server {{ unit }} {{ address }}:{{ ports[1] }} check check-ssl verify none
+    server {{ unit }} {{ address }}:{{ ports[1] }} check check-ssl verify none send-proxy-v2 check-send-proxy
```
freyes, line 91: why are you using check-send-proxy only for https traffic?

https://docs.haproxy.org/2.5/configuration.html#check-send-proxy

freyes (changes requested):
something we haven't talked about is what happens when you have only side of the connection with the proxy protocol enabled, specifically:

1. haproxy set with send-proxy-v2, but apache doesn't have remoteip enabled
2. haproxy is NOT set with send-proxy-v2, but apache has enabled remoteip

I would assume in the case of (2) nothing happens, and the old behavior where the client ip address is not received by apache, right?, but what about in the case of (1)?

charmhelpers/contrib/openstack/templates/haproxy.cfg
xtrusia: right I think I missed. I added it to for http as well. Thanks!

xtrusia:
> something we haven't talked about is what happens when you have only side of the connection with the proxy protocol enabled, specifically:
> 
> 1. haproxy set with send-proxy-v2, but apache doesn't have remoteip enabled
> 2. haproxy is NOT set with send-proxy-v2, but apache has enabled remoteip
> 
> I would assume in the case of (2) nothing happens, and the old behavior where the client ip address is not received by apache, right?, but what about in the case of (1)?

This is why I mentioned in the beginning they should be fixed at the same time.(for keystone) but sorry my explanation was not enough.
I remember that (1) didn't work. keystone command returned error.
But I'll double check it today and share the result here.
Thanks so much.

xtrusia:
@freyes I'm wondering if you might check the latest comment and have some time to review this issue. Thanks for your review.

xtrusia:
re-added configuration option since we discussed it in sprint.

lathiat:
1. I think it would be nice to have a draft PR up for charm-keystone incorporating these changes, so we can review/test the implementation before this is merged. As we may find we later need changes and have to do a second PR. Could you upload such a change in draft mode for review?

It would also be great to explicitly document the following

2. Do both apache2 and haproxy support these exact same modules/configuration options on focal (for focal-yoga) as well as jammy + noble. The requesting environment is Jammy-Yoga so the intent is to backport this feature to at least Yoga. But the same charm has to support both Jammy-Yoga and Focal-Yoga.

3. Do we (now) intend for the config option to be disabled or enabled by default. While avoiding an unexpected change on existing environments is ideal, I can't really think of scenarios where people would NOT want this option enabled - if they are using both haproxy and apache2. I agree it would be ideal to have an option to turn it off in case of a problem. If we are going to go ahead with having it disabled by default, can we document exactly why - is it only abundance of caution, or otherwise what specific scenarios may be broken without it.

4. What happens during either charm upgrades (or just charm config change) when only 1 or 2 units have finished upgrading their config file and restarting. (since you do 1 unit at a time, which hosts both) - in cases where the config option is both off and on. Will it still work, or fail during that time until all units have updated. This may be an issue for managed upgrades where daemon restarts are not being done. When enabled, RemoteIPProxyProtocol requires that clients connect with PROXYv2 by default, so I suspect this won't work.

Lastly, some review notes

5. When deployed in a non-HA configuration, that this is not enabled. I am not sure this is the case currently? Perhaps by default we would only activate this option when haproxy is in use for a HA deployment with the hacluster relation, but the config option could force it on otherwise.

6. When enabled, RemoteIPProxyProtocol requires that clients connect with PROXYv2. We should check whether any charms have NRPE checks or other reasons the port is conneted to with a non-PROXYv2 client that will be broken by this. It may also frustrate us at debug time, when you try to directly query the apache2 server when you are trying to debug problems for example. Perhaps it's desirable to configure mod_remoteip to -only- request PROXYv2 from the specific list of haproxy IP addresses. Seems this may be possible with RemoteIPProxyProtocolExceptions.

7. That other clients cannot speak Proxy V2 and spoof the client source IP. Currently, we don't firewall off apache2 access and many clients of the customer's network can often connect directly to the apache2 port, bypassing haproxy. So we need to prevent this by configuring either a firewall, or using configuration options like RemoteIPTrustedProxy or RemoteIPInternalProxy to limit this access to certain IP addresses - e.g. the IP addresses of the other units haproxy instance. 

8. Looking at the apache2 documentation, it seems it ignores RFC1918/Intranet IP addresses unless a specific RemoteIPInternalProxy is configured. I think in practice, most Charmed OpenStack customers will be using such addresses, and so we may need to configure this option specifically otherwise the config option may be mostly useless.

charmhelpers/contrib/openstack/context.py
```diff
@@ -1034,6 +1034,9 @@ def __call__(self):
         if config('haproxy-connect-timeout'):
             ctxt['haproxy_connect_timeout'] = config('haproxy-connect-timeout')
 
+        if config('haproxy-enable-proxyv2'):
+            ctxt['haproxy_enable_proxyv2'] = config('haproxy-enable-proxyv2')
```
dosaboy: this can be set to True since the config is actually boolean. Just makes it clearer that it doesn't need to be set to a more specific value.

tests/contrib/openstack/test_os_contexts.py
```diff
@@ -2929,6 +2933,66 @@ def test_haproxy_context_with_prometheus_exporter(self, local_unit, local_addres
                                                call('cluster'),
                                                call('haproxy-exporter')])
 
+    @patch('charmhelpers.contrib.openstack.context.local_address')
+    @patch('charmhelpers.contrib.openstack.context.local_unit')
+    def test_haproxy_context_with_data_proxyv2(self, local_unit, local_address):
```
dosaboy: consider renaming this to imply testing apache context rather than haproxy

dosaboy (commented):
Please also add the ```Closes-Bug: #2107999``` to the commit message

tests/contrib/openstack/test_os_contexts.py
xtrusia: this is for haproxy context and for apache, there is another one.

charmhelpers/contrib/openstack/context.py
```diff
@@ -1128,7 +1131,8 @@ class ApacheSSLContext(OSContextGenerator):
     user = group = 'root'
 
     def enable_modules(self):
-        cmd = ['a2enmod', 'ssl', 'proxy', 'proxy_http', 'headers']
+        # NOTE(seyeongkim) : remoteip is for proxy protocol v2
```
dosaboy: should adding this new module be conditional on if config('haproxy-enable-proxyv2')?

xtrusia: Right. I considered it in the beginning but like ssl module, I just put 'enabling remoteip module' only, Because we can enable/disable it by configuration option.
How do you think about this approach? conditional enabling could be better? 
Thanks so much.

```
    def enable_modules(self):
        cmd = ['a2enmod', 'ssl', 'proxy', 'proxy_http', 'headers']
        check_call(cmd)
        # remoteip is only needed for proxy protocol v2
        if config('haproxy-enable-proxyv2'):
            check_call(['a2enmod', 'remoteip'])
        else:
            # Disable remoteip module
            try:
                check_call(['a2dismod', 'remoteip'])
            except CalledProcessError:
                # ignore error
                pass
```

lathiat:
Reviewed further today with Seyeong. The problem with the current approach is that the PROXYv2 protocol allows anyone to connect from anywhere, and send a client IP that is spoofed. While anyone can connect from anywhere in the current version of the charm, they cannot spoof the client IP because the mod_remoteip module is not enabled.

We found that my previously analysis was slightly wrong, the RemoteIPTrustedProxy and RemoteIPInternalProxy options -only- apply when you are using an "X-Forwarded-For" header over the standard HTTP protocol, and do not work/apply when using the PROXYv2 protocol (as currently enabled by the patch). When using the PROXYv2 protocol there is no Apache2 option to enable/disable the protocol.

So I have requested that the patch is instead changed to use X-Forwarded-For, and use RemoteIPInternalProxy to limit the hosts which can send this header, to the keystone unit IPs. This also has the advantage of preventing any issue if clients (e.g. NRPE or manual tests) connect directly to the backend port as it now just speaks the regular HTTP protocol. In the same way it also removes any upgrade issues. That said, mostly those issues seemed to not be an issue in practice, but it can depend a little on the consuming charm. So with this new method, we'll have confidence it's all good.

This also means we can probably just enable this behaviour by default, as the only difference is a client can optionally send an X-Forwarded-For header and it will only be processed when you are coming from the keystone unit IPs. So the need to turn it off/on is not really required any longer. However for back-porting to Yoga/Caracal we may want to consider such an option still to minimise regression potential but strictly it shouldn't be necessary.

Once that is done all of my concerns are resolved and we can move forward.

lathiat:
@dosaboy It would be good to get your input on whether you think we can just skip making this a configuration option with the new approach, or whether you would be hesitant to backport it without that.

In reality this behaviour should be enabled by default anyway, it's a real miss that we never had it this long. And with the new approach, there are no connection compatibility issues to worry about like we had when it was using a separate wrapper "PROXYv2" connection. Now it will just be inserting an extra HTTP header instead.

xtrusia:
updated the commit to use x forwarded for and tested.
1. mode http : xforwardfor option needed this. func test passed but need another eyes.
2. config option is removed for now. maybe no need
3. remoteip module is needed but no need to uninstall when conf is false.

Raven-182:
@xtrusia I tested this with the neutron-api charm with TLS enabled. Without the `check` keyword on the server lines, healthchecks were being skipped. After adding `check`, the healthchecks started, but they are failing.  

Switching `send-proxy-v2` (PROXYv2) to `send-proxy` (PROXYv1) resolves the issue. I haven’t been able to pinpoint the exact cause, do you know if there’s a specific reason PROXYv2 was chosen? Functionally, the two seem equivalent in this context.

Issue #920, opened 2025-04-08
SharedDB context fails when there's not yet a db_host


gboutry:
Trying to deploy a simple keystone + mysql router + mysql is failing on noble-epoxy ppa with:

```
unit-keystone-0: 14:28:40 WARNING unit.keystone/0.shared-db-relation-joined Traceback (most recent call last):
unit-keystone-0: 14:28:40 WARNING unit.keystone/0.shared-db-relation-joined   File "/var/lib/juju/agents/unit-keystone-0/charm/hooks/shared-db-relation-joined", line 952, in <module>
unit-keystone-0: 14:28:40 WARNING unit.keystone/0.shared-db-relation-joined     main()
unit-keystone-0: 14:28:40 WARNING unit.keystone/0.shared-db-relation-joined   File "/var/lib/juju/agents/unit-keystone-0/charm/hooks/shared-db-relation-joined", line 948, in main
unit-keystone-0: 14:28:40 WARNING unit.keystone/0.shared-db-relation-joined     assess_status(CONFIGS)
unit-keystone-0: 14:28:40 WARNING unit.keystone/0.shared-db-relation-joined   File "/var/lib/juju/agents/unit-keystone-0/charm/hooks/keystone_utils.py", line 2688, in assess_status
unit-keystone-0: 14:28:40 WARNING unit.keystone/0.shared-db-relation-joined     assess_status_func(configs)()
unit-keystone-0: 14:28:40 WARNING unit.keystone/0.shared-db-relation-joined   File "/var/lib/juju/agents/unit-keystone-0/charm/charmhelpers/contrib/openstack/utils.py", line 1825, in _assess_status_func
unit-keystone-0: 14:28:40 WARNING unit.keystone/0.shared-db-relation-joined     state, message = _determine_os_workload_status(*args, **kwargs)
unit-keystone-0: 14:28:40 WARNING unit.keystone/0.shared-db-relation-joined                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
unit-keystone-0: 14:28:40 WARNING unit.keystone/0.shared-db-relation-joined   File "/var/lib/juju/agents/unit-keystone-0/charm/charmhelpers/contrib/openstack/utils.py", line 1024, in _determine_os_workload_status
unit-keystone-0: 14:28:40 WARNING unit.keystone/0.shared-db-relation-joined     state, message = _ows_check_generic_interfaces(
[...]

PR #921, opened 2025-04-08
c/network/ip: return false when no address given

gboutry:
Since netaddr 1.2.0 [0], netaddr fails to parse an IPv6 if the given ipv6 is None. Consider a falsy address as not an IPv6 address.

This code used to return False when given a falsy ipaddress.

0: https://github.com/netaddr/netaddr/commit/d39b7d12428006b2ca1e260f92da2b66948b0d6d

Closes: #920


wolsen (approved):
LGTM

PR #923, opened 2025-04-16
Enable python 3.12 compatibility in testing environment

gboutry:
Fix github workflows by moving to jammy, drop support for python 3.6 and 3.7.

F824 rule appeared when switching tox env to 3.10 (consequence of moving to jammy host)

Switch default nosetests command to nose2.

Enable python 3.12 compatibility.


wolsen (approved):
LGTM

Issue #924, opened 2025-05-21
Failure to handle dns.exception.Timeout can cause dns timeout and subsequently lead to hook failure


zhhuabj:
A customer got an intermittent alert in the juju status output related to cinder-volume units with the update-status hook failure. The following unit-cinder-volume-1.log shows that the issue was caused by a dns timeout.

```
2024-09-25 17:16:32 WARNING unit.cinder-volume/1.update-status logger.go:60 ERROR no relation id specified
2024-09-25 17:16:37 WARNING unit.cinder-volume/1.update-status logger.go:60 Traceback (most recent call last):
......
2024-09-25 17:16:37 WARNING unit.cinder-volume/1.update-status logger.go:60 dns.exception.Timeout: The dns operation timed out after 5.4017493724823 seconds
```

My guess is that the dns.exception.Timeout exception was previously unhandled , which may have led to performance degradation in the cinder-volume units. This could have resulted in timeout problems when using the default value resolv.lifetime=5.

PR #925, opened 2025-05-23
catch dns.exception.Timeout to prevent hook failure

zhhuabj:
Avoid unhandled dns.exception.Timeout which were causing intermittent update-status hook failures in cinder-volume units. Now we log a warning and return None, improving robustness without increasing dns timeout.

Closes: #924

### DIFF
--- a/charmhelpers/contrib/network/ip.py
+++ b/charmhelpers/contrib/network/ip.py
@@ -469,8 +469,20 @@ def ns_query(address):
         return None
 
     try:
-        answers = dns.resolver.query(address, rtype)
+        resolv = dns.resolver.Resolver()
+        # The dnspython library sets a default DNS query lifetime of 5.0 seconds,
+        # as defined in BaseResolver.reset() (see https://github.com/rthalley/
+        # dnspython/blob/7ed1648b/dns/resolver.py#L709, commit 7ed1648b).
+        resolv.lifetime = config('dns-query-timeout') or 5.0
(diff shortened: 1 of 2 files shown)


charmhelpers/contrib/network/ip.py
```diff
@@ -469,9 +469,15 @@ def ns_query(address):
         return None
 
     try:
-        answers = dns.resolver.query(address, rtype)
+        resolv = dns.resolver.Resolver()
+        resolv.lifetime = config('dns-query-timeout') or 5.0
+        answers = resolv.query(address, rtype)
     except (dns.resolver.NXDOMAIN, dns.resolver.NoNameservers):
         return None
+    except (dns.exception.Timeout):
```
dosaboy, line 477: no need for brackets - only need if catching more than one exception

dosaboy (commented):
looks good, just one nit

dosaboy (approved):
lgtm

charmhelpers/contrib/network/ip.py
```diff
@@ -469,9 +469,18 @@ def ns_query(address):
         return None
 
     try:
-        answers = dns.resolver.query(address, rtype)
+        resolv = dns.resolver.Resolver()
```
dosaboy, line 472: While im confident that 5 seconds is the default dns timeout in Ubuntu I think this needs a comment to say that and if possible how this can be asserted.

dosaboy (approved):
lgtm

PR #927, opened 2025-05-28
Allow non-dynamic addresses for ipv6 relation_ips

wolsen:
When looking for relation IP addresses, the current code filters for only dynamic addresses. This doesn't work on MAAS where it assigns fixed-address6 for the dhcpd6 configuration - e.g.

```
host 5c-ed-8c-e8-d5-b4 {
   #
   # Node DHCP snippets
   #
   # No DHCP snippets defined for host

   hardware ethernet 5c:ed:8c:e8:d5:b4;
   fixed-address6 fc00:1:1::2;
}
```

This causes the local node to not have the address as dynamic. Change the default option to set dynamic_only=False

### DIFF
--- a/charmhelpers/contrib/network/ip.py
+++ b/charmhelpers/contrib/network/ip.py
@@ -620,7 +620,7 @@ def get_relation_ip(interface, cidr_network=None):
         # Currently IPv6 has priority, eventually we want IPv6 to just be
         # another network space.
         assert_charm_supports_ipv6()
-        return get_ipv6_addr()[0]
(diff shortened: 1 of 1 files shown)


freyes (approved):
We discussed this change in length, we considered what possible use causes the original author had in mind to leave non-dynamic addresses out, we couldn't come up with a good reason in how things are today (e.g. how current MAAS configures ip addresses).

So we will be changing the filtering, and backporting this to stable releases since we believe this is the desired behavior.

PR #933, opened 2025-06-02
Skip netaddr call if address is None

MylesJP:
A change was introduced to netaddr that no longer allows a None-type object to be passed to the IP parsing functions. This PR allows the netaddr call to be skipped if no IP address is provided. See bug report:

https://bugs.launchpad.net/ubuntu/+source/python-netaddr/+bug/2112346

### DIFF
--- a/charmhelpers/contrib/network/ip.py
+++ b/charmhelpers/contrib/network/ip.py
@@ -137,7 +137,10 @@ def is_ipv6(address):
     if not address:
         return False
     try:
-        address = netaddr.IPAddress(address)
+        if address:
+            address = netaddr.IPAddress(address)
+        else:
+            return False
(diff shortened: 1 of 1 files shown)


charmhelpers/contrib/network/ip.py
```diff
@@ -137,7 +137,10 @@ def is_ipv6(address):
     if not address:
         return False
     try:
-        address = netaddr.IPAddress(address)
+        if address:
```
gboutry, line 140: Is that not covered by the check at L137?

gboutry (changes requested):
Isn't https://github.com/juju/charm-helpers/pull/921 enough to fix your issue?

That was never synced to the charms themselves, but on locally built charm, I remember it did fix that particular issue.

PR #956, opened 2026-04-08
Add the ability to specify PCI address in dpdk-bond-mappings.

PanosKostopoulos:
Since mac addresses change often in deployment, an option for specifying PCI address is very useful.

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -2678,6 +2678,56 @@ def resolve_pci_from_mapping_config(config_key):
 
     return resolved_devices
 
+def is_pci_address(value):
+    parts = value.split(':')
+
+    # Expected: 0000:3b:00.0
+    if len(parts) != 3:
+        return False
+
+    last = parts[-1]
+
+    if '.' not in last:
+        return False
+
+    slot, function = last.split('.', 1)
+
+    return (
+        len(parts[0]) == 4 and
+        len(parts[1]) == 2 and
+        len(slot) == 2 and
+        len(function) == 1
+    )
+
+
+def is_mac_address(value):
+    parts = value.split(':')
+
+    # Expected: aa:bb:cc:dd:ee:ff
+    return len(parts) == 6 and all(len(part) == 2 for part in parts)
+
+def split_ports(mapping):
+    """Split a map of mixed bond-pci, bond-mac elements into two.
+    :returns mac-ifname bond map, pci bond map
+    :returns: Dict,Dict
+    """
+    pci = {}
+    macs = {}
+    names = {}
+
+    for port, bond in mapping.items():
+        if is_pci_address(port):
+            pci[port] = bond
+        elif is_mac_address(port):
+            macs[port] = bond
+        else:
+            names[port] = bond
+
+    return names, macs, pci
+
+def build_mapping_string(mapping):
+    """ Create dpdk-bond-mappings config option string from port, bond map."""
+    return ' '.join(f"{bond}:{port}" for port, bond in mapping.items())
 
 class DPDKDeviceContext(OSContextGenerator):
 
@@ -2706,8 +2756,14 @@ class DPDKDeviceContext(OSContextGenerator):
             return {}
         # Resolve PCI devices for both directly used devices (_bridges)
         # and devices for use in dpdk bonds (_bonds)
+        data_port_mappings = parse_data_port_mappings(config(self.bonds_key))
+        ifname_bonds, mac_bonds, pci_bonds = split_ports(data_port_mappings)
+
+        updated_bonds_key = build_mapping_string(ifname_bonds | mac_bonds)
+
         pci_devices = resolve_pci_from_mapping_config(self.bridges_key)
-        pci_devices.update(resolve_pci_from_mapping_config(self.bonds_key))
+        pci_devices.update(pci_bonds)
+        pci_devices.update(resolve_pci_from_mapping_config(updated_bonds_key))
         return {'devices': pci_devices,
                 'driver': driver}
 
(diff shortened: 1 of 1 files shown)


charmhelpers/contrib/openstack/context.py
```diff
@@ -2660,24 +2660,28 @@ def resolve_pci_from_mapping_config(config_key):
...
-            db.set(mac, pcidev.pci_address)
-            db.flush()
-
-        pci_address = db.get(mac)
-        if pci_address:
-            resolved_devices[pci_address] = EntityMac(entity, mac)
-
-    return resolved_devices
-
+    PCI_REGEX = r'^([0-9a-fA-F]{4}:)?[0-9a-fA-F]{2}:[0-9a-fA-F]{2}\.[0-7]$'
```
fnordahl: 1. The purpose of the `resolve_pci_from_mapping_config` function is to translate MAC into PCI addresses and as such I'd not modify it for your purpose.
   Instead, I'd suggest to create a new function that can be called from somewhere around here: https://github.com/juju/charm-helpers/blob/5488b35a8d271ab2666111784121d4f65e1eca2e/charmhelpers/contrib/openstack/context.py#L2698-L2712
2. Regular expressions are prohibitively difficult to read and maintain, please make use of the rich Python string helpers instead, there is some potential inspiration in the `parse_*` functions here: https://github.com/juju/charm-helpers/blob/5488b35a8d271ab2666111784121d4f65e1eca2e/charmhelpers/contrib/openstack/neutron.py#L274-L351

fnordahl (changes requested):

charmhelpers/contrib/openstack/context.py
PanosKostopoulos: The issue here is that the parsing of the input is being done inside the resolve_pci_from_mapping_config. Bring it out of the function will break others using this. 
I will implement something like this, first extracting the pci_addresses from the bonds_key and proceeding as before with the rest

pci_devices = resolve_pci_from_mapping_config(self.bridges_key)
pci_addresses, new_bonds_key = extract_pci_devices(self.bonds_key))
pci_devices.update(pci_adresses)
pci_devices.update(resolve_pci_from_mapping_config(new_bonds_key)

fnordahl: > The issue here is that the parsing of the input is being done inside the resolve_pci_from_mapping_config. Bring it out of the function will break others using this.

Not sure I understand this comment, as my comment was not to change `resolve_pci_from_mapping_config`.

> I will implement something like this, first extracting the pci_addresses from the bonds_key and proceeding as before with the rest

There is an opportunity to have this generically apply to both `bridge-interface-mappings` and `dpdk-bond-mappings` configuration options, and I see no reason not to do that?

charmhelpers/contrib/openstack/context.py
```diff
@@ -2818,8 +2845,14 @@ def devices(self):
...
         """
+        data_port_mappings = parse_data_port_mappings(config(self.bonds_key))
+        mac_ifname_bonds, pci_bonds = split_ports(data_port_mappings)
+
+        updated_bonds_key = build_mapping_string(mac_ifname_bonds)
+
         pci_devices = resolve_pci_from_mapping_config(self.bridges_key)
-        pci_devices.update(resolve_pci_from_mapping_config(self.bonds_key))
+        pci_devices.update(pci_bonds)
+        pci_devices.update(resolve_pci_from_mapping_config(updated_bonds_key))
```
fnordahl, line 2884: The `resolve_pci_from_mapping_config` function takes a string argument that refers to a configuration option.

The dict you are passing in will not match any configuration option, and as such you have effectively made this call a NO-OP.

charmhelpers/contrib/openstack/context.py
```diff
@@ -2706,8 +2727,14 @@ def __call__(self):
...
         # and devices for use in dpdk bonds (_bonds)
+        data_port_mappings = parse_data_port_mappings(config(self.bonds_key))
+        mac_ifname_bonds, pci_bonds = split_ports(data_port_mappings)
+
+        updated_bonds_key = build_mapping_string(mac_ifname_bonds)
+
         pci_devices = resolve_pci_from_mapping_config(self.bridges_key)
-        pci_devices.update(resolve_pci_from_mapping_config(self.bonds_key))
+        pci_devices.update(pci_bonds)
+        pci_devices.update(resolve_pci_from_mapping_config(updated_bonds_key))
```
fnordahl, line 2766: The `resolve_pci_from_mapping_config` function takes a string argument that refers to a configuration option.

The dict you are passing in will not match any configuration option, and as such you have effectively made this call a NO-OP.

charmhelpers/contrib/openstack/context.py
```diff
@@ -2678,6 +2678,27 @@ def resolve_pci_from_mapping_config(config_key):
...
+        if '.' in port and port.count(':') >= 2:
+            pci[port] = bond
+        else:
+            others[port] = bond
+
+    return others, pci
+
+def build_mapping_string(mapping):
+    """ Create dpdk-bond-mappings config option string from port, bond map."""
+    return ' '.join(f"{bond}:{port}" for port, bond in mapping.items())
```
fnordahl, line 2731: This function is likely not necessary.

charmhelpers/contrib/openstack/context.py
```diff
@@ -2678,6 +2678,27 @@ def resolve_pci_from_mapping_config(config_key):
...
+    others = {}
+
+    for port, bond in mapping.items():
+        # PCI: 0000:3b:00.0 pattern
+        if '.' in port and port.count(':') >= 2:
+            pci[port] = bond
+        else:
+            others[port] = bond
+
+    return others, pci
```
fnordahl: Instead of doing this, make the parsing functions only parse the entities they have superior knowledge about (e.g. one for macs another for PCI addresses, a third for port names) and compose them into the desired higher level logic by calling them in succession.

fnordahl (changes requested):
See in-line comments, and also note that there is a rich set of unit tests here, and we likely want unit test coverage for any changes.

charmhelpers/contrib/openstack/context.py
PanosKostopoulos: updated_bonds_key is the output of build_mapping_string which is a string, not a dict. This is because when parsing the original config option string, to extract pci addresses, it is split into dicts, and then is reconstructed.

PR #957, opened 2026-04-13
Add per-client IP rate limiting via Apache mod_qos

Raven-182:
When apache-rate-limit-enabled=true, Apache limits how many HTTP requests a client IP can make within a time window and returns HTTP 429 when limit is hit.

Builds on proxy protocol support (f516806411319fd967dee205701c2fabceb3144f). HAProxy forwards the real client IP, which mod_remoteip extracts, but mod_qos cannot use directly. A small Lua script bridges this gap by running after mod_remoteip has extracted the real IP and copying it into `X-Real-Client-IP` header that mod_qos can read.

The mod_qos module is installed and enabled automatically.

**A few notes:**
- When services are running behind HAProxy, proxy protocol must be enabled for rate limiting to take effect
- Limits are enforced per unit (not cluster-wide), so the effective limit scales with the number of backend units
- Cluster peer IPs are exempted to avoid throttling health checks
- [`LuaScope thread`](https://httpd.apache.org/docs/2.4/mod/mod_lua.html#luascope) and [`LuaCodeCache forever`](https://httpd.apache.org/docs/2.4/mod/mod_lua.html#luacodecache) directives are set for performance because the lua script is stateless and does not need to be re-read and re-compiled on every request

**Testing**
- Integration tested on 3-unit neutron-api HA deployment with HTTPS and proxy protocol
-  Integration tested on single-unit neutron-api with HTTPS:
    - Rate limiting still requires proxy protocol for neutron-api (local HAProxy service masks client IP without it)

**Note on mod_qos and mod_lua**

Noble ships version 11.74 for mod_qos which cannot use the client IP from mod_remoteip directly, newer versions of mod_qos (11.75+) improve this, so the lua script workaround may not be needed in the future.

See: https://mod-qos.sourceforge.net/CHANGES.txt

### DIFF
--- a/charmhelpers/contrib/openstack/context.py
+++ b/charmhelpers/contrib/openstack/context.py
@@ -1133,6 +1133,83 @@ class ImageServiceContext(OSContextGenerator):
         return {}
 
 
+# Script used with apache mod_qos to copy the real client IP from
+# mod_remoteip into a header
+LUA_QOS_REAL_IP_SCRIPT = '/etc/apache2/qos_real_ip.lua'
+LUA_QOS_REAL_IP_CONTENT = """\
+-- Copy the real client IP (extracted from the PROXY protocol
+-- header by mod_remoteip) into a request header that
+-- mod_qos can read via QS_ClientIpFromHeader.
+function translate(r)
+    r.headers_in['X-Real-Client-IP'] = r.useragent_ip
+    return apache2.DECLINED
+end
+"""
+DEFAULT_RATE_LIMIT_REQUESTS = 200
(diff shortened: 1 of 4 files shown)


Raven-182:
neutron-api patch: https://review.opendev.org/c/openstack/charm-neutron-api/+/985552

freyes (changes requested):
@Raven-182 , I left the copilot suggestions that make sense, maybe the one that requires some extra explanation would be the lack of `umask`, for CIS hardened systems, the default umask is changed , I don't remember the exact value used, it's better we use an explicit permission to be set when rendering the file.

